PR #5, opened 2018-02-05
TST: Rework .travis.yml to run tests

thequackdaddy:
Hello,

I took a shot at re-writing the travis script so that you can run all the tests you've built. Might also be nice to add coverage. If this works, I'll add that. 

I should mention there are a few caveats:
* nlopt seems broken. I'm not exactly sure how to install it but there's lots of recent issues getting it to work. I *think* this approach works, but its probably not ideal. 
* I have to install `rpy2` and then download, compile, install all of R `ChainLadders` dependents--of which there are many. It would be much, much faster if we could just extract those cause over 12 minutes of the build time is building all the R stuff which is useless. 
* We might want to discuss which versions of numpy/pandas you want to build against. Since this is a new package, I guess we can be somewhat strict. 

Let me know if this is of interest to you.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,31 @@
+# .coveragerc to control coverage.py
+[run]
+branch = True
+
+[report]
+show_missing = True
+
+# Regexes for lines to exclude from consideration
+exclude_lines =
+    # Have to re-enable the standard pragma
+    pragma: no cover
+
+    # Don't complain about missing debug-only code:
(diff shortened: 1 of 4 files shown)


thequackdaddy:
Update: I realized that conda provides most of the R dependencies pre-compiled, so I added as many as I could find. I think this should be much faster now. About 10 minutes vs. 18 before.

PR #9, opened 2018-02-15
Chart Dev Rebased try 3

thequackdaddy:
This is the latest and greatest attempt to rebase #3. The git history was a little messy, but I think this is clean enough.


thequackdaddy:
Ok. I think this is finished--Assuming this run passes... realized I should've xfailed the problem section.

You may wish to revisit the tabular format code, I don't think it works.

jbogaardt:
@thequackdaddy , you rock, thanks for fixing this.  Yah I know the tabular format is not working as it used to.  I had previously had some experimental column detection in there that would map origin and development, but it was weak and highly experimental, and not a high priority enhancement.  I am good with overriding the unit tests on tabular data without specifying origin and dev.  Thanks again.

PR #16, opened 2018-04-09
WIP: Simplify triangle class, lotsa PEP fixes

thequackdaddy:
So experimenting with this has been very enlightening. 

We don't have very good coverage for most of the codebase. I've also become a little confused by some of the code that is in here. I tried simplifying a few things, but I erred on the side of consistency. 

I think moving forward it would be good to move forward with a more simplified approach. I understand that reserving and reserving data can come in a variety of shapes and sizes. But if we are strict in how we require the `Triangle` class to be specified, I think we'll find getting all these pieces to work together much, much easier. To that end... 

* I think internally, we should store the triangles as `pd.Series`. We should use a `MultiIndex` where the first level of the index is the origin and second is the development. 
* I think a lot of the `np.nan_to_num`, `np.array`, etc. hackery can be avoided if we rely on pandas functionality. There's no need to re-invent the wheel if we don't have to. A lot of the code seems like it could have been thought of more simply as a pandas function--along with intelligent use of the pandas `Index`.
* `grain` and `TriangleSet` for now have been disabled by me. `Grain` I think could be much more elegantly handled uses pandas `PeriodIndex` for Origin.

Still a work-in-progress. But here is what I have. Its not as feature-complete as what you had before, but I think if moves the code to a place where it is more stable and easier to add functionality in the future. Still have a lot of re-write to go, but I think this is some progress.


jbogaardt:
@thequackdaddy , this is great.  Lot's of changes that I look forward to going though.

Some high level thoughts before heading into work today.  I really like the idea of cleaning up cleaning out the Triangle.__set_period() method using pandas dates.  It was ugly indeed and I am glad you made significant headway here.  I am all for simplifying the code base for ease of by putting slightly stricter requirements on the users.  However, there is critical functionality I'd like to not lose as follows:

* Users will want to extract a number of triangles from a single dataset. This was the impetus for the TriangleSet class.  It was hacky and deserves to be rewritten, but I would like to aspire to (some day) having a class of multiple triangles that the user can effortlessly 'slice' individual triangles from.
* Users will want to perform arithmetic on triangles.  For example, taking a case Incurred Loss triangle and dividing by a claim count triangle to produce a severity triangle is common practice and the Triangle class should facilitate this.  I notices the arithmetic methods have been removed form the class.  Were these causing side-effects for you?  
* Users will want to move between finer and coarser grains of triangles.  For example,  they may do their analysis at the accident month/development month grain, but might want to explore visualizations at the accident year/development quarter grain.  Triangle.grain() method provides the functionality to quickly move between grains.
 
These are just some high level thoughts, that I'd be happy to hear a different perspective on.  Thanks for the PR and again, I look forward to reviewing it.

jbogaardt:
Sorry, Not sure how I missed the arithmetic operators.  They are in there.  

I've begun going through the Triangle class updates and a very desirable feature that seems to have been lost is method chaining and I wanted to discuss that before diving too deep.
Previously, the methods of the class would return a Triangle with **return self** (with an option to mutate the original or not).  This allowed for the ability to method chain on object instances.  
For example a user could write something like **tri.data_as_table().incr_to_cum()** and it would be legal.  Additionally, the methods had the desired property of being commutative.  That is  **tri.data_as_table().incr_to_cum()** would produce the same results as **tri.incr_to_cum().data_as_table()** This is no longer possible with this PR as the methods generally return DataFrames and I don't think the inplace option on the methods does anything anymore.

Was the removal of method chaining an intent of your updates?

thequackdaddy:
Very good point. 

Let me re-think some of this. 

I think what we might want to focus on having the `Triangle` class "transformations" (almost) always return a `Triangle`. Only if you specfically ask for `data` or `triangle_data` should it return data in `Series` or `DataFrame` format. 

Then the focus of the different reserving techniques (Chainladder, Stochastic, Mack, etc) would always rely on a `Triangle` input. 

Let me experiment with that idea.

PR #52, opened 2019-11-25
Fix for weights when development grain not equal to origin grain

allenclong:
Small change to _assign_n_periods_weight_int in development>base.

If development is quarterly, but origin in accident years, for example:
If n_periods= x then the last valuation should be n_periods*4 -1 prior to the last valuation. 
So if last valuation is 2006Q1, the minimum valuation to get weights in a 3 yr average should be 2003Q1. In a 5 year average 2001Q1.


jbogaardt:
This is great.  Thanks for fixing!

Issue #55, opened 2019-12-01
Development lag calculation is off by one quarter


allenclong:
Importing data and creating a triangle with origin grain of 'Y' and development grain of 'Q' results in development calculation being off by 3.

Recreate:
`dict = {'state':['ST']*21, 'AY':['2017']*11 + ['2018']*7 + ['2019']*3,
        'evaluation_date':['3/31/2017','6/30/2017','9/30/2017','12/31/2017', '3/31/2018','6/30/2018','9/30/2018','12/31/2018', '3/31/2019','6/30/2019','9/30/2019'
                          , '3/31/2018','6/30/2018','9/30/2018','12/31/2018', '3/31/2019','6/30/2019','9/30/2019', '3/31/2019','6/30/2019','9/30/2019'],
        'paid': list(range(3,36,3)) + list(range(3,24,3)) + list(range(3,12,3))
       }
df = pd.DataFrame(dict)

triangle = cl.Triangle(df, 
        origin='AY',
        development=['evaluation_date'], 
        index = ['state'],
        columns = ['paid']
    )

print(triangle[triangle.development <= 6])`

Adding a zero to QuarterEnd in _period_end in chainladder/core/base.py from:
`offset = {12: pd.tseries.offsets.MonthEnd(),
                  4: pd.tseries.offsets.QuarterEnd(),
                  1: pd.tseries.offsets.YearEnd()}`
to 
`offset = {12: pd.tseries.offsets.MonthEnd(),
                  4: pd.tseries.offsets.QuarterEnd(0),
                  1: pd.tseries.offsets.YearEnd()}`
seems to fix the problem. It doesn't appear to be an issue with annual grains. I have not tested month grains.

jbogaardt:
Running your example (from latest chainladder release), I'm not sure I see the issue.  Here is my [In]/[Out]
```python
dict = {'state':['ST']*21, 'AY':['2017']*11 + ['2018']*7 + ['2019']*3,
'evaluation_date':['3/31/2017','6/30/2017','9/30/2017','12/31/2017', '3/31/2018','6/30/2018','9/30/2018','12/31/2018', '3/31/2019','6/30/2019','9/30/2019'
, '3/31/2018','6/30/2018','9/30/2018','12/31/2018', '3/31/2019','6/30/2019','9/30/2019', '3/31/2019','6/30/2019','9/30/2019'],
'paid': list(range(3,36,3)) + list(range(3,24,3)) + list(range(3,12,3))
}
[...]

PR #56, opened 2019-12-01
Fix to address issue 55

allenclong:
This is the fix mentioned in the issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,16 +18,16 @@ matrix:
   - python: 3.6
     env:
     - PANDAS: 0.23
-    - SCIKIT-LEARN: 0.19.0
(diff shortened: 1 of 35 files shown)


jbogaardt:
@allenclong , this PR has a conflict with master.  I think it is because I likely already fixed the issue as part of 4ba876e.  In, that commit, I completely eliminated `_period_end` method altogether.  That was only two days ago, so your branch may have been slightly behind.

As far as I can tell, #55 was corrected with that commit, but I would appreciate if you can confirm that the lastest code base works on your end.  If so, I'll close out this PR.

jbogaardt:
pretty simple resolution, so just resolving conflicts and merging

Issue #76, opened 2020-05-13
Independence tests for Chain Ladder 


gig67:
It is important that some critical assumptions around chain ladder are tested before applying the method. Thomas Mack suggested some tests, for example in "Measuring the variability of Chain Ladder reserve estimates", 1997. 
Below are implementations for tests of correlation among development factors and for the impact of calendar years. I don't feel confident enough to modify the package code directly but hopefully this can be a useful template.

```
def developFactorsCorrelation(df):
    # Mack (1997) test for Correlations between Subsequent Development Factors
    # results should be between -.67x and +.67x stdError otherwise too much correlation
    m1=df.rank() # rank the development factors by column
    m2=df.to_numpy(copy=True) #does the same but ignoring the anti-diagonal
    np.fill_diagonal(np.fliplr(m2),np.nan)
    m2=pd.DataFrame(np.roll(m2,1),columns=m1.columns, index=m1.index).iloc[:,1:] #leave out the first column
    m2=m2.rank()
    numerator=((m1-m2) **2).sum(axis=0)
    SpearmanFactor=pd.DataFrame(range(1,len(m1.columns)+1),index=m1.columns, columns=['colNo'])
    I = SpearmanFactor['colNo'].max()+1
    SpearmanFactor['divisor'] = (I-SpearmanFactor['colNo'])**3 - I +SpearmanFactor['colNo'] #(I-k)^3-I+k
    SpearmanFactor['value']= 1-6*numerator.T/SpearmanFactor['divisor']
    SpearmanFactor['weighted'] = SpearmanFactor['value'] * (I-SpearmanFactor['colNo']-1) / (SpearmanFactor[1:-1]['colNo']-1).sum() #weight sum excludes 1 and I
    SpearmanCorr=SpearmanFactor['weighted'].iloc[1:-1].sum() # exlcuding 1st and last elements as not significant
    SpearmanCorrVar = 2/((I-2)*(I-3))
    return SpearmanCorr,SpearmanCorrVar
```

```
from scipy.stats import binom
def calendarCorrelation(df, pCritical=.1):
    # Mack (1997) test for calendar year effect
    # A calendar period has impact across developments if the probability of the number of small (or large)
    # development factors in that period occurring randomly is less than pCritical
    # df should have period as the row index, on the assumption that the first anti-diagonal is in relation to the same period (development=0) 
    m1=df.rank() # rank the development factors by column
    med=m1.median(axis=0) # find the median value for each column
[...]

PR #77, opened 2020-05-14
Update triangle.py

gig67:
Introduce two tests for correlation of dev factors 1) across dev years and 2) by calendar year (period)
I would have liked to test before pulling but have been unable so far to install this branch on my local machine. Any advice welcome.

### DIFF
--- a/chainladder/core/triangle.py
+++ b/chainladder/core/triangle.py
@@ -4,6 +4,7 @@
 
 import pandas as pd
 import numpy as np
+from scipy.stats import binom
 from chainladder.utils.cupy import cp
 import copy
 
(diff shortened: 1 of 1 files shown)


jbogaardt:
Merging. Thanks for the contribution.  I can work through the testing.  As for playing with a development version you can `pip uninstall chainladder` to get rid of chainladder and then from the root directory of the project use `python setup.py develop`.  That will install the package in a way where you can modify the code and have the updates reflect the next time you `import chainladder as cl`.  Keeping this open until testing is complete.

jbogaardt:
Linking to issue #76.

gig67:
Thanks a lot, let me know if you wish me to do something on the documentation (and how).
Also, you will see I imported binom from scipy.stats and maybe this needs to also be listed in the requirements file.

PR #79, opened 2020-05-21
update correlation.py

gig67:
fixed bug in sizing T lookup table


jbogaardt:
Beautiful.  I think with this one, issue #76  is done.

Issue #92, opened 2020-08-25
to_frame() with index as datetime


johalnes:
Trying some make a frontend in use for analytical purposes with the chainladder package. And while period index is good in many case, many visiualization tools don't like them with the error "Object of type Period is not JSON serializable". This includes streamlit and dash/plotly.

Even if it's quite easy to fix - could it be a idea with a "convert_index" option in the to_frame() method?

```
import numpy as np
import pandas as pd

import plotly.express as px
import chainladder as cl
import plotly as pl
print(f'CL: {cl.__version__} \n PD:{pd.__version__} \n NP: {np.__version__} \n Plotly: {pl.__version__}' )
data = pd.read_csv('https://raw.githubusercontent.com/casact/chainladder-python/master/chainladder/utils/data/prism.csv')
data['AccYr'] = data['AccidentDate'].str[:4]

x = cl.Triangle(data=data,
            origin='AccYr', development='PaymentDate',
            columns=['Paid', 'Incurred'],
            origin_format='%Y', development_format='%Y-%m-%d').incr_to_cum()
mack = cl.MackChainladder()
dev = cl.Development(average='volume')
mack.fit(dev.fit_transform(x))
```

__Now this works:__
```
summary_df = mack.summary_['Paid'].to_frame()
summary_df.index = summary_df.index.astype('datetime64[ns]')
px.line(summary_df)
```

__This breaks:__
`px.line(mack.summary_['Paid'].to_frame())`

__would like to do:__
`px.line(mack.summary_['Paid'].to_frame(convert_index=True)`

or 
`px.line(mack.summary_['Paid'].to_frame(index_astype='date'))`

johalnes:
Didn't copy the output since it's mostly graphical. But I'm using CL: 0.7.1 PD:1.0.3 NP: 1.18.1 PlotlyExpress: 4.9.0

jbogaardt:
This makes sense as an additon. I've run into this myself.  There is more flexibility in option 2 so I like that one better.

jbogaardt:
It may need tb be an 'origin_astype' argument rather than 'index_astype'.  The `to_frame` method often doesn't set the origin as the dataframe index as in this example which places the origin into the dataframe columns:
```python
import chainladder as cl
clrd = cl.load_sample('clrd')
clrd.latest_diagonal['CumPaidLoss'].to_frame()
```

johalnes:
[...]

PR #93, opened 2020-08-26
to_frame() with origin as datetime 

johalnes:
Reference to #92  - added a first attempt.

Saddly I'm working on windows where rpy2 doesn't work so haven't been able to run Pytest. Will make a Docker image for this soon, but for now relying on github Action which passed.

### DIFF
--- a/chainladder/core/display.py
+++ b/chainladder/core/display.py
@@ -62,8 +62,8 @@ class TriangleDisplay():
         else:
             return '{:,.0f}'
 
-    def _repr_format(self):
-        odims, ddims = self._repr_date_axes()
+    def _repr_format(self, origin_as_datetime=False):
+        odims, ddims = self._repr_date_axes(origin_as_datetime)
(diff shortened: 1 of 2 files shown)


johalnes:
I made a dockerfile in my forked chainladder repo to be able to test this easily on windows. On the rpy2 website they state that it's not that easy to install on windows. Feel free to take it into the master if you think it could be hepful. 

__Build docker__ by running `docker build --tag chainladder_test:0.1 .` in the root project folder. 
__Run__ all test with `docker run chainladder_test:0.1`
Get into the shell script with `docker run --entrypoint "/bin/sh" -it chainladder_test:0.1`

jbogaardt:
Sure, I'll take it. I'm not great with docker myself, but having a consistent environment should be helpful with more collaborators.

I normally use Conda to install rpy2 to a Windows environment, but docker makes sense as well.

Issue #113, opened 2021-01-27
Too small LDFs included in exponential tail curve fit


Brian-13:
When fitting an exponential tail curve with the log regression, only LDFs < 1.0 are excluded. However, also LDFs close to 1, e.g. <1.0001 can distort the tail fit considerably.

**Suggestion:**
Replace in curve.py rows 111-113
`if self.errors == "ignore":`
`_w[_y <= 1.0] = 0`
`_y[_y <= 1.0] = 1.01`

with something like

`if self.errors == "ignore":`
`_w[_y <= threshold] = 0`
`_y[_y <= threshold] = 1.01`

and let the user define the `threshold` with a default like 1.0001 or 1.00001.

jbogaardt:
This us a good suggestion and probably not too difficult to do.

jbogaardt:
@Brian-13 , you seem to be willing to dive into the modules and you've forked the code. Would you like me to assign this to you?

jbogaardt:
Another alternative is to allow for more refined controls than is currently supported by the `fit_period` parameter.

Brian-13:
Hi @jbogaardt, yes, feel free to assign it to me. I'm currently testing a solution and make this available in the next days.

jbogaardt:
closed by #114

PR #114, opened 2021-02-02
Solution for issue #113

Brian-13:
Created a reg_threshold tuple to set lower and upper thresholds for the log regression used in the curve fit. See issue #113 for more details.


jbogaardt:
Nice work @Brian-13 .  I like the thresholding tuple idea and especially appreciate keeping existing default bevahior unchanged.

In the future, we might want to consider changing the inequality: `(_y <= lower_threshold)` to `(_y < lower_threshold)` so that the default threshold can be `(1.0, None)`

PR #116, opened 2021-02-02
Add VotingChainladder (resolves #115)

### LINKED ISSUE #115: Allow ensembling of methods by origin period
In practice actuaries may opt to apply different methods by origin period. For example, applying a basic chainladder for origin periods less than 75% developed, and a Bornhuetter-Ferguson thereafter. Alternatively, we may wish to combine methods such as predicting ultimate claims as the average of a basic Chainladder, Bornhuetter-Ferguson, and Cape-Cod.

Currently, there is no built in approach to achieve this. One can define an array of weights and manually derive the ultimate, but a more elegant built in solution is preferable.

I propose a `VotingChainladder` method be incorporated that can take a list of estimators and an array of weights and predict based on the weights.

To achieve this:

- Create an ensemble module within chainladder
- Define a `_BaseTriangleEnsemble` class that inherits from _sklearn_'s `_BaseHeterogeneousEnsemble` for the underlying logic and methods (this necessitates adding an attribute to `MethodBase` to identify a chainladder method, as well as adding an "`is_chainladder`" check function to `core.base`.
- Define a `_BaseChainladderVoting` class that inherits `_BaseTriangleEnsemble` and _sklearn_'s `_BaseVoting` for functionality. This class will be responsible for validation (and creation if required) of weights and defining the fit method.
- Define the `VotingChainladder` method which inherits `_BaseChainladderVoting` and _chainladder_'s `MethodBase`. This will be the actual voting method used and will allow prediction of ultimates.

The functionality of the voting is as follows:

Define an array of weights shaped `(index, columns, origin, n_estimators, 1)`, then your ultimate is calculated as:
```
ultimate = sum([
        est.predict(X, sample_weight).ultimate_ * weights[..., i, :]
        for i, est in enumerate(self.estimators_)
    ]) / weights.sum(axis=-2)
```

A pull request with the proposed additions will follow.

cbalona:


cbalona:
Tests seems to be failing due to an sklearn import error: ImportError: cannot import name 'delayed' from 'sklearn.utils.fixes' (/usr/share/miniconda/envs/cl_test/lib/python3.7/site-packages/sklearn/utils/fixes.py)

jbogaardt:
This is an awesome PR!  I wouldn't have been able to come up with something this good.  

Seems to only need a resolution for:
1. `the `ImportError` - is it a specific version of sklearn that we need to limit to?

Not needed for the PR, but we should also:
2. Add the class to to docs/modules/classes.rst TOC so its docstring shows up in the API docs
3. Perhaps take your unit test and add a plot to get it as a stand-alone 'Example' in the /examples folder for those that browse the docs by the example gallery
4. Add ensemble.rst to docs/modules/index.rst so the docs show up.  
4. Add it to README.rst on the main repo page

cbalona:
> This is an awesome PR! I wouldn't have been able to come up with something this good.

Thank you John! I am glad you like it.

> Seems to only need a resolution for:
> 
> 1. `the `ImportError` - is it a specific version of sklearn that we need to limit to?

I don't think we should limit as it seems that the need to import delayed from `sklearn.utils.fixes` is temporary, and sklearn intends to remove that function once this issue is resolved: https://github.com/joblib/joblib/issues/1071

As a more permanent fix, I am using `delayed` from joblib instead--it should do the same thing and all tests pass on my side.

> Not needed for the PR, but we should also:
> 2. Add the class to to docs/modules/classes.rst TOC so its docstring shows up in the API docs
> 3. Perhaps take your unit test and add a plot to get it as a stand-alone 'Example' in the /examples folder for those that browse the docs by the example gallery
> 4. Add ensemble.rst to docs/modules/index.rst so the docs show up.
> 4. Add it to README.rst on the main repo page

Agreed on all the above.

jbogaardt:
Fantastic contribution @cbalona.  Thank you so much!

RonRichman:
Thanks everyone - great contribution @cbalona!

PR #187, opened 2021-08-09
Revised the Development Tutorial

kennethshsu:
Improved tutorial, the diffs are bigger than expected for base.py, can you double-check what's going on there?


jbogaardt:
I did some refactoring of base.py this weekend, so wouldn't worry to much about that one.

Issue #191, opened 2021-08-15
heatmap() should utilize a diverging color spectrum and use ranking instead of percentile


kennethshsu:
Diverging color is better so we can see which values are above or below the medium LDF
Using ranking instead will force the display to have half the LDFs being high (or "hot"), and half being low (or "cold")

jbogaardt:
Resolved by #194

jbogaardt:
I think we need to rework this one.  The argument `gmap`  is only available in pandas 1.3 and later.  Chainladder currently supports pandas>=0.23.  Many environments are not yet on pandas==1.3.  For example google colab is on 1.1.5 and when installing chainladder, it doesn't force a reinstall of pandas == 1.3.

PR #193, opened 2021-08-16
Reran the first 2 tutorials with 0.8.6 and a better cmap choice in heatmap()

kennethshsu:


jbogaardt:
Does this cmap change address #191?

kennethshsu:
No, it does not, not the part where we apply the gradient on rank instead. I am working on it, let's see if I can figure it out.

PR #194, opened 2021-08-16
Modified the heatmap so that colors are apply over rank instead

kennethshsu:


jbogaardt:
I believe this resolves #191

PR #237, opened 2021-12-06
Citations & Bibliography

SRwcf:
I have updated the references used in the jupyter book and made a bibliography to systematically store references in one place. Thank You!


jbogaardt:
Thank you @SRwcf , what a great contribution!

Issue #269, opened 2022-03-08
Bug in .full_triangle()


kennethshsu:
I think there's a bug in `.full_triangle_`

```python
prism = cl.load_sample("prism")
bf_model = cl.BornhuetterFerguson(apriori=0.90)
bf_fit = bf_model.fit(
    X=prism["Incurred"].sum(), sample_weight=prism["Incurred"].latest_diagonal.sum()
)
```

My understanding is that `.full_expectation_`, is the expected triangle, backfilling all known values. This looks good.
```python
bf_fit.ultimate_ - bf_fit.full_expectation_.iloc[:, :, :, -1]
```

However, when calling `.full_triangle_`, which is the expected triangle, not backfilling all known values (upper left triangle). I am getting non-zero results.
```python
bf_fit.ultimate_ - bf_fit.full_triangle_.iloc[:, :, :, -1]
```

jbogaardt:
Interesting. `full_triangle_` is incremental, likely because the underlying triangle is incremental.
```python
print(bf_fit.full_expectation_.is_cumulative)
print(bf_fit.full_triangle_.is_cumulative)
print(bf_fit.ultimate_ - bf_fit.full_triangle_.incr_to_cum().iloc[..., -1])
```
We should probably make `full_expectation_` also behave as the underlying triangle in incremental fashion.  I don't think we should state them as cumulative (unless the user does so with their original triangle). It is better to assume the user is using incremental triangles to save on RAM. Incremental is my prefered format when working with large triangles.

kennethshsu:
Instead of making it defaulted to incremental, don't you think it's better that the triangles match? If the underlying triangle is cumulative, the `full_expectation_` and `full_triangle_` should both be cumulative. I think this is more intuitive. What do you think?

jbogaardt:
sorry, yes, thats what I mean. Whichever `is_cumulative` status the main triangle has should be inherited by both `full_expectation_` and `full_triangle_`

kennethshsu:
Ok cool, glad we are on the same page now. Let me try to fix this one.

kennethshsu:
Is there a way to check if the underlying triangle passed to an IBNR method is a cumulative or incremental triangle?

I see there's self.X_, but I think this is always is cumulative form?

jbogaardt:
Interesting...again. I would  `X_` to preserve its `is_cumulative`. It does so during `fit`, but it mutates the triangle during `predict`:

```python
import chainladder as cl
[...]

Issue #143, opened 2021-04-06
full_triangle_ for BF method


mlbugs:
BF method doesn't seem to be creating the runoff triangle correctly.  In my data, accident months can have volatile reported losses as of the latest valuation date.  Using an expected IBNR method can therefore produce volatile LDF's (implied, using ultimate_ / latest.diagonal ).  In my particular case, when reported losses are an outlier on the low end, the implied LDF is very high.  When I extract the first diagonal from the runoff triangle and divide that into my ultimate*, the resulting LDF's are much more homogeneous than they were the previous months.  Assuming my development pattern is relatively benign, there shouldn't be this much smoothing (development) in a one month projection.  For example, if the percent of ultimate for month _n_ is estimated to be 90% and for _n+1_ it's 95%, the LDF shouldn't drop from 3.0 to 1.2, it should drop to 1.6 (only 3.3% of ultimate has been reported for the month in questions as of month _n_).  The projected ultimate is only 13.3% and an expectation of 5% being reported next month would yield an expected reported loss of 8.3% and an implied ldf of 1.6.

Use case here is daily interpolation between month start and month end (ie, subsequent) LDF's.  It's better to interpolate to the expected month-end ldf than the subsequent accident month's ldf as this could produce large swings in your projected ultimate losses during the month.

*First diagonal from runoff triangle:
bf_table2 = bf_table.full_triangle_.dev_to_val()
bf_table3 = bf_table2[bf_table2.valuation==(triangle.valuation_date + relativedelta(months=+1))]
next_mo_ldf = bf_table.ultimate_ / bf_table3

for relativedelta, use "from dateutil.relativedelta import *"

jbogaardt:
Hi @mlbugs , Let me state how it _currently_ works - The run-off for future valuations of a triangle is entirely a function of the estimators `ultimate_` vector and its `cdf_`.  Specifically it is `estimator.ultimate_ / estimator.cdf_`.  This "run-off" formula is shared across all estimators including the expected loss methods. See [code implementation](https://github.com/casact/chainladder-python/blob/master/chainladder/core/common.py#L19-L20).   

Are you suggesting that for expected loss methods, the run-off should also be a function of the previous diagonal as well?  

It would help me if we could work from a reproducible example.  Is it possible to reproduce the issue from this sample as a baseline as well as your expectation for the correct answer?

```python
import chainladder as cl

triangle = cl.load_sample('raa')
exposure = triangle.latest_diagonal*0 + 20000
[...]

PR #271, opened 2022-03-11
#269 Corrected an error of inconsistency in full_expectation_ and full_triangle_

kennethshsu:
Addressed an error and refactored the code per #269 

Test code:
``` python
def test_full_triangle_and_full_expectation(raa):
    raa_cum = raa
    assert raa_cum.is_cumulative == True

    raa_incr = raa_cum.cum_to_incr()
    assert raa_incr.is_cumulative == False
    assert raa_incr.incr_to_cum().is_cumulative == True
    assert raa_incr.incr_to_cum() == raa_cum

    cl_fit_incr = cl.Chainladder().fit(X=raa_incr)
    assert cl_fit_incr.X_.is_cumulative == False

    cl_fit_cum = cl.Chainladder().fit(X=raa_cum)
    assert cl_fit_cum.X_.is_cumulative == True

    assert cl_fit_incr.cdf_ == cl_fit_cum.cdf_
    assert cl_fit_incr.ultimate_ == cl_fit_cum.ultimate_

    assert cl_fit_cum.full_expectation_ == cl_fit_incr.full_expectation_.incr_to_cum()
    assert (
        cl_fit_cum.full_triangle_ - cl_fit_incr.full_triangle_.incr_to_cum() < 0.00001
    )
    assert (cl_fit_cum.full_triangle_ - raa_cum) - (
        cl_fit_incr.full_triangle_.incr_to_cum() - raa_incr.incr_to_cum()
    ) < 0.00001

    bf_fit_incr = cl.BornhuetterFerguson(apriori=1).fit(
        X=raa_incr, sample_weight=raa_incr.incr_to_cum().latest_diagonal * 0
    )
    assert bf_fit_incr.X_.is_cumulative == False

    bf_fit_cum = cl.BornhuetterFerguson(apriori=1).fit(
        X=raa_cum, sample_weight=raa_cum.latest_diagonal * 0
    )
    assert bf_fit_cum.X_.is_cumulative == True

    assert bf_fit_incr.cdf_ == bf_fit_cum.cdf_
    assert bf_fit_incr.ultimate_ == bf_fit_cum.ultimate_

    assert bf_fit_cum.full_expectation_ == bf_fit_incr.full_expectation_.incr_to_cum()
    assert (
        bf_fit_cum.full_triangle_ - bf_fit_incr.full_triangle_.incr_to_cum() < 0.00001
    )
    assert (bf_fit_cum.full_triangle_ - raa_cum) - (
        bf_fit_incr.full_triangle_.incr_to_cum() - raa_incr.incr_to_cum()
    ) < 0.00001
```

### DIFF
--- a/chainladder/core/common.py
+++ b/chainladder/core/common.py
@@ -10,11 +10,19 @@ from chainladder.utils.utility_functions import concat
 from chainladder import options
 
 
-def _get_full_expectation(cdf_, ultimate_):
+def _get_full_expectation(cdf_, ultimate_, is_cumulative):
     """ Private method that builds full expectation"""
     full = ultimate_ / cdf_
 
-    return concat((full, ultimate_.copy().rename("development", [9999])), axis=3)
+    if is_cumulative:
+        return concat((full, ultimate_.copy().rename("development", [9999])), axis=3)
+
+    else:
+        tail_ = full.iloc[:, :, :, -1] - ultimate_
+
+        return concat(
+            (full.cum_to_incr(), tail_.copy().rename("development", [9999])), axis=3
+        )
 
 
 class Common:
@@ -52,7 +60,7 @@ class Common:
                 + "' object has no attribute 'full_expectation_'"
             )
 
-        return _get_full_expectation(self.cdf_, self.ultimate_)
+        return _get_full_expectation(self.cdf_, self.ultimate_, self.X_.is_cumulative)
 
     @property
     def full_triangle_(self):
@@ -63,7 +71,9 @@ class Common:
                 + "' object has no attribute 'full_triangle_'"
             )
 
-        full_expectation = _get_full_expectation(self.cdf_, self.ultimate_)
+        full_expectation = _get_full_expectation(
+            self.cdf_, self.ultimate_, self.X_.is_cumulative
+        )
         frame = self.X_ + full_expectation * 0
         xp = self.X_.get_array_module()
         fill = (xp.nan_to_num(frame.values) == 0) * (self.X_ * 0 + full_expectation)
--- a/chainladder/core/common.py
+++ b/chainladder/core/common.py
@@ -6,39 +6,15 @@ import pandas as pd
 from chainladder.utils.cupy import cp
 from chainladder.utils.sparse import sp
 from chainladder.utils.dask import dp
+from chainladder.utils.utility_functions import concat
+from chainladder import options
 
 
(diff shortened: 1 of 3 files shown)


jbogaardt:
@kennethshsu, I notice that this code coerces `full_expectation_` and `full_triangle_` to use the same underlying calculation.  When these two properties were first developed they worked that way. However, they we separated out the calcs because of #143.  Can you take a look and let me know your thoughts?

kennethshsu:
I think the discussion makes a lot of sense, and I do like that the emergence should really only be based on `_ibnr` so we don't get that kink to start with. However, I don't really like the fact that we are defining `full_expectation_` vs `full_triangle_` as the identical besides the known upper left triangle. Here's what I propose:

- `full_expectation_`: continue to use `.ultimate_/.cdf_`
- `full_triangle_`: continue to use `.ultimate_/.cdf_` for the lower part of the triangle, and known values for the upper part of the triangle. I think there's value in the relationship between `full_expectation_` and full_triangle_`, this is good for back-testing actual vs expected.
- a new attribute, perhaps `expected_ibnr_`? This will correct for the kinks, and focuses only on bridging the `latest_diagonal` to `ultimate_`

What do you think?

kennethshsu:
On second thought, how about in both `full_expextation_` and `full_triangle_` we add an option to normalize the pattern with `ibnr_`? Maybe `normalized_latest` as a boolean variable and default it to `True`?

jbogaardt:
This topic has only received light research, but I'd really like to resist the temptation to change the calcs without additional support from actuarial literature.  In #143, there is reference to the [Clark paper](https://www.casact.org/sites/default/files/2021-03/7_Clark.pdf). We set `full_triangle_` to adhere to the loss emergence formulation on page 47 (page 7 by page count). Per the paper, incremental loss emergence is different between Chainladder "LDF" methods and Expected Loss "Cape Cod" methods. Its not clear to me that it should be an option to be selected by the user. Its seems a byproduct of the type of IBNR method chosen.

Perhaps this PR can be reworked to address the incremental/cumulative inconsistency between `full_expectation_` and `full_triangle_` highlighted in #269, and move the general topic of loss emergence to a discussion that aims to solicit deeper research on the topic?

kennethshsu:
Yap sounds good, will do that.

kennethshsu:
I submitted another PR, but I got a lot of conflict, hopefully I resolved everything correctly. Can you take a closer look in case I made a mistake?

PR #273, opened 2022-03-14
Reverted the full_expectation_ to the old method of calculating expected values per #269

kennethshsu:


kennethshsu:
Test code:

```python
import chainladder as cl
import numpy as np
import matplotlib.pyplot as plt

raa_cum = cl.load_sample("raa")
raa_incr = raa_cum.cum_to_incr()
assert cl.Chainladder().fit(raa_incr).full_triangle_ - cl.Chainladder().fit(raa_cum).full_triangle_.cum_to_incr() <= 0.0001
```

jbogaardt:
This introduced some [unit test failures](https://github.com/casact/chainladder-python/runs/5532630226?check_suite_focus=true#step:4:950). I think they're easy fixes.  Can you address?

kennethshsu:
Hmm don't know how I didn't spot that one, it should've failed the test on my end too.

Issue #274, opened 2022-03-15
Heatmap - pandas deprecation of .render()


genedan:
Can I take a shot at making a pull request? The following code, when run in an interpreter with Pandas 1.4.1 will trigger a deprecation warning (the warning doesn't appear when I tried it in Jupyter):

![image](https://user-images.githubusercontent.com/6165650/158465846-80d5fcac-ca66-49bc-8259-3b12194ece64.png)

https://github.com/casact/chainladder-python/blob/bc8a52b72f2e202335ca48df5192c85ee0306e44/chainladder/core/display.py#L123-L134

I'd like to see if I can make my first contribution, are the steps to 1) change the code 2)  run pytest, and 3) issue the pull request?

jbogaardt:
Yes, that's pretty much how it works. If you forget to run tests, they will run in the  PR. Also, it's good practice to mention the issue number in your commit message(s) so we can link code changes to their source. I Look forward to seeing a PR!

jbogaardt:
Resolved by #280

PR #280, opened 2022-03-19
- Addresses future deprecation of .render method for pandas.io.format…

genedan:
…s.style.Styler class in pandas version 1.4.0, mentioned in #274.

- No tests concerning display.py were found, so I added one that simply renders the heatmap to catch any exceptions or warnings.

- I thought about either adding another condition to https://github.com/casact/chainladder-python/blob/bc8a52b72f2e202335ca48df5192c85ee0306e44/chainladder/core/display.py#L123-L134, or to just replace .render() with .to_html(). I tried the latter and the test passed, so I think it should be ok as this method was introduced in pandas 1.3.0. As expected, running the test on the original code with pandas 1.4.1 triggers the warning but does not if run with pandas 1.3.5.

- Running pytest on the whole repository at HEAD triggered a bunch of errors and warnings so I focused on just this one test I added. It did trigger 2 warnings prior to any edits to display.py.

### DIFF
--- a/chainladder/core/display.py
+++ b/chainladder/core/display.py
@@ -131,7 +131,7 @@ class TriangleDisplay:
                         subset=subset,
                         gmap=gmap,
                     )
-                    .render()
+                    .to_html()
                 )
             else:
                 default_output = (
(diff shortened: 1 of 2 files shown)


jbogaardt:
Looks great.  Welcome aboard @genedan!

kennethshsu:
What's funny is that I think this was also the first function that I updated. :)

Issue #283, opened 2022-03-22
Bug in the calculation of LDFs in CaseOutstanding()


kennethshsu:
Offending code:

```python
import chainladder as cl
usauto = cl.load_sample('usauto')
model = cl.CaseOutstanding(paid_to_incurred=('paid', 'incurred'), paid_n_periods = 3, case_n_periods = 3).fit(usauto)

model.paid_to_prior_case_ 

model.paid_ldf_

model.case_to_prior_case_

model.case_ldf_
```

`model.paid_ldf_` and `model.case_ldf_` do not reflect `model.paid_to_prior_case_` and `model.case_to_prior_case_` and weird results are returned.

PR #284, opened 2022-03-25
Addressed 283

kennethshsu:
Addressed #283.

Test code:
```python
import chainladder as cl
import numpy as np
usauto = cl.load_sample("usauto")
model = cl.CaseOutstanding(
    paid_to_incurred=("paid", "incurred"), paid_n_periods=3, case_n_periods=3
).fit(usauto)
expected_paid_ldf = np.array(
    [[0.833, 0.701, 0.714, 0.714, 0.653, 0.631, 0.553, 0.437, 0.524,]]
)
assert (
    model.paid_ldf_.to_frame(origin_as_datetime=False).values - expected_paid_ldf
    < 0.001
).all()

expected_case_ldf = np.array(
    [[0.526, 0.566, 0.528, 0.486, 0.511, 0.555, 0.652, 0.674, 0.580,]]
)
assert (
    model.case_ldf_.to_frame(origin_as_datetime=False).values - expected_case_ldf
    < 0.001
).all()
```

### DIFF
--- a/chainladder/development/outstanding.py
+++ b/chainladder/development/outstanding.py
@@ -10,7 +10,7 @@ import pandas as pd
 
 
 class CaseOutstanding(DevelopmentBase):
-    """ A determinisic method based on outstanding case reserves.
+    """A determinisic method based on outstanding case reserves.
 
     The CaseOutstanding method is a deterministic approach that develops
     patterns of incremental payments as a percent of previous period case
@@ -49,8 +49,10 @@ class CaseOutstanding(DevelopmentBase):
     paid_ldf_:
         The selected paid to prior case ratios of the fitted estimator
     """
-    def __init__(self, paid_to_incurred=None, paid_n_periods=-1,
-                 case_n_periods=-1, groupby=None):
+
+    def __init__(
+        self, paid_to_incurred=None, paid_n_periods=-1, case_n_periods=-1, groupby=None
+    ):
         self.paid_to_incurred = paid_to_incurred
         self.paid_n_periods = paid_n_periods
         self.case_n_periods = case_n_periods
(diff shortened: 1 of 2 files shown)


jbogaardt:
I think we're getting regressions in the test_workflow.py test. I think the issue is here:

```python
new_case_ldf_ = (
            self.case_to_prior_case_.to_frame(origin_as_datetime=True)
            .replace(0, np.NaN)
            .mean()
            .values
        ).reshape(ldf_shape)
```
Using `to_frame` + pandas operations only works for single triangles and doesn't scale to a full 4D Triangle.  Based on how you wrote your fix, can't we use the new `fillna` method you wrote and keep everything in triangles/arrays?
```python
new_case_ldf_ = self.case_to_prior_case_.fillna(self.case_to_prior_case_.mean(2))
```

kennethshsu:
Yes, good idea, and that worked! 

Is it useful to have `num_to_nan` as a method instead of a function? That way we wouldn't have to declare the variable and then manipulate its `.values`. Also, I think `num_to_nan` is a bit confusing, is `zero_to_nan` more appropriate?

kennethshsu:
On my local pytest I am still getting lots of errors though...?

jbogaardt:
> Is it useful to have `num_to_nan` as a method instead of a function?

It would make the code easier to manage. We should consider making it a private method though (denoted by a leading underscore).   In my mind, a triangle in its stable state doesn't have any zeros, only nans. Said differently, I don't think there is a use case for a public method for `num_to_nan`.

> Also, I think num_to_nan is a bit confusing, is zero_to_nan more appropriate?

I've always considered it an array function and the inverse of the [np.nan_to_num](https://numpy.org/doc/stable/reference/generated/numpy.nan_to_num.html) which is why I named it so.

PR #289, opened 2022-04-01
Address origin_as_datetime bug

kennethshsu:
Using try/except to check if origin_as_datetime can be converted

I am not sure if I am preserving the *args or **kargs correctly. Can you take a look?

The following test code works:
```python
import chainladder as cl
raa = cl.load_sample("raa")
cl.Chainladder().fit(raa).ultimate_.to_frame(origin_as_datetime=False)
cl.Chainladder().fit(raa).ultimate_.to_frame(origin_as_datetime=True) # no warning thrown
cl.Chainladder().fit(raa).cdf_.to_frame(origin_as_datetime=False)
cl.Chainladder().fit(raa).cdf_.to_frame(origin_as_datetime=True) # warning thrown
```

### DIFF
--- a/chainladder/core/pandas.py
+++ b/chainladder/core/pandas.py
@@ -7,6 +7,11 @@ import warnings
 from sklearn.utils import deprecated
 from chainladder.utils.utility_functions import num_to_nan
 
+try:
+    import dask.bag as db
+except:
+    db = None
+
 
 class TriangleGroupBy:
     def __init__(self, obj, by, axis=0, **kwargs):
@@ -60,7 +65,6 @@ class TrianglePandas:
         """
 
         if origin_as_datetime == None:  # origin_as_datetime not specified
-
             warning = "In an upcoming version of the package, `origin_as_datetime` " + \
                 "will be defaulted to `True` in to_frame(...), use " + \
                 "`origin_as_datetime=False` to preserve current setting."
--- a/chainladder/core/pandas.py
+++ b/chainladder/core/pandas.py
@@ -7,6 +7,7 @@ import warnings
 from sklearn.utils import deprecated
 
 from chainladder.utils.utility_functions import num_to_nan
+
 try:
     import dask.bag as db
 except:
@@ -18,8 +19,8 @@ class TriangleGroupBy:
         self.obj = obj.copy()
         self.axis = self.obj._get_axis(axis)
         self.by = by
-        if kwargs.get('groups', None):
-            self.groups = kwargs.get('groups', None)
+        if kwargs.get("groups", None):
+            self.groups = kwargs.get("groups", None)
         elif self.axis == 0:
             self.groups = obj.index.groupby(by)
         elif axis == 1:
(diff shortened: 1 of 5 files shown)


jbogaardt:
try/except are normally reserved for error handling beyond our control. For example, if a database connection can be established.  They generally shouldn't be used to manage control flow because it makes readibility a bit harder. Truth be told, this function has poor readibility to begin with and that's on me. If we can afford to make it more readible, the better.

In this case, it seems simple enough to just check whether the origin is a Period.

```python
import chainladder as cl
import pandas as pd

tri = cl.load_sample('raa')
cdf = cl.Development().fit(tri).cdf_

isinstance(tri.origin, pd.PeriodIndex) # True, so can be converted
isinstance(cdf.origin, pd.PeriodIndex) # False, so should not be converted
```

Alternatively, we can just use the handling in from pandas which will ignore the conversion if it encounters an error.
```python
pd.to_datetime(cdf.origin, errors='ignore')
```

jbogaardt (dismissed):

kennethshsu:
Hold on this is not done yet... lol. I still need to put in your changes. And also attempting to refactor the `to_frame()` method.

kennethshsu:
I think it works now, thanks for the feedback!

What ideas do you have on how to refactor the `.to_frame()` method?

jbogaardt:
There are errors:
E       NameError: name 'db' is not defined

kennethshsu:
I need to learn get in the habit of running pytest before every commit, especially PR.

Do you know how run pytests for errors? Look at `test_display.py` (is this the best place to put it?), it's not kicking out on line 18?

kennethshsu:
I need to learn to better get in the habit of running pytest before every commit, especially a PR.

Do you know how run pytests for errors? Look at `test_display.py` (is this the best place to put it?), it's not kicking out on line 18? Will need to look at this a bit more.

jbogaardt:
You can run pytest in your `cl_dev` environment. From the project root folder you'd type `pytest chainladder` at the command line. If that doesn't work, your dev environment might not be configured correctly.

kennethshsu:
I think just `pytest` by itself works.

I am having some trouble with tests that depend on the R package though. I wonder if there's an easy way to clean that up.

jbogaardt:
You can bypass the R tests with: 
`pytest chainladder -m "not r"`

kennethshsu:
Hmm ok, I think that allowed me to skip all of the tests that depends on the R package.

Did you get a chance to look at the PR? It should work now.

jbogaardt:
I did. Looks like the PR makes to_json tests fail. I could probably handle if you need, but thought you intended to clear those out

kennethshsu:
I don't think I saw that one. Let me take a look first.

kennethshsu:
Actully I don't think I understand the error. How are the tests setup? There's a `normal_run` then a `sparse_only_run`?

I ran this piece of code but I didn't get an error.
```python
a = cl.read_json(cl.Chainladder().fit_predict(raa).to_json()).full_triangle_
b = cl.Chainladder().fit_predict(raa).full_triangle_
assert abs(a - b).max().max() < 1e-4
```

There was a warning though. `RuntimeWarning: All-NaN slice encountered obj.values = func(obj.values, axis=axis, *args, **kwargs)` What does this mean?

kennethshsu:
In addition, are we missing an `assert` in the `test_json_subtri()` function?

Issue #304, opened 2022-06-17
Issue on page /tutorials/tail-tutorial.html


henrydingliu:
The code block under Chaining Multiple Transformers is slightly confusing. 

print("This doesn't pass") under the first try block should be a comment or be in the except block

print("This doesn't fail") under the second else block should be a comment or be in an else block.

kennethshsu:
I think they make sense the way they are laid out right now. Why do you think otherwise? If the print statements aren't printed to the console, that means that the code blocks weren't executed.

```python
print("First attempt:")
try:
    cl.TailCurve().fit(cl.Development().fit(quarterly))
    print("This doesn't pass")
except:
    print("This fails because we did not transform the triangle")

print("\nSecond attempt:")
try:
    cl.TailCurve().fit(cl.Development().fit_transform(quarterly))
    print("This passes because we transformed the triangle")
except:
    print("This doesn't fail")
```
```
First attempt:
This fails because we did not transform the triangle

Second attempt:
This passes because we transformed the triangle
```

henrydingliu:
The code executes fine. It's just the inherent paradox in having this doesn't pass in a try block and this doesn't fail in an except block.

kennethshsu:
Ahh I see what you are saying now.

What if we do something like this?
```python
print("First attempt:")
try:
    cl.TailCurve().fit(cl.Development().fit(quarterly))
    print("This works.")
except:
    print("This fails.")

print("\nSecond attempt:")
try:
    cl.TailCurve().fit(cl.Development().fit_transform(quarterly))
    print("This works.")
except:
    print("This fails.")
```

I'll move the commentary to the body instead of the code. Do you think this is clearer?

henrydingliu:
I think we can also add in the "This fails because we did not transform the triangle" and "This passes because we transformed the triangle" as comments

kennethshsu:
Ahh ok, I think I get what you are saying now. Thank you. I'll get this in.

Issue #305, opened 2022-06-18
Issue on page /tutorials/development-tutorial.html


henrydingliu:
The ValuationCorrelation class implements Appendix H from Mack 93, which tests "assumption (4) of the independence of the accident years".

The DevelopmentCorrelation class implements Appendix G from Mack 93, which tests "assumption (1) of the chain ladder method implies that subsequent development factors are not correlated". 

Essentially, ValuationCorrelation tests across calendar periods (diagonals), and by extension, origin periods (rows). DevelopmentCorrelation tests across maturities (columns). 

This section of the tutorial contain some confusing wordings regarding the delineation between these two tests.

kennethshsu:
Thanks, I cleaned up the commentary and the language used should be easier to read now.

Issue #299, opened 2022-05-23
Issue on page /tutorials/development-tutorial.html


henrydingliu:
Is Mack 1997 the same as Mack 1999 in the bibliography?

jbogaardt:
I'm not sure, there is a link to "Mack 1997" in that tutorial that resolves to this:https://www.casact.org/sites/default/files/2021-03/7_Mack_2000.pdf. I assume that's what any commentary is referencing.

henrydingliu:
ah, I thought that was a mislink since the year is 2000. Thanks!

henrydingliu:
Had a chance to read through https://www.casact.org/sites/default/files/2021-03/7_Mack_2000.pdf. It does not appear to be the correct paper.

kennethshsu:
Hmm, I can remove the reference. How many Mack papers are there? 🤔

henrydingliu:
In total I've seen 5. 2 on this page. 3 in the bibliography. However Mack 93 on this page is the same as the Mack 94 in the bibliography. And the Mack 97 on this page is linking to a Mack 2000. Mack 2000 should be listed separately because it's the academic source for Benktander. 

I think the following would address everything:

1. Change Mack 93 from this page to Mack 94 (just the text. the link is fine.)
2. Find the right Mack 97 and change the link of Mack 97 on this page to that
3. List Mack 2000 (from the Mack 97 link) separately in the bibliography

kennethshsu:
Is [this](https://www.actuaries.org.uk/system/files/documents/pdf/crm2-D6.pdf) the right Mack 1997 paper?

henrydingliu:
This looks like the Mack 94 paper (the paper that this page refers to as Mack 93)

kennethshsu:
You think so? On the top left it says 1997.

I found [this paper](https://actuaries.asn.au/Library/accomp04papergerigk.pdf), at the bottom it referenced many of Mack's (previous?) papers.

henrydingliu:
I see the bibliography at the end of the second paper.

[1994] Measuring the variability of chai n–ladder reserve estimates. In: CA S
Forum Spring 1994, pp. 101–182.
[1997] Measuring the variabi lity of chain–ladder reserve estimates. In: Claim s
Reserving Manual, vol. 2. London: Institute of Actuaries.

I guess this author wouldn't have listed it twice unless they papers were different somehow? But I'm not seeing how they are different
[...]

PR #307, opened 2022-06-21
addressed #304, 299, and 305

kennethshsu:


kennethshsu:
Revised commentary and removed reference to specific Mack papers per #299

kennethshsu:
Also revised the commentary per #305

jbogaardt (approved):

kennethshsu:
@jbogaardt there's an error `error: numpy 1.22.3 is installed but numpy<1.22,>=1.18 is required by {'numba'}`. Do you know what we need to do to fix this?

jbogaardt:
Seems to be in the [Unit Tests](https://github.com/casact/chainladder-python/blob/master/ci/environment-test.yaml) action but does not afffect the [CI Upstream](https://github.com/casact/chainladder-python/blob/master/ci/environment-latest.yaml) action. It is not clear straight away to me why only one of these environments would cause that They look similar with respect to numpy.  

Given no code changes to the library in this PR, I would recommend merging as is.

jbogaardt:
Interesting, docs failed because of:
```
sphinx.errors.ExtensionError: Could not import extension sphinx_panels (exception: No module named 'sphinx_panels')

Extension error:
Could not import extension sphinx_panels (exception: No module named 'sphinx_panels')
```
This was never explicitly declared in our [docs conda environment](https://github.com/casact/chainladder-python/blob/master/environment-docs.yaml). I'll add it and see if it resolves the docs.

jbogaardt:
Docs still failing because notebooks were executed and committed under the cl_dev environment. This gets hardcoded into the notebook JSON and readthedocs doesn't build that environment. It uses cl_docs, complains that cl_dev doesn't exist and crashes the build.

Trying to figure out how to get readthedocs to override the conda environment embedded in the notebook json. Otherwise, we may need to recommit the notebooks with them using the cl_docs kernel.

kennethshsu:
Hmm, I think I have been doing it using the `cl_dev` environment, but let me submit another PR using the stable kernel.

Issue #343, opened 2022-10-13
Mack development correlation test: t attribute only holds one T_k value


genedan:
According to Mack 1997, we calculate I - 3 T_k values, that is, the number of development periods minus 1 for the first, and minus 2 for the last two:

![image](https://user-images.githubusercontent.com/6165650/195719265-a23b0bf0-03b5-4462-9a0e-010fdc220a86.png)

The current version of [DevelopmentCorrelation](https://github.com/casact/chainladder-python/blob/046b9de9280147de562afa617c682454477b85f7/chainladder/core/correlation.py#L10) correctly calculates these values:

![image](https://user-images.githubusercontent.com/6165650/195719408-98eddfe0-8254-420e-9568-c47e15f02c0c.png)

but the subsequent assignment to the self.t attribute excludes all but the first, and the axes labels are called "Case Incurred" and "Total" which should really be something like "k" and "T_k" instead.

I have a PR in the works for this, will update you shortly.

genedan:
closed via [30db99c](https://github.com/genedan/chainladder-python/commit/30db99c851e5f03f351147bccef89cb0b6c066e8)

PR #344, opened 2022-10-14
Refine Mack DevelopmentCorrelation class

genedan:
Mostly addresses #343, changes include:

1. Added annotations to describe algorithm
2. Changed default value in docstring from .1 to .5 to be consistent with class signature and Mack 97 paper
3. Fixed attribute self.t to include all T_k values
4. Changed weights from internal variable to class attribute
5. Added attribute for hypothesis reject/non-reject as a bool


jbogaardt (approved):

jbogaardt:
These are great improvements @genedan. I like how you code. It has a software-development focus that should help the long term stability of the project.

Issue #332, opened 2022-08-28
Unpredictable behaviour of drop_high


henrydingliu:
drop_high does some weird stuff sometimes

```
prism_df = pd.read_csv(
    "https://raw.githubusercontent.com/casact/chainladder-python/master/chainladder/utils/data/prism.csv"
)
prism_df["AccidentDate"] = pd.to_datetime(prism_df["AccidentDate"])
prism_df["PaymentDate"] = pd.to_datetime(prism_df["PaymentDate"])
prism_df.loc[(prism_df["AccidentDate"].dt.year > 2008) & (prism_df["PaymentDate"].dt.year > 2012),"Incurred"] = 0
prism = cl.Triangle(
    data=prism_df,
    origin="AccidentDate",
    development="PaymentDate",
    columns="Incurred",
    cumulative=False,
)
prism = prism.grain("OYDY").incr_to_cum()
prism_dev = cl.Development(drop_high = 1, drop_low = 1).fit_transform(prism)
print(prism.age_to_age)
print(prism_dev.age_to_age)
```

drop_high seems to not work on 12-24, 24-36, 36-48, and 48-60,

kennethshsu:
Ahh, this is because it's not a "triangle", it's a trapezoid in this case. This is indeed a bug. In fact, `drop_low` doesn't work property as well.

kennethshsu:
@henrydingliu, did you resolve this? I can take a look at this one.

henrydingliu:
i believe so, yes

kennethshsu:
I don't think I see it working correctly, can you check again, even just using the code above?

henrydingliu:
it didn't work for me when i did pip install chainladder --pre on my existing vm. however, when i created a brand new vm, directly cloned https://github.com/casact/chainladder-python.git, and pip installed locally, it seemed to work. nb attached. 

[Untitled.zip](https://github.com/casact/chainladder-python/files/10028609/Untitled.zip)

henrydingliu:
tried pip install chainladder --pre again after uninstalling and clearing cache. still installing 0.8.13

kennethshsu:
My mistake, I was in the wrong environment and got confused. I see the changes now.

While I was playing with it, I can think of another improvement for this. Consider a scenario, where `drop_low = 1` is set, but ther are multple periods with the same LDF rank, which one should be dropped? For example:

AY 2015 $200 -> $200
AY 2016 $500 -> $500

[...]

PR #349, opened 2022-11-01
Updating drop high logic

henrydingliu:
This is a fix for #332. the size check for number of valid AtA factors is changed to directly count the number of non-nans, this comes up on excess lines where the first couple of development periods may sporadically have no claims.


jbogaardt (approved):

jbogaardt:
🎉🎉🎉 Thank you @henrydingliu! Glad to have you as a contributor. 🎉🎉🎉

Issue #346, opened 2022-10-30
Friedland data sets


genedan:
I have put together a collection of csv files from the Friedland paper, and thought I should add them here. I don't have all of them yet, but will be committing them as I go through the paper.

I can also submit the main example triangle from the Mack 97 paper.

PR #353, opened 2022-11-02
Add Friedland gl and xyz disposal rate data sets.

genedan:
Partially addresses #346.


jbogaardt:
Thanks @genedan. Quite a day, 3 PRs from 3 different contributors - it's really exciting to see this community grow.

On this PR in particular,
For the datasets to be packaged with the releases, they have to be registered in [ MANIFEST.in](https://github.com/casact/chainladder-python/blob/master/MANIFEST.in)

I haven't actually tried using [load_sample](https://github.com/casact/chainladder-python/blob/master/chainladder/utils/utility_functions.py#L17) with these new datasets, but I do wonder if it works. Its a horribly written function from my earlier days that has a lot of hardcodes of column names to simply call the `Triangle` constructor on pre-packaged data. It makes me cringe looking at it, but it may block users from accessing the triangles.  

We now have quite a library of sample data, it might even be a good idea to make them discoverable somehow from python itself. Open to ideas on this. sklearn makes each dataset a discoverable function on the datasets module:
```python
from sklearn import datasets
datasets.load_iris() # accessible through tab-complete
```
Very different to our setup, maybe some sort of `list_samples` function might solve this without causing too many breaking changes.

Issue #321, opened 2022-08-10
Incorrect Cape Cod reserve when incurred loss = 0


cdietrich215:
I noticed a minor issue when reserving using the Cape Cod. When incurred or paid loss is $0 for an accident period, the Cape Cod method will predict $0 of Ultimate Loss, when theoretically there should still be a reserve > $0. This may matter for some smaller insurers.

The BF and Benktander methods in this package both correctly set ultimate losses to be (a priori) * (sample weight) * (1 - 1/CDF) in these situations. I believe that the Cape Cod method should do the same. This method currently sets IBNR to 0 (or null).

jbogaardt:
Thanks. I didn't know about this error. 

Reproducible example:
```python
import chainladder as cl

triangle = cl.load_sample('raa')
premium = triangle.latest_diagonal * 0 + 50000
triangle.loc[:,:,'1987',48] = 0
cl.CapeCod().fit(triangle, sample_weight=premium).ultimate_
```

kennethshsu:
Can close the ticket once #368 is merged.

jbogaardt:
Reopening until release of 0.8.14

PR #368, opened 2022-11-17
addressed #321

kennethshsu:
Addressed #321, problem was in the `trend_array` near [line 159](https://github.com/casact/chainladder-python/blob/master/chainladder/methods/capecod.py#L159).

### DIFF
--- a/chainladder/methods/benktander.py
+++ b/chainladder/methods/benktander.py
@@ -4,8 +4,9 @@
 from chainladder.methods import MethodBase
 import numpy as np
 
+
 class Benktander(MethodBase):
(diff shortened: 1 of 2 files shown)


jbogaardt:
This PR seems to break a few existing unit tests.

jbogaardt:
Interestingly, there is one unrelated failure. It seems to be from a recent update to `pandas==1.5.1` used by github actions.
Since it's unrelated, I'll merge,  but will need to address the pandas issue soon.

jbogaardt (approved):

kennethshsu:
I saw that too since I didn't get an error locally, I'll take a look shortly.

PR #370, opened 2022-11-18
Adding 4D support to drop_hi/lo

henrydingliu:
removed for loop. changed all 2D references to 4D. 

note: this was only done for _drop_n. similar work needs to be done for _drop_x


jbogaardt (approved):

jbogaardt:
thank you @henrydingliu!

PR #384, opened 2022-11-25
add semi-annual key

wleescor:
Semi-annual option not available, causing errors in chainladder fits for semi-annual triangles


jbogaardt (approved):
Welome @wleescor! This is great. Thank you for this contribution.

Issue #386, opened 2022-11-27
weight_ and link_ratio appears broken


henrydingliu:
I think 4D support for drop_hi/lo etc broke link_ratio and weight_. 

```python
clrd = cl.load_sample('clrd').groupby("LOB").sum()[["CumPaidLoss","IncurLoss"]]
#this is working
print(cl.Development(n_periods = 4).fit_transform(clrd.iloc[0,0]).link_ratio)
#this is not
print(cl.Development(n_periods = 4).fit_transform(clrd).link_ratio.iloc[0,0])
try:
    cl.Development(n_periods = 4).fit(clrd).weight_
except:
    print("this errored")
```

henrydingliu:
@jbogaardt while `index=self.ldf_.origin,` isn't the issue causing the error, but it doesn't seem to be working either. is there any existing attribute in Development that would have the full set origins? otherwise we'd have to make w_ into a triangle during fitting?

```python
    @property
    def weight_(self):
        return pd.DataFrame(
            self.w_[0, 0],
            index=self.ldf_.origin,
            columns=list(self.ldf_.development.values[:, 0]),
        )
```

jbogaardt:
Thanks for finding and fixing this. We should just remove `weight_`. It serves no purpose other than making `w_` display prettier as a pandas dataframe. Nothing downstream relies on it.

jbogaardt:
I pushed one more small fix on this.

PR #387, opened 2022-11-27
Fixing link_ratio

henrydingliu:
removing a redundant slicer in link_ratio that causes w_ to be ignored whenever there are more than 1 triangle columns


henrydingliu:
half of the fit for #386

jbogaardt (approved):

Issue #376, opened 2022-11-19
[ENH] Adding Pipeline support to Development


henrydingliu:
**Is your feature request related to a problem? Please describe.**
within the default development class, ata selections (drop, n_period, etc.) is performed via a fixed order. custom order is not possible since the transformer currently do not pass on the previous ata selections (e.g. Development(n_period = 4).fit_transform(Development(drop_low = 1).fit_transform(tri) is not equal to Development(n_period = 4,drop_low = 1).fit_transform(tri))

**Is your feature request at odds with the scope of the package?**
this feature is strictly for reserving and would offer greater alignment with the sklearn API

**Describe the solution you'd like**
have the transformer pass on the previous ata selection

**Describe alternatives you've considered**
parameterize orders. messy

**Additional context**
probably clean up ata selection logic in the meantime

henrydingliu:
@jbogaardt from my understanding, it looks like just adding self.w_ to the product [here](https://github.com/casact/chainladder-python/blob/23ee15f2bd276f6e99c03ddc6d4f77b06890258f/chainladder/development/development.py#L166) would work? probably not thinking through all the other side effects.

jbogaardt:
I can't think of any harm from that.  Assuming `hasattr(X, 'w_') is True`, it may very well be all that is needed.

henrydingliu:
do you think there's value to standardizing all the n_period and drop functions? right now some are passing parameters, some are not, etc. what if they are all retrieve parameters from self and directly modify self.w_?

jbogaardt:
That makes sense. Passing arguments around that are already bound to `self` seems improper form.

henrydingliu:
after some more tinkering, as well as reflecting on how n_period and drops are being used [elsewhere](https://github.com/casact/chainladder-python/blob/3088fcdcf87226d85c117bc2f905abec95a2988f/chainladder/development/incremental.py#L118), i take back what i said about w_. here is where my head's now at. 

It might make sense to add a function set_weights(factor, secondary_rank, previous_weight, n_periods, drop_high...) under development base, intentionally skipping the leading _ and passing in self. Then all development classes can use this function to drop undesirable factors (the current development.fit approach can bug. see snippet below), regardless of what kind of factor (Munich CL can drop paid-to-incurred or residuals; Outstanding can drop paid-to-reserve or reserve-to-reserve factors). 

```python
genins = cl.load_sample("genins")
genins_dev = cl.Development().fit_transform(genins)
[...]

PR #397, opened 2022-11-30
next step for #376

henrydingliu:
adding w_v2_ in development to further beta _set_weight_func. also adding pipeline support for w_v2_ as well

switching incremental to using _set_weight_func with pipeline support

tests added for both


jbogaardt:
Beautiful. This is a model PR.

* It is tied to a specific issue (#376)
* It has 100% pass rate on tests
* It introduces new tests for new functionality

PR #403, opened 2022-12-07
Expanded mappings to include semester sampling

A108669:
Closes #402

PR #410, opened 2022-12-18
fix one more ref to semiannual

wleescor:
was going to submit similar to pull request #403 but got beaten to it. seemed to be missing one more file still


jbogaardt (approved):

jbogaardt:
Thank you again @wleescor!

Issue #438, opened 2023-06-15
Wrong claims values after 24 development periods [BUG]


andrejakobsen:
**Describe the bug**
While using the package on monthly data, I noticed that the second origin row was missing claims values for the last development period. It looked like one value was missing and the rest got offset. Upon closer inspection, the 25th development period gets set equal to the sum of the values at development periods 23 and 25, where 25 gets the value of 26 and 27 gets the value at 28, etc. So after 24 development periods the values are offset by 1 period. This also occurs for the first origin period but at development period 25. However, all origin periods after the second do not seem to have this issue.

**To Reproduce**
Steps to reproduce the behavior. Code should be self-contained and runnable against publicly available data. Here is a test file that makes it easy to see the issue: [test_triangle.csv](https://github.com/casact/chainladder-python/files/11759656/test_triangle.csv)
```python
import chainladder as cl
import pandas as pd
test_triangle = pd.read_csv("test_triangle.csv")
triangle = cl.Triangle(
    test_triangle,
    origin="origin",
    eevelopment="development",
    columns="claims",
    cumulative=True
)
```

**Expected behavior**
A clear and concise description of what you expected to happen. If it can be expressed in code, then that is better. The complete code can serve as a unit test.
```python
assert triangle.loc[..., "claims", "2018-01", 25] == 25
```

**Desktop (please complete the following information):**
 - Numpy 1.24.3
 - Pandas  2.0.2
 - Chainladder 0.8.15

jbogaardt:
Thanks, I haven't yet run our test suite against pandas>=2.0. Sounds like there are breaking changes that need to be addressed. This issue gives me a good place to start looking.

kennethshsu:
@andrejakobsen welcome!

I've been trying to figure out what's going on, but as a temporary solution, you can do something like 
```python
.dev_to_val().val_to_dev()
```

This will at least gets you back to the correct age periods. Will continue to investigate later today...

PR #440, opened 2023-06-15
#438 bug

kennethshsu:
Addressing #438


kennethshsu:
@jbogaardt can you help resolve some of the unit tests regarding installation?

Also, I don't think my solution will work, it will break when the valuations are less granular than annual. Looking deeper.

jbogaardt (approved):

jbogaardt:
it looks like pandas no longer supports TimeDelta of Month as it is ambiguous. It can be anywhere between 28 and 31 days.  

Numpy treats a month as 30 days regardless, and this is causing conversion from days to months to fail:
```python
imprt numpy as np
np.timedelta64(1, 'M').astype('timedelta64[D]')
```
I've played with using (365/12) as the monthly value and it seems to restore the correct ages.

kennethshsu:
Right, I think the only way to get around it is to actually calculate the development age using year and month manually. Using anything from `pandas` or `numpy` will force the calculation to be too accurate, and we'll then need to be forced to round and result in possible error.

PR #441, opened 2023-06-16
add max version of pandas in requirements.txt

andrejakobsen:
As discussed in https://github.com/casact/chainladder-python/issues/438, there are currently breaking changes with Pandas 2.0.


jbogaardt:
Thanks for the PR @andrejakobsen . I'm also hopeful we can patch this weekend to support pandas 2.0 and get a quick release going.

Issue #474, opened 2023-11-23
[ENH] Support polars via Python dataframe interchange protocol 


lorentzenchr:
**Is your feature request related to a problem? Please describe.**
I would like to preprocess my data in other data containers than pandas, e.g. pyarrow and polars, and then apply `Triangle` to it.

**Is your feature request at odds with the scope of the package?**
```python
import chainladder as cl
import pandas as pd
import polars as pl


raa_df = pd.read_csv(
    "https://raw.githubusercontent.com/casact/chainladder-python/master/chainladder/utils/data/raa.csv"
)

# This works just fine
raa = cl.Triangle(
    raa_df,
    origin="origin",
    development="development",
    columns="values",
    cumulative=True,
)

# This gives an error
raa_pl = pl.read_csv(
    "https://raw.githubusercontent.com/casact/chainladder-python/master/chainladder/utils/data/raa.csv"
)

cl.Triangle(
    raa_pl,
    origin="origin",
    development="development",
    columns="values",
    cumulative=True,
)
```
results in
```
AttributeError: 'DataFrame' object has no attribute 'iloc'
```

**Describe the solution you'd like**
Supporting data via the [Python dataframe interchange protocol](https://data-apis.org/dataframe-protocol/latest/index.html) might be an optimal approach.

**Describe alternatives you've considered**
One could also consider polars specific code path. But this might result in more maintenance burden: Where to stop? Add pyarrow, too?

**Additional context**
A similar feature request is https://github.com/scikit-learn/scikit-learn/issues/25896 which is currently worked on.

johalnes:
Do you think something like [this in Plotly](https://github.com/plotly/plotly.py/blob/99e63da4df59a6cad4e7ea42583f292150db9e18/packages/python/plotly/plotly/express/_core.py#L1328), just to get a zero-copy to pandas @lorentzenchr? 

I see you have an Polars rewrite as a branch @jbogaardt - is converting to polars what you consider the future of the package? 😄

jbogaardt:
For some reason i didnt see this issue when @lorentzenchr created it. It is an worthwhile enhacement since thats the direction other tools seem to be going.  

[...]

PR #482, opened 2023-12-15
Support for Polars and other dataframes

johalnes:
# Overview
As discussed in #474, added [Python dataframe interchange protocol](https://data-apis.org/dataframe-protocol/latest/index.html). 

This open up the possibility for using the chain ladder package with multiple different dataframe packages, like Spark, Polars, Vaex etc. 

Behind the scene it uses the "new" pandas function  `pd.api.interchange.from_dataframe(data)` that magically converts a dataframe to the correct types and other dataframe adjustments to make it pandas compatible. 

## Testing
Tested with the example from issue #474 and added test to ensure same result with polars and pandas

### DIFF
--- a/chainladder/core/base.py
+++ b/chainladder/core/base.py
@@ -410,11 +410,14 @@ class TriangleBase(
         Convert an object supporting the __dataframe__ protocol to a pandas DataFrame.
         Requires pandas version > 1.5.2.
 
-        Parameters:
-        data: Dataframe object supporting the __dataframe__ protocol.
-
-        Returns:
-        pd.DataFrame: A pandas DataFrame.
+        Parameters
+        ----------
+        data :
+            Dataframe object supporting the __dataframe__ protocol.
+        Returns
+        -------
+        out: pd.DataFrame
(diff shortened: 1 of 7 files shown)


johalnes:
I'm using Ruff for linting and formatting, which complains about some best practices not being followed. Mostly unused imports, isort complains about sorting of imports, but also that `type(value) is str` should be changed with is `instance(value, str)` according to some PEP.

Should I create a new issue and pull request with ruff as formatter and use auto fix to remove those smaller "issues"?

jbogaardt (approved):

jbogaardt:
Thanks for these updates @johalnes !

jbogaardt:
> I'm using Ruff for linting and formatting, which complains about some best practices not being followed. Mostly unused imports, isort complains about sorting of imports, but also that `type(value) is str` should be changed with is `instance(value, str)` according to some PEP.
> 
> Should I create a new issue and pull request with ruff as formatter and use auto fix to remove those smaller "issues"?

Sure!

PR #488, opened 2024-01-18
fix broken image links

MatthewCaseres:
I couldn't figure out where `/auto_examples/images/sphx_glr_plot_capecod_001.png` is supposed to be in the repo, but that link is broken so I added a TODO

edit: not sure what is up with the diff on the triangle.ipynb, seems to be a formatting difference. editing in codespaces so maybe like a different version of vscode than previously used idk.


jbogaardt (approved):

jbogaardt:
Thank you yet again @MatthewCaseres!

Issue #494, opened 2024-03-07
[BUG] Unable to convert between grains when using pandas==2.2.0


00milsg:
**Describe the bug**

When converting a triangle with grain OMDM to OQDQ every cell is converted to nan. 

**To Reproduce**

To reproduce the behaviour I used the following package versions. *Note: When I downgraded pandas to a version less than 2.2.0 the issue disappears. I will use this as a workaround for now.*

```
pandas==2.2.0
numpy==1.26.4
chainladder==0.8.18
```

Full code to reproduce the behaviour is attached in a text file (it's 50 lines long and I don't want to bloat this post). I used the prism example dataset and back worked a dataset to create monthly & quarterly triangles. The code will run self contained. See link below. 

If you'd rather I add the example code as a comment then let me know.

[chain-ladder-grain-conversion-error-reprex.txt](https://github.com/casact/chainladder-python/files/14524554/chain-ladder-grain-conversion-error-reprex.txt)

**Results of my digging into the issue**

I think this is being caused by the initial dev_to_val conversion at the top of the method (possibly related to this [issue](https://github.com/casact/chainladder-python/issues/480)?). This then flows through and upsets the calculation of the d_start variable. In the earlier versions the calculation of the d_start variable is correct. In version >= 2.2.0 d_start gets set to the start of the previous month. This triggers the conditional block below, which sets the data to nan. 

https://github.com/casact/chainladder-python/blob/0a0d0cb9fb2db4743cf5efd8f6e34cdf5635b24b/chainladder/core/triangle.py#L689-L702

**Question:** Could we add a warning stating that this is what's happening (and potentially why)? We could also add an accompanying suggestion on how to correct it (filter your input data so that min(dev_month) >= min(orig_month)).

**Expected behavior**

When converting a triangle from grain OMDM to OQDQ the total sum of the triangle contents should remain the same. Please see below for a potential unit test.

```
# Test that triangles contain the same starting information
qtr_sum = qtr_triangle['reportedCount'].sum().sum()
mth_sum = mth_triangle['reportedCount'].sum().sum()
assert qtr_sum == mth_sum, "Triangles are not equal before grain change."

# Test that we still get the same answers when we convert the monthly tri to quarterly
mth_conv_sum = mth_triangle['reportedCount'].grain('OQDQ').sum().sum()
assert mth_conv_sum == mth_sum, "Triangles are not equal after grain change."
```

jbogaardt:
[...]

Issue #491, opened 2024-01-21
Empty triangle cannot be printed [BUG]


MatthewCaseres:
**Describe the bug**
I cannot print an empty triangle.

**To Reproduce**
```python
import chainladder as cl
print(cl.Triangle())
```

error is
```
>>> print(cl.Triangle()
... )
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/matthewcaseres/Documents/GitHub/chainladder-python/chainladder/core/display.py", line 16, in __repr__
    if (self.values.shape[0], self.values.shape[1]) == (1, 1):
```

**Expected behavior**
The empty triangle should print

kennethshsu:
Now this

```python
import chainladder as cl
empty_tri = cl.Triangle()
empty_tri
```

Returns `Triangle is empty.`.

PR #501, opened 2024-03-26
497

kennethshsu:
- Fixed a calculation bug in #497 
- Introduced `approximation_grain` in `chainladder.ParallelogramOLF()`
- Added two test cases as discussed in #497


kennethshsu:
@jbogaardt, are you familiar with these errors? I looked at it in detail and I don't think I understand what is going on...

```python
=========================== short test summary info ============================
ERROR chainladder/methods/tests/test_predict.py - ValueError: Fancy indexing on origin/development is not supported.
ERROR chainladder/workflow/tests/test_predict.py - ValueError: Fancy indexing on origin/development is not supported.
!!!!!!!!!!!!!!!!!!! Interrupted: 2 errors during collection !!!!!!!!!!!!!!!!!!!!
```

Is this [line](https://github.com/casact/chainladder-python/blob/6c9ad1c6ae39926f96d9660a18d27959a2a36e1d/chainladder/methods/tests/test_predict.py#L5) causing the error? Why?

jbogaardt:
I'm not sure the specifics on the error from this PR, but the error stems from non-contiguous slices out of origin/development.

Here is a bare-bones example that is not supported by the package where we slice an arbitrary list of origins that are out of sequence:
```python
import chainladder as cl
clrd = cl.load_sample('clrd')
clrd.iloc[:, :, [2, 4, 3]]
```

jbogaardt:
It may be differences in dependencies. Github grabs the latest packages. It is using pandas==2.2.1 .


My suspicion is there are some differences in these versions that is causing it to fail. In issue #494, we have this problem with pandas>2.2.0. I think you just made the same issue visible in our unit tests. I've committed a fix since the pandas team gave a suggestion and also signaled they are unlikely to change the behavior. It seemed to clear those errors, but now has more pointed suggestions for the on-leveling updates.

```shell
FAILED chainladder/utils/tests/test_utilities.py::test_non_vertical_line - assert 0.1746221283369899 < 0.0
FAILED chainladder/utils/tests/test_utilities.py::test_vertical_line - UnboundLocalError: cannot access local variable 'cum_avg_rate_non_leaps' where it is not associated with a value
```

jbogaardt:
they aren't resolved, but they are display so I assumed they were related to package dependencies and not the PR itself. Looking closer, I see you have he following in display.py:
```python
        try:
            self.values
        except:
            print("Triangle is empty")
            return
```
I suspect this is causing it.  Can you let me know your intent with this logic?

kennethshsu:
Ahh, I think that's from another branch to address #491, it's not quite done... I think I messed up my version control while working on the two branches. Let me try to resolve this.

PR #420, opened 2023-03-02
Store origin label as an attribute

genedan:
Related to #416. I propose we keep the origin label metadata, this way the user can distinguish between different origin bases, such as accident year vs policy year, etc.

```
cl.load_sample('xyz')
                                       Triangle Summary
Valuation:                                      2008-12
Grain:                                             OYDY
Shape:                                   (1, 5, 11, 11)
Index:                                          [Total]
Origin:                                  [AccidentYear]
Columns:    [Incurred, Paid, Reported, Closed, Premium]
```

PR #507, opened 2024-04-05
Addressed #420 and #491

kennethshsu:
- Addressed #420 
- Addressed #491 
- Added `xyz` as test fixture
- Removed some debugger from a prior commit
- Added `polars-lts-cpu` into `environment-dev.yaml` since `polars` by itself is sometimes not supported on certain newer CPUs


jbogaardt (approved):

jbogaardt:
I'm surprised we have polars in the dev environment. I've been experimenting with it in lieu of sparse, but haven't hit a point where it should be in the main branch.

kennethshsu:
`polars` was already there I think, I only added the newer cpu version of it. 

From the very limited few hours that I spent messing with it, I like it a lot and prefer it over `sparse` too.

Issue #510, opened 2024-04-11
[MAINT] Pandas deprecation notices


jbogaardt:
**Is your feature request related to a problem? Please describe.**

There are several pandas deprecation notices with pandas 2.2 and later.  These should be handled and silenced. The most common is our usage of the `infer_datetime_format`.

```python
chainladder/core/base.py:250: UserWarning: The argument 'infer_datetime_format' is deprecated and will be removed in a future version. A strict version of it is now the default, see https://pandas.pydata.org/pdeps/0004-consistent-to-datetime-parsing.html. You can safely remove this argument.
```

**Is your feature request at odds with the scope of the package?**
No

**Describe the solution you'd like**
No pesky deprecation notices regardless of which version of pandas is installed.

kennethshsu:
@jbogaardt did you resolve this?

jbogaardt:
I have not, but it probably makes the most sense to explicitly set date formats of datasets within the package to silence it for training purposes, but otherwise persist the deprecation.

kennethshsu:
Hmm, can you elaborate more on what you mean? 

There's a PR on what I think fixes the issue, it's just using the version package to check for versions.

PR #511, opened 2024-04-24
#510

kennethshsu:
Addressed #510

### DIFF
--- a/chainladder/core/base.py
+++ b/chainladder/core/base.py
@@ -274,16 +274,16 @@ class TriangleBase(
         diffs = np.diff(np.sort(months))
         if len(dates.unique()) == 1:
             grain = (
-                "Y"
+                "A"
                 if version.Version(pd.__version__) >= version.Version("2.2.0")
-                else "A"
+                else "Y"
             )
 
         elif len(months) == 1:
             grain = (
-                "Y"
+                "A"
                 if version.Version(pd.__version__) >= version.Version("2.2.0")
-                else "A"
+                else "Y"
             )
 
         elif np.all(diffs == 6):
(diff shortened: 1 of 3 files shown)


kennethshsu:
@jbogaardt can you review this PR?

jbogaardt:
> @jbogaardt can you review this PR?

I thought I did. Did you not see my comments?  Pandas is deprecating "Y" and moving towards "A" for 2.2 and later. Refer to their [release note](https://pandas.pydata.org/docs/whatsnew/v2.2.0.html#deprecate-aliases-m-q-y-etc-in-favour-of-me-qe-ye-etc-for-offsets).  It looks like you're code change says to use the deprecated "Y" value for newer versions of pandas which seems wrong. 

There is also a new conditional which doesn't actually compare against a version number:
```python
"Y" if version.Version(pd.__version__) else "A"
```
Everything is "truthy" in python so this will always return "Y" regardless of the pandas version.

jbogaardt (approved):

kennethshsu:
I never saw those comments, but thanks for the feedback and approving the PR. :)

I think I'll keep the equal sign the way it is in the PR, since `2.2 and later` means `>= 2.2`.

kennethshsu:
@jbogaardt, can you double-check that `A` is preferred and that `Y` is being deprecated by `pandas`? I just checked the doc you linked and I don't see it.

In fact, I am still getting the warning:
```python
FutureWarning: 'A-DEC' is deprecated and will be removed in a future version, please use 'Y-DEC' instead.
```

jbogaardt:
> @jbogaardt, can you double-check that `A` is preferred and that `Y` is being deprecated by `pandas`? I just checked the doc you linked and I don't see it.
> 
> In fact, I am still getting the warning:
> 
> ```python
> FutureWarning: 'A-DEC' is deprecated and will be removed in a future version, please use 'Y-DEC' instead.
> ```

Thanks @kennethshsu , you're right. My brain is backwards. 'A' is being deprecated in favor of 'Y'.

Issue #517, opened 2024-05-17
Bugs in drop_high and drop_low


kennethshsu:
### Discussed in https://github.com/casact/chainladder-python/discussions/514

<div type='discussions-op-text'>

<sup>Originally posted by **graciepii** May 13, 2024</sup>
Hello, I am currently using the below line of code, which works completely as expected for grain OYDY triangles. 

dev = cl.Development(n_periods=-1, drop_high=False, drop_low=False).fit_transform(tri)

However when I use OYDM it displays the below error:

ValueError: all keys need to be the same shape

Could anyone help?</div>

Test:

```python
tri = cl.load_sample("prism")["Paid"].sum().grain("OYDQ")
no_drop = cl.Development().fit_transform(tri).cdf_.to_frame().values
drop_high = cl.Development(drop_high = True).fit_transform(tri).cdf_.to_frame().values
drop_low= cl.Development(drop_low = True).fit_transform(tri).cdf_.to_frame().values
assert (drop_low >= no_drop).all()
assert (no_drop >= drop_high).all()
```

PR #519, opened 2024-05-17
#514

kennethshsu:
- Addressed #514.
This is done by removing [secondary_rank](https://github.com/casact/chainladder-python/compare/%23514?expand=1#diff-e4811ac35addcd44905fb8e503aeac1de4816ff160955cd40efe4c3a82c0cc7fL520-L531).
@henrydingliu, do you remember why you did this? None of the tests failed, and I couldn't figure out what its supposed to do.
- Added test in #517.

### DIFF
--- a/chainladder/core/triangle.py
+++ b/chainladder/core/triangle.py
@@ -690,7 +690,7 @@ class Triangle(TriangleBase):
         obj = self.dev_to_val()
 
         if ograin_new != ograin_old:
-            freq = {"Y": "Y", "S": "2Q"}.get(ograin_new, ograin_new)
+            freq = {"Y": "YE", "S": "2Q"}.get(ograin_new, ograin_new)
 
             if trailing or (obj.origin.freqstr[-3:] != "DEC" and ograin_old != "M"):
                 origin_period_end = self.origin[-1].strftime("%b").upper()
--- a/chainladder/core/triangle.py
+++ b/chainladder/core/triangle.py
@@ -690,7 +690,7 @@ class Triangle(TriangleBase):
         obj = self.dev_to_val()
 
         if ograin_new != ograin_old:
-            freq = {"Y": "YE", "S": "2Q"}.get(ograin_new, ograin_new)
+            freq = {"Y": "Y", "S": "2Q"}.get(ograin_new, ograin_new)
 
             if trailing or (obj.origin.freqstr[-3:] != "DEC" and ograin_old != "M"):
                 origin_period_end = self.origin[-1].strftime("%b").upper()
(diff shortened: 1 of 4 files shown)


kennethshsu:
If this PR is approved, I think I will need to go back and remove all the dependent sub functions to not pass in that argument. We will need to modify the tests too. I'm sure Henry has a reason for this, but it's also been a while. I wonder if it was a way to get around some upstream bug.

henrydingliu:
someone had an issue where drop_hilo behaved unpredictably. on a column like this

1.2
1
1
1
1.1
1

if you do ex hilo and volume weighted, it wasn't obvious which 1 would be dropped. secondary rank would allow you to specify if you want to drop maybe the oldest 1 or the 1 with the smallest underlying loss. 

I'm surprise that removing secondary_rank would do anything. it looked to be an optional argument that defaults to blank.

kennethshsu:
I see, that makes a lot of sense now after reading the comments again. But this is not currently captured in any of the tests right? That's why this PR is still passing all tests.

Removing the argument does fix #514 because these [lines](https://github.com/casact/chainladder-python/blob/d21f12523d45441f8b300af2296b00f68b7618fd/chainladder/development/development.py#L139-L142) are causing problems. The primary is `obj.age_to_age` but the secondary is `obj.iloc[...,:-1,:-1]`. However, I believe we can only do this when the triangle is symmetrical (i.e. an isosceles right triangle, with origin and development having the same size?).

I propose that we accept this PR first, we need to write new tests and maybe start a discussion on how to handle all the conflict collision, especially when dealing with very special cases.

henrydingliu:
i don't think secondary_rank is causing the problem. it's optional. you can simply delete obj.iloc[...,:-1,:-1] from the failing line.

henrydingliu:
@jbogaardt do you recall what obj.iloc[...,:-1,:-1] was changed to elsewhere in the package to deal with scalene triangles?

jbogaardt (approved):

jbogaardt:
> @jbogaardt do you recall what obj.iloc[...,:-1,:-1] was changed to elsewhere in the package to deal with scalene triangles?

I'm not sure, but my instinct is to rely on the logic that produces the valid entries for a scalene triangle's link_ratios:
```python
cl.load_sample('quarterly').link_ratio.nan_triangle
```

henrydingliu:
Thanks @jbogaardt. Found it in the link_ratio code. 

@kennethshsu can you simply delete obj.iloc... from [here](https://github.com/casact/chainladder-python/blob/d21f12523d45441f8b300af2296b00f68b7618fd/chainladder/development/development.py#L139-L142)? That should fix the issue, as secondary_rank is optional. I'll find time this week to add back the right secondary rank.

kennethshsu:
Yap sounds good, thanks Henry, I think that's a good temporary solution for now. 

I added a couple more tests to hopefully catch this. Once `secondary_rank` properly ranks using the underlying triangle, we should be able to pass `test_new_drop_10`.

Do we need another round of code review? I know this request is technically already approved. But @jbogaardt, if it looks good feel free to merge!

Issue #542, opened 2024-11-05
[BUG] AttributeError: `np.NINF` was removed in the NumPy 2.0 release. Use `-np.inf` instead.


fornoj:
**Describe the bug**
As soon as I import the package chainladder, I get the error :
```AttributeError: `np.NINF` was removed in the NumPy 2.0 release. Use `-np.inf` instead.```

**To Reproduce**
```python
import chainladder as cl
```

**Expected behavior**
No error while importing

**Desktop (please complete the following information):**
 - python 3.12.7
 - Numpy Version 2.0.2
 - Pandas Version 2.2.3
 - Chainladder Version 0.8.23

jallali22:
Hi,
Thanks for the correction. My company retrieves packages from pypi.org. Would it be possible to have a version upgrade on pypi, please ? it seems that version 0.8.23 is still with this bugged

jbogaardt:
v0.8.24 pushed and resolves the numpy>2.0 issue.

PR #549, opened 2025-02-11
Limit Numpy Version `<2`

KSafran:
As mentioned in issue #542 there is at least one compatibility issue with numpy 2.x.x. Until that's fixed let's just set a max numpy version?


jbogaardt:
Thanks [KSafran](https://github.com/KSafran), I ended up pushing code fix for this, negating the need for updated requirements.txt. Thanks for your recommendation and making this hit the top of my radar.

Issue #553, opened 2025-03-15
[BUG] Berquist Sherman Settlement Rate adjustment errors on multi-triangles


cdietrich215:
**Describe the bug**
When running the Berquist-Sherman settlement rate adjustment on a triangle with an index (`kdims`) length >1, this should be equivalent to running the settlement rate adjustment on each "sub-triangle" individually, but this is not the case as shown below. I believe this is caused by the fact that we use the lookup values for the **first index** to look up the correct `a` and `b` parameters to use. The error occurs in lines [149](https://github.com/casact/chainladder-python/blob/master/chainladder/adjustments/berqsherm.py#L149) and [159](https://github.com/casact/chainladder-python/blob/master/chainladder/adjustments/berqsherm.py#L159) of the code. 

**To Reproduce**
```python
import chainladder as cl
lob = 'prodliab'

# Set up triangle with multiple index rows
triangle = cl.load_sample('clrd')[['IncurLoss','CumPaidLoss','EarnedPremDIR']]
triangle = triangle.groupby('LOB').sum()
triangle['Rept_Counts'] = triangle['CumPaidLoss'] * 0.1
triangle['Paid_Counts'] = triangle['Rept_Counts'] - 0.2 * (triangle['Rept_Counts'].latest_diagonal - triangle['Rept_Counts'])

triangle_prodliab_only = triangle.loc[lob]

# Berquist Sherman adjust on main triangle
berq_tri_all = cl.BerquistSherman(
    paid_amount='CumPaidLoss', incurred_amount='IncurLoss',
    reported_count='Rept_Counts', closed_count='Paid_Counts',
    trend=0.00).fit_transform(triangle)

# Berquist Sherman adjust on ppauto only
berq_tri_prodliab = cl.BerquistSherman(
    paid_amount='CumPaidLoss', incurred_amount='IncurLoss',
    reported_count='Rept_Counts', closed_count='Paid_Counts',
    trend=0.00).fit_transform(triangle_prodliab_only)

display(berq_tri_all['CumPaidLoss'].loc[lob])
display(berq_tri_prodliab['CumPaidLoss'])
```

**Expected behavior**
Ideally, running the Berquist-Sherman triangle with index length >1 should produce the same result on every "sub-triangle" as if the adjustment had been run on each sub-triangle individually. The "lookup" values for the settlement rate a and b parameters should be based on the corresponding sub-triangle

**Desktop:**
 - Numpy Version: 2.0.2
[...]

PR #555, opened 2025-03-19
553

cdietrich215:
This pull request resolves issue [553](https://github.com/casact/chainladder-python/issues/553), which notes that the Berquist Sherman settlement rate adjustment (in the `berqsherm.py` module) sometimes adjusts incorrectly for triangles with a kdims length >1

All tests pass after making the change, and the issue seems to be resolved. Let me know if there are any concerns.


jbogaardt (approved):

jbogaardt:
@cdietrich215 , sorry for the delay. This looks great. Thanks for making this package better!

PR #558, opened 2025-04-04
Add annotations to triangle.py

genedan:
Continuing with annotations and type hints as usual. There is one non-annotation change which is converting TriangleBase._get_date_axes() to a static method since it appears to be one, so I've decorated it accordingly.


genedan:
Hey @jbogaardt  and/or @kennethshsu, I've got a question about the following lines in triangle.py:

        # Set remaining triangle properties
        val_date = data_agg["__development__"].max()
        val_date = val_date.compute() if hasattr(val_date, "compute") else val_date
        self.key_labels = index
        self.valuation_date = val_date

What is the purpose of val_date.compute()? This would indicate to me that val_date could be of a type other than just a regular pandas Timestamp, which I don't think has that method. I'll update this section with a comment once I understand what it does.

jbogaardt:
If I recall, compute was meant for dark support which was never implemented. From my vantage point it probably shouldn't ever be implemented as there is a lot of better scalable dataframe tooling out there these days.

jbogaardt (approved):

genedan:
K. I did see other mentions of dask in other parts of the code. I'll make a comment on this line and maybe make a separate discussion about dask for another day if we want to do something about it and leave the line as-is.

genedan:
I'm getting a red warning in my IDE which would indicate breaking code:

image

The official warning is, `"Cannot find reference 'dmatrix' in '__init__.py'",` but I ran the line and everything imports just fine. Maybe it's a bug in PyCharm or something. Just making a note since I'm going to mark this line with a # noqa and we can come back to this comment if we ever do run into a problem with the import.

genedan:
I'm putting in one non-annotation change which concerns the following lines of TriangleBase.get_array_module():

image

The return type of this function should be ModuleType, but there is an IDE warning of the possibility of returning None type. I don't see that happening since we raise an error in that case, but I still think it would be good to eliminate code inspection warnings when we can so that future people looking through the code aren't overwhelmed with them, even if they are harmless. So, I'm changing this to:

image

genedan:
Hey @jbogaardt, I have a question on this annotating this function in utility_functions.py:

```
def num_to_value(arr: ArrayLike, value):
    """
    Function that turns all zeros to nan values in an array.
    """
    backend = arr.__class__.__module__.split(".")[0]
    if backend == "sparse":
        if arr.fill_value == 0 or sp.isnan(arr.fill_value):
            arr.coords = arr.coords[:, arr.data != 0]
            arr.data = arr.data[arr.data != 0]


            arr = sp(
                coords=arr.coords, data=arr.data, fill_value=sp.nan, shape=arr.shape
            )
        else:
            arr = sp(num_to_nan(np.nan_to_num(arr.todense())), fill_value=value)
    else:
        arr[arr == 0] = value
    return arr
```

It seems like we can actually assign a non-nan value according to the last else statement. So I wasn't sure if this function was intended to have any kind of type being assigned to the **value** parameter or if it was supposed to be the nan of the corresponding backend.

If it's the latter, cupy is lacking a nan whereas the other three backends have one. So, I was wondering if it would make sense to try to add one. I could also add some validation to enforce the typing and intended use of the function.

jbogaardt:
I don't think the function is used for anything other than setting nans. But I don't know how well cupy is supported at this point either. We don't run gpu unit/regression tests and I haven't used in a while. I suspect supporting cupy as it's currently implemented is not robust. The repo is heavily tested against numpy and sparse backends.

PR #564, opened 2025-05-20
Dangling commits?

genedan:
I know we did a few merges over the last couple months, but for some reason my fork seems to always be perpetually 20-something commits ahead even though I did see that my last 2 PRs were merged. Can we get these in somehow?


kennethshsu:
Ha this annoys me too, but I think this is because for some reason, you didn't pick up the right commit history.

You should be able to fix this with a reset and a force push:

```bash
git fetch upstream
git checkout main
git reset --hard upstream/main
git push origin main --force
```

As an alternative, since no files have changed, you can just refork again from the main CAS repo. I think this way is actually cleaner. Your PR is currently proposing to incorporate your commit history to the CAS repo, even though nothing (no code files) changed. I don't think this makes sense.

PR #577, opened 2025-06-27
Add support for Weibull curve in tail fitting methods

andrewwilson201:
This PR extends the tail fitting functionality in chainladder by adding support for a Weibull curve as an alternative to the existing exponential and inverse_power options.

Summary of changes:
- Added "weibull" as a valid curve type in self.curve logic
- Implemented the log–log transformation needed for Weibull fitting
- Adjusted the extrapolated tail LDF calculation accordingly
- Updated the error message for invalid curve types

### DIFF
--- a/chainladder/tails/curve.py
+++ b/chainladder/tails/curve.py
@@ -71,9 +71,10 @@ class TailCurve(TailBase):
 
         if curve not in [
             'exponential',
-            'inverse_power'
+            'inverse_power',
+            'weibull'
         ]:
-            raise ValueError("Invalid curve type specified. Accepted values are 'exponential' and 'inverse_power'.")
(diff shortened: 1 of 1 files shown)


jbogaardt (approved):
@andrewwilson201, this is very cool! Thanks for adding the additional functionality. The code looks great. We may want to add something to the docstrings so users can know about the new curve. 
A few of us meet up every couple weeks to discuss the direction of the chainladder-python project. If you're interested in an invite, let me know.

Otherwise, fantastic work!

kennethshsu:
Hi @andrewwilson201, would you like to share me your email address? Or you can message me on [LinkedIn](https://www.linkedin.com/in/kennethshsu/)?

PR #586, opened 2025-07-16
Issues template

kennethshsu:
I revised the bug report and feature request template using GitHub forms. I don't know how I can preview them locally so something might not show up correctly.


genedan:
Okay, I have a question though, how do you usually go about reviewing GitHub Actions/administrative changes? On my solo projects I just commit, hope for the best, and then fix things if they look wrong. I guess that won't work with peer reviews.

I could read through the new files to see if they make sense and then we can fix the template if there are any problems, but I'm open to other ways of reviewing.

kennethshsu:
Ha, I don't think so, I thought I could go to the different branches to preview the changes, but I don't think it works that way. So ya, let's just hope for the best. I am sure there are mistakes in there.

genedan (approved):

Issue #583, opened 2025-07-11
Voting chainladder got error


kennethshsu:
### Discussed in https://github.com/casact/chainladder-python/discussions/530

<div type='discussions-op-text'>

<sup>Originally posted by **Vu1992** June 12, 2024</sup>
I've encounter this problem wehn try to use Voting Chainladder. I think it could be that sklearn update their _fit_single_estimator(). Please take your time to update this. thanks
![image](https://github.com/casact/chainladder-python/assets/166093797/33ff07c5-f7a6-4e58-8cdd-b2a65ea0b20c)
</div>

kennethshsu:
To reproduce:

```python 
raa = cl.load_sample('RAA')

bcl = cl.Chainladder()
cc = cl.CapeCod()

vot = cl.VotingChainladder(estimators=[('bcl', bcl), ('cc', cc)], weights=np.array([[1, 0]] * 4 + [[0, 1]] * 3 + [[0, 1]] * 3))
vot.fit(raa)
vot.ultimate_
```

kennethshsu:
Expected result:
```python
                      2262
        1981  18834.000000
        1982  16875.500226
        1983  24058.534810
        1984  28542.580970
        1985  28236.843134
        1986  19905.317262
        1987  18947.245455
        1988  23106.943030
        1989  20004.502125
        1990  21605.832631
```

kennethshsu:
In array form
```
[18834.00, 16875.50, 24058.53, 28542.58, 28236.84, 19905.32, 18947.25, 23106.94, 20004.50, 21605.83]
```

henrydingliu:
Can't reproduce

<img width="1215" height="527" alt="Image" src="https://github.com/user-attachments/assets/012de9a2-070d-492f-91f2-b5be39cbcbf7" />

PR #593, opened 2025-08-14
583

kennethshsu:
Addressed #583. Added a test for it.


henrydingliu (approved):
param bug fix makes sense. voting test makes sense. rest are formatting changes

Issue #608, opened 2025-10-10
Fix lab_black usage in docs


kennethshsu:
### Discussed in https://github.com/casact/chainladder-python/discussions/607

<div type='discussions-op-text'>

<sup>Originally posted by **henrydingliu** October  9, 2025</sup>
This [page ](https://chainladder-python.readthedocs.io/en/latest/getting_started/tutorials/development-tutorial.html)has an error for me. </div>

SRwcf:
@kennethshsu I have opened a quick PR replacing the deprecated lab_black with jupyter_black.

However, encountered an issue in triangle-tutorial.ipynb with axis="segments". 
Let me know if I should tweak anything or turn this into a draft for now!

genedan:
Is this issue good to close? I think this one was solved by a PR a while back.

PR #632, opened 2025-12-17
Fix for #608

henrydingliu:
fixing jupyter_black dependency. 

Tested on [separate RTD instance](https://hdlclp.readthedocs.io/en/latest/getting_started/tutorials/triangle-tutorial.html) based on my repo.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,6 +66,7 @@ docs = [
     "parso>=0.8",
     "polars",  # For docs examples
     "statsmodels",  # For docs examples
+    "jupyter_black",
 ]
 test = [
     "lxml",


kennethshsu:
@henrydingliu I thought we agreed to not put linter as a dependency?

kennethshsu:
Ya, I actually was thinking maybe we should break out the doc repo vs the code repo.... Some projects do that. Maybe we can discuss more...

henrydingliu:
lol, our doc setup is indeed a bit outdated. in researching the latex issue, i canvassed many repos and saw that nobody is using .ipynb in their doc anymore.

kennethshsu:
Ya I hate the review of ipynb's. We should probably discuss this more.  I'm all about documentation, as it's the first thing people see...

Issue #634, opened 2025-12-22
Issue on page /gallery/plot_ptf_resid.html


henrydingliu:
The bottom right graph doesn't match the paper
<img width="677" height="531" alt="Image" src="https://github.com/user-attachments/assets/45524812-f7d7-4bef-8034-df95548f0b8b" />

This is likely caused by the paper using the exposure vector
<img width="496" height="759" alt="Image" src="https://github.com/user-attachments/assets/4c5257e6-f305-44de-afd9-f19158fa2741" />

This is not explicitly discussed in the main body of the paper, and only mentioned in Appendix C
<img width="681" height="475" alt="Image" src="https://github.com/user-attachments/assets/cdef6c0b-70d2-440d-8e5a-4ddb401f9a30" />

PR #638, opened 2026-01-03
Fix for #634

danielfong-act:
After adjusting the incrementals by their exposure in plot_ptf_resid.py, the resultant graphs match those of Barnett and Zehnwirth's paper (p. 281 of [this paper](https://www.casact.org/sites/default/files/database/proceed_proceed00_00245.pdf)). Exposures are given on the previous page. Also, the origin graph needed its axis limits tweaked. For reasons unknown, the data was stretched to the edges of the plot.

Thanks to @henrydingliu for noting that you don't need to construct a triangle to do triangle arithmetic; a numpy array works fine.


danielfong-act:
I swore I fixed that

PR #642, opened 2026-01-07
[Work in progress] BarnettZehnwirth formula helper (with some refactor)

### LINKED ISSUE #623: BarnettZehnwirth does not support categorical valuation predictors

### Describe the bug in words

Discussed in #579 

The BarnettZehnwirth estimator fails to convert some values of `valuation` to categories if given a formula that contains `C(valuation)` or some similar term. In particular, a `PatsyError` is raised: `Error converting data to categorical: observation with value X does not match any of the expected levels`, where X is the next valuation date after the latest diagonal converted to an ordinal. 

```
import chainladder as cl
abc = cl.load_sample('abc')
cl.BarnettZehnwirth(formula='C(valuation)').fit(abc)
```

BarnettZehnwirth should estimate coefficients for valuations up to the latest diagonal. Future valuation categories could be dropped or have their coefficients set to zero. 
Fitting data with BarnettZehnwirth also populates the resultant triangle with model expectations (and predictions) for each cell in `Triangle.model_.triangle_ml_`. The estimator should use valuation category coefficients for this estimation when it can.

I'm not sure how to do the above in a way that works with Patsy and fitted triangle calculations.

danielfong-act:
This PR is intended to close #623. 

The heavy lifting by @henrydingliu of replicating the model specified [in this paper](https://www.casact.org/sites/default/files/database/forum_08fforum_3barnett_zehnwirth.pdf) has been generalized and placed into the helper function PTF_formula located in chainladder.utils.utility_functions. 

Also, the slapdash multicollinearity warning I added a while ago was removed, because any formula constructed by PTF_formula will trigger it.

Things that aren't done:
1. The documentation and examples haven't been updated yet. They should show how to build patsy formulas to create models akin to those of the papers. 
2. Coefficient names in `coef_` aren't so great. I'd like to replace them with numbered alphas, gammas and iotas to make what's going on more obvious. Pandas.DataFrame.rename seemed like the right tool for that job, but I had no luck with it. 

Things to consider:
1. The docstring for PTF_formula could be made clearer. Feel free to suggest edits.
2. Should the functionality of PTF_formula be inside BarnettZehnwirth itself? I opted to place it outside to avoid dealing with the internal interaction between user specified formulas and Greek parameters.
3. Should how alpha, gamma, and iota are specified to PTF_formula be tweaked? Currently, origin groups are disjoint while development and valuation groups overlap.

### DIFF
--- a/chainladder/development/tests/test_barnzehn.py
+++ b/chainladder/development/tests/test_barnzehn.py
@@ -1,6 +1,7 @@
 import numpy as np
 import chainladder as cl
 import pytest
+from chainladder.utils.utility_functions import PTF_formula
 abc = cl.load_sample('abc')
 
 def test_basic_bz():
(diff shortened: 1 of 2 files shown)


henrydingliu:
some recommendations

- since the ptf formulas are only used in constructing a patsy formula, we can build the ptf formulas directly into the [patsyformula](https://github.com/casact/chainladder-python/blob/749eb51e7b1a1ea6f4aff8428e921b603a586e76/chainladder/utils/utility_functions.py#L646) class. the three lists can be 3 optional parameters, and create terms that are added to the main formula (which could be left as empty so only the ptf terms are used). 
- if we did the above, we'd also need to add those lists as optional parameters to glm and bz
- renaming can be done directly within the design matrix specification. this becomes possible if we incorporate the ptf formulas into the patsyformula class. see their [doc](https://patsy.readthedocs.io/en/latest/expert-model-specification.html) for info. 
- we can simplify the ptf formula specification to lists of numbers `[2,3,8]`, rather than the lists of tuples `[(2,3),(3,8),(8,99)]`. then we dont have to worry about overlaps etc.

kennethshsu:
@henrydingliu do you want to be the reviewer for this one?

danielfong-act:
Done. Also, I changed back how the origin coefficients are fitted. The way PTF_formula fits origin parameters differs from the specification of alphas in the papers. First, we fit an intercept, which seems to me to be a necessary way to get the first alpha, $\alpha_0$, to fit properly. Each later origin coefficient is fit as an incremental value, with the intercept acting as a baseline. Thus, $\alpha_i$ in the paper corresponds to the sum of the intercept and the first $i$ origin coefficients. Stated differently, origin coefficient $i$ is the difference $\alpha_i-\alpha_{i-1}$.

I tried to reformulate things a number of different ways to bring the origin coefficients in line with the paper. I gave up after a few tries, but inadvertently left things so that origin coefficent $i$ was actually equal to $\alpha_i-\alpha_0$. I think that's a lot more confusing, so thanks for making me check.

henrydingliu (approved):

PR #649, opened 2026-01-21
document 3.14 support, add to test matrix

hutch3232:
- in addition to 3.14 support, adding more badges to the README for fun


kennethshsu (approved):

hutch3232:
I think it's because we never made a new release once this change was merged. So it is currently only supported in dev (on GitHub). Need to publish sometime to support it broadly.

PR #657, opened 2026-01-21
Barnett Zehnwirth documentation and formulation rework

danielfong-act:
### Documentation Editing Concerns
I'd like to start here, just in case I have to close this PR to rework everything. This all took some time, in part because of my concerns over how to edit the docs. I found that merely opening a notebook in jupyter lab (invoked by `uv run --with jupyter jupyter lab`) immediately changed many parts of the file. While I can still build the docs locally, I'm worried that this will break things upstream. Editing the notebook using a text editor is cumbersome outside of small edits, and would have made adding cells incredibly annoying. I couldn't find a safe way to edit things otherwise, so I went ahead and used jupyter lab. Please tell me if there's a better way of editing the docs, or if I am worrying about nothing. 
EDIT: I built the docs [here](https://danielfong-act-chainladder-python.readthedocs.io/en/latest/intro.html). This seems like a non-issue.

### Changes
This PR is intended to put any BarnettZehnwirth related issues or concerns in the ground. The documentation related to this estimator in user_guide/development has been changed in the following ways:
1. The PTF formula was updated to match implementation/fix a typo
2. The 3 direction linear (ordinal) model was removed
3. Documentation for chainladder.utility_functions.PTF_formula was added
4. The model from [this paper](https://www.casact.org/sites/default/files/database/forum_08fforum_3barnett_zehnwirth.pdf) (pp. 48-49) was added as an example for how to apply PTF_formula. 
5. Added the above paper to the references.
6. Things were moved around/reworded where I saw appropriate.

In addition, I changed how gamma and iota are passed to PTF_formula, and changed test_barnzehn accordingly.

### To be changed
Before merging I would like to address some things, including implementation considerations raised earlier. 
1. Do we want to integrate PTF_formula somewhere into BarnettZehnwirth/PatsyFormula/etc.? 
2. Do we want to change how the greek parameters are passed to PTF_formula? Currently, parameter groups are passed as tuples denoting boundaries. This convention is consistent and very explicit, but adds some unnecessary redundancy. 
3. Documentation rewording/restructuring. The main goal is clarity.
4. Any other concerns not listed here.
If any discussion runs long or major changes are suggested, we could also create a new branch and iterate from there.

### DIFF
--- a/chainladder/core/pandas.py
+++ b/chainladder/core/pandas.py
@@ -20,7 +20,11 @@ if TYPE_CHECKING:
     from collections.abc import Callable
     from numpy.typing import ArrayLike
     from types import ModuleType
-    from typing import Type
+    from typing import (
+        Literal,
+        Self,
+        Type
+    )
 
 
 
@@ -308,7 +312,11 @@ class TrianglePandas:
 
         return concat((self, other), 0)
 
-    def rename(self, axis, value):
+    def rename(
+            self,
+            axis: Literal['index', 'columns', 'origin', 'development'] | int,
+            value: list | str
+    ) -> Self:
         """Alter axes labels.
 
         Parameters
@@ -328,12 +336,18 @@ class TrianglePandas:
         value = [value] if type(value) is str else value
         if axis == "index" or axis == 0:
             self.index = value
-        if axis == "columns" or axis == 1:
+        elif axis == "columns" or axis == 1:
             self.columns = value
-        if axis == "origin" or axis == 2:
+        elif axis == "origin" or axis == 2:
             self.origin = value
-        if axis == "development" or axis == 3:
+        elif axis == "development" or axis == 3:
             self.development = value
+        else:
+            raise ValueError(
+                "Invalid value provided to the 'axis' parameter. Accepted values are a string of 'index', "
+                 "'columns', 'origin', or 'development', or an integer in the interval [0, 4] specifying the"
+                 " axis to be modified."
+            )
         return self
 
     def astype(self, dtype, inplace=True):
--- a/chainladder/core/tests/test_triangle.py
+++ b/chainladder/core/tests/test_triangle.py
@@ -89,6 +89,30 @@ def test_append(raa):
     raa.append(raa2).sum() == raa * 2
     assert raa.append(raa2).sum() == 2 * raa
 
+def test_rename_columns(genins, clrd) -> None:
+    """
+    Test the renaming of triangle columns.
+    """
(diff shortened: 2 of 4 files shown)


henrydingliu:
>but I'll have to add some logic to handle BZ being passed a formula and Greeks

would it break anything if you just add the greeks to the formula if provided? 

>I'll update the docs shortly when I change those two things above

I think about the theoretical progression as the [ELRF residual](https://chainladder-python.readthedocs.io/en/latest/gallery/plot_elrf_resid.html) being beginner, [PTF residual](https://chainladder-python.readthedocs.io/en/latest/gallery/plot_ptf_resid.html) being intermediate, and the BZ08 PTF residual being advanced. So do we really want something as advanced as BZ08 PTF in the main body in the development user guide? Would it be more suitable as an example?

danielfong-act:
>would it break anything if you just add the greeks to the formula if provided?

Kinda yeah. The formulation in my updated docs matches the output produced by fitting models with PTF_formula, with a minor difference from the authors' formulation. Using Patsy formulas to fit categorical variables directly leads to a fairly different formulation.
1. The intercept is no longer equal to $\alpha_0$, since fitting `C(development)` directly estimates a coefficient for the first development period. Thus, $\alpha_0$ is decomposed into the intercept plus a coefficient for the first development year, $\gamma_0$.
2. Each origin coefficient is the incremental trend added onto the intercept. In terms of the PTF_formula formulation (NOT the authors' formulation), the $i$'th origin coefficient is equal to $\sum_{k=1}^i\alpha_k$.
3. Similarly, each development coefficient represent that year's incremental trend above the baseline. In terms of the authors' formulation, the $j$'th development trend is $\sum_{k=1}^j \gamma_k$, which does not include the first year's development trend $\gamma_0$.
4. The first development year is year 0 in the papers and in PTF_formula. Using Patsy directly causes the first year to be whatever the development grain is. This causes ordinal valuation trends to affect the intercept (origin and valuation years are indexed from 0, and thus do not affect the intercept).

I think mixing the two would lead to coefficients that are harder to interpret. Side note, I should add something to make it clear that the formulation is different for the [pathological example](https://chainladder-python.readthedocs.io/en/latest/gallery/plot_ptf_resid.html). 

>So do we really want something as advanced as BZ08 PTF in the main body in the development user guide? Would it be more suitable as an example?

Maybe, but I don't have another example that uses PTF_formula. I think the functionality provided by PTF_formula is core to the authors' intentions; without it, we're left with categorical/ordinal only models. We'll have to come up with something to take its place.

Issue #655, opened 2026-01-21
Triangle dimension metadata fails to update after TriangleBase.rename


genedan:
### Are you on the latest chainladder version?

- [x] Yes, this bug occurs on the latest version.

### Describe the bug in words

Renaming the columns of the triangle fails to update the attribute `Triangle.columns_label`. I would suspect the same issue would happen to `Triangle.origin_label`.

I would also think the reverse issue would happen, where attributes `Triangle.columns_label` and `Triangle.origin_label` are re-assigned, but with those changes failing to cascade to the origin and columns indexes.

I think the columns_label case is easier to fix than the origin_label case. We can make Triangle.columns_label a property, which can read the column names from the corresponding index.

`origin_label` is trickier though, since it seems like that information is lost when creating the origin axis unless stored elsewhere.

### How can the bug be reproduced?

```python
import chainladder as cl

genins = cl.load_sample('genins')
genins.rename('columns', 'foo')

print(genins.columns.values)
print(genins.columns_label)
```
`print(genins.columns.values)` will output `['bar']` whereas `print(genins.columns_label)` will output `['values']`.

### What is the expected behavior?

`print(genins.columns_label)`  should output `['bar']`.

henrydingliu:
index renaming now working per #676 

@genedan is there still an issue with origin? otherwise we can close

genedan:
@henrydingliu yeah I think it's working, let's close.

PR #659, opened 2026-01-22
FEAT: Add validation to TrianglePandas.rename. Also add tests to ensure cascading of rename to triangle.columns_label.

### LINKED ISSUE #654: Add some validation logic to TrianglePandas.rename

### Describe the bug in words

Some nonsense values can be passed to TrianglePandas.rename without triggering an exception. Instead, the method simply returns `self` if none of the conditional branches are triggered:

```
value = [value] if type(value) is str else value
        if axis == "index" or axis == 0:
            self.index = value
        if axis == "columns" or axis == 1:
            self.columns = value
        if axis == "origin" or axis == 2:
            self.origin = value
        if axis == "development" or axis == 3:
            self.development = value
        return self
```

Users may end up believing that something was renamed, when instead nothing actually happened.

The following is an example where the user mistypes "columns" as "colunms".
```
genins = cl.load_sample('genins')
genins.rename('colunms', 'bar')

```

A ValueError should be triggered, notifying the user that an invalid value was passed to the `axis` argument.

genedan:
Closes #654. Partially addresses #655 regarding the columns_label issue.

### DIFF
--- a/chainladder/core/tests/test_triangle.py
+++ b/chainladder/core/tests/test_triangle.py
@@ -101,6 +101,18 @@ def test_rename_columns(genins, clrd) -> None:
     # Test the cascading of rename to triangle.columns_label.
     assert genins.columns_label == ['foo']
 
+def test_rename_exception(genins, clrd) -> None:
+    # Test incorrect axis argument - misspelling of string.
+    with pytest.raises(ValueError):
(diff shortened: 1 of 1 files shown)


genedan:
k, added a few tests cases (incorrect # columns, misspelled strings, incorrect integer supplied).

PR #729, opened 2026-04-29
Pandas 3.0

### LINKED ISSUE #664: Pandas 3.0 Prep

### Describe the bug in words

Pandas 3.0 breaks the package. 

Amongst the detected bugs so far, the most notable is latest_diagonal, which is caused by triangle.valuation and triangle.valuation_date being out of sync. 

Will create branch and provide link

It's hard to see us moving completely out of pandas, even after adding arrow suppport.

### LINKED ISSUE #724: Drop Python 3.10 support
We'll need to drop Python 3.10 support once this is done, since pandas 3.0 requires at least Python 3.11.

### LINKED ISSUE #725: Adust datetime constants to use Pandas default datetime precision
Pandas 3.0 has changed the default datetime precision to microsecond, when I believe it was nanosecond previously:

https://pandas.pydata.org/docs/dev/whatsnew/v3.0.0.html#datetime-timedelta-resolution-inference

There are some places in the package where we specify nanosecond, which I assume was to conform to the previous pandas default precision:

https://github.com/casact/chainladder-python/blob/6ea35773ab940ae361260b76877ffb7d01e69a6f/chainladder/core/base.py#L535-L543

However this leads to a breaking change whenever we compare these datetimes to those generated from the default pandas conversions, which happens in the `valuation == valuation_date` comparison discovered by Henry, due to a nanosecond-to-microsecond comparison.

I propose that we ~~adopt microsecond as the default datetime, to stay consistent with pandas.~~ adjust constant datetimes such as `ULT_VAL` to use the Pandas default precision, nanosecond for Pandas 2, microsecond for Pandas 3.

### LINKED ISSUE #726: Wrap raw HTML in StringIO
`pd.read_html()` no longer accepts raw HTML input. You'll have to wrap it around StringIO:

https://pandas.pydata.org/docs/dev/whatsnew/v3.0.0.html#other-removals

This leads to the following test failure:

https://github.com/casact/chainladder-python/blob/6ea35773ab940ae361260b76877ffb7d01e69a6f/chainladder/core/tests/test_triangle.py#L14-L18

In this case, I think it's just part of the test setup, so it can be changed there.

### LINKED ISSUE #727: Remove instances of DataFrame.groupby(...,axis=...).
Pandas 3.0 removed the `axis` argument from `DataFrame.groupby()` the maintainers want you to use transpose instead:

https://pandas.pydata.org/docs/dev/whatsnew/v3.0.0.html#other-removals
https://github.com/pandas-dev/pandas/issues/51203

This leads to this test failing:

https://github.com/casact/chainladder-python/blob/6ea35773ab940ae361260b76877ffb7d01e69a6f/chainladder/core/tests/test_triangle.py#L316-L327

~~Unfortunately, I think we'll need to do a bit of rewriting for groupby. The test still seems to be testing a real-life use case, so instead of testing `axis=1`, it'll test the transpose functionality of the Triangle.~~

### LINKED ISSUE #728: Replace object type comparisons with string comparisons when pandas dtype is of type string
Pandas now uses a default `string` type when it used to default to `object` type. 

https://pandas.pydata.org/docs/dev/whatsnew/v3.0.0.html#dedicated-string-data-type-by-default

This will cause existing code that checks for type `object` to fail, such as line 424 in _drop_func:

https://github.com/casact/chainladder-python/blob/6ea35773ab940ae361260b76877ffb7d01e69a6f/chainladder/development/base.py#L414-L436

As far as I can tell, this is the only place in the package where this happens and causes test failures.

genedan:
Updates chainladder to be compatible with pandas 3.0. Closes #664.

Applies the following updates:

- Drops Python 3.10 support, since pandas 3.0 requires Python >= 3.11. Closes #724
- Adopts microsecond as default datetime precision to match pandas. Closes #725
- Wraps raw HTML and JSON in StringIO, since pandas requires it. Closes #726.
- Does **not** deprecate axis argument from `TrianglePandas.groupby()`, instead, changes the test expectation to match pandas 3.0 behavior. Closes #727.
- Replace object type comparisons with string type comparisons, since pandas now defaults to type string. Closes #728

---

> [!NOTE]
> **Medium Risk**
> Updates core triangle datetime handling to use pandas’ default datetime precision (ns vs us) and adjusts serialization/parsing behavior, which can subtly affect valuation boundaries and ultimate-date sentinel comparisons. Also relaxes the pandas version upper bound, so downstream behavior now depends on whichever pandas major version is installed.
> 
> **Overview**
> Updates chainladder to be compatible with pandas 3 by **making datetime precision configurable at runtime**. It introduces `options.DT64_UNIT`/`options.DT64_DTYPE`, computes `options.ULT_VAL` from pandas’ max-supported timestamp, and replaces hardcoded `datetime64[ns]`/1ns adjustments in valuation calculations and aggregations with these options.
> 
> Adjusts pandas-3 API expectations by wrapping raw HTML/JSON strings in `StringIO`, loosening dtype checks (e.g., `is_datetime64_dtype`, `is_string_dtype`), and updating tests for pandas-3 groupby semantics and valuation-date expectations via a new `date_delta_adjustment()` helper. Finally, it removes the `<3.0` pandas dependency cap in `pyproject.toml`/`uv.lock`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2c1fa9f1e03c9a654d2e4abd7eeecb1d4fb23c36. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pyproject.toml
```diff
@@ -14,23 +14,22 @@ maintainers = [
...
-    "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
     "Programming Language :: Python :: 3.13",
     "Programming Language :: Python :: 3.14",
 ]
 keywords = ["actuarial", "reserving", "insurance", "chainladder", "IBNR"]
 dependencies = [
-    "pandas >=2.0, <3.0",
+    "pandas >=3.0",
```
henrydingliu: This makes sense in the Pandas3 branch. But do we need to force everyone to pandas 3?

chainladder/core/base.py
```diff
@@ -540,12 +540,12 @@ def valuation(self):
         ddim_arr = ddims - ddims[0]
         origin = np.minimum(self.odims, np.datetime64(self.valuation_date))
         val_array = origin.astype("datetime64[M]") + np.timedelta64(ddims[0], "M")
-        val_array = val_array.astype("datetime64[ns]") - np.timedelta64(1, "ns")
+        val_array = val_array.astype("datetime64[us]") - np.timedelta64(1, "us")
```
henrydingliu: lol, I never knew us stood for microsecond. Good stuff

kennethshsu:
I'm good with this PR. @henrydingliu, do you think we should hold off?

I generally prefer not to maintain too many versions, especially when it means supporting older dependencies, and I encourage users to adopt the latest versions of dependent packages.

henrydingliu:
jumping from requiring 2.0 straight to requiring 3.0 seems a little steep

agreed on not maintaining too many versions. but if we are not maintaining too many versions, isn't the best practice to allow as old as possible on dependencies?

genedan:
One thing users can do if they want to keep using Pandas 2 is to use an old version of Chainladder. At our next release (or some other future one), we can announce in the release notes that it'll be "the last tested version on pandas 2.3.3. Future versions of Chainladder will support Pandas 3 only". This will give users a searchable way to find the latest version of Chainladder that still works with their pandas 2 setup while giving them a heads up that they need to take steps to upgrade their environment.

henrydingliu:
> One thing users can do if they want to keep using Pandas 2 is to use an old version of Chainladder. At our next release (or some other future one), we can announce in the release notes that it'll be "the last tested version on pandas 2.3.3. Future versions of Chainladder will support Pandas 3 only". This will give users a searchable way to find the latest version of Chainladder that still works with their pandas 2 setup while giving them a heads up that they need to take steps to upgrade their environment.

The mechanism of providing a warning and still letting users use an older version is sound. But I just don't see how it applies to us. Do we actually need anything that's new in pandas 3.0? Aren't we dissatisfied with the axis situation and think Pandas will eventually revert in a future version?

genedan:
Pandas is facing some pretty steep competition from other libraries, like Polars, so my hope is that as time passes, most of the changes on the whole will be better for the user.  That's a situation I think we need to keep an eye on, potentially making a switch if deemed necessary (like we did from pip -> uv).

Performance enhancements and bug fixes will continue to be made in Pandas 3, whereas they won't be in Pandas 2.3.3. The `axis` change was unfortunate, but I think more users will be moving on to the latest version or Polars, so my preference overall is to take steps to keep up with the broader community.

So, here's what I propose as our strategy. When we get word that a new major release is out (let's say from version 2 -> 3), we fix up whatever we can, test it on the last previous version, and announce in the release notes "This is the last stable Chainladder version that will support package v2. For future releases, users are expected to upgrade to version v3." We're a small team, so there's not much capacity for maintaining backwards compatibility. I think this approach gives a good balance between giving users a heads-up, while freeing us to move forward.

Chainladder is open source, while we release new versions, the old versions are still available, can be forked, etc.

As far as Pandas 2 is concerned, I think we should do one more release. Let's bump the `pyproject.toml` to have `pandas >2.3.3, <3`, make sure that it passes the tests, and then inform the users that this is the last version they can use for both Pandas 2 and Python 3.10. Then our next release will require Pandas 3 and Python >= 3.11.

henrydingliu:
>When we get word that a new major release is out (let's say from version 2 -> 3), we fix up whatever we can, test it on the last previous version, and announce in the release notes "This is the last stable Chainladder version that will support package v2.

I don't know if we have the manpower to actually do this. The package currently has 7 dependencies, 3 of which doesn't have any version requirements.

kennethshsu:
@henrydingliu do you want to review this one?

henrydingliu:
> @henrydingliu do you want to review this one?

This PR doesn't reflect [any](https://github.com/casact/chainladder-python/issues/724#issuecomment-4347951724) of the [discussion](https://github.com/casact/chainladder-python/pull/729#issuecomment-4351857103) that we've been having on this topic. It's deprecating 3.10 ahead of schedule. And it's pushing Pandas 3.0 without another release. As presently committed, this PR is not aligned with the goals of the collaborator group. 

The original intent of #664 was to work towards Pandas 3.0 _compatibility_, which resolving 4 of the 5 sub-issues should achieve. Perhaps we should split up this PR for the time being.

genedan:
After making good faith attempts at dual Pandas 2 & 3 compatibility in the commits of this PR, I don't think it's a good idea, and neither would be trying to hang on to soon-to-be-deprecated Python 3.10. Dual compatiblity can be technically achieved by:

1) Adding a nano- or microsecond time adjustment in several locations depending on which installation of Pandas the user has
2) Doubling the number of GitHub actions workflows by testing each combo of pandas 2 and 3 with with Python 3.10 thru 3.14. To ensure compatibility, we would need to run this matrix of workflows for as long as we plan on supporting both versions of Pandas due to the breaking changes. How long would we plan to do this for? If it's just until the deprecation of Python 3.10 in 6 months, it might not be worth it.
3) Any new features we add to Chainladder will have to be checked for both Pandas 2 and Pandas 3 compatibility until we drop Pandas 2 support.

Maintaining backwards compatibility increases the complexity of both the codebase and testing, so I conclude now that if a user wants to stay on Python 3.10 or Pandas 2, they'll need to use an old version of Chainladder.

henrydingliu:
We keep talking past each other. Let's align on some first principles. 

The initial issue was raised for Pandas 3 compatibility. Supporting the newest version of a key dependency is a no-brainer, assuming we are not giving up something. If there are now technical issues with dual compatibility, that doesn't mean the issue automatically changes itself into full migration to Pandas 3. Very few noticed this issue when Pandas 3 came out, meaning most users of the package are still on an older version of Pandas. So requiring them to go to 3 will have a sizable impact. I'm not comfortable with increasing the Pandas requirement on the main branch with little to no notice. At a minimum, we should hold off on this PR till the next meeting where others can voice their opinions and preferences. 

Pushing for newer versions of dependent packages is a good thing in general. It's also something that this package historically has not cared about. As I have previously noted, of the 7 dependencies in this package, we currently don't have any version requirements on 3. Pushing for Pandas 3 when we aren't planning on leveraging any of the Pandas 3 enhancements in this package is a poor Product Management decision. We would be adding to the user's burden without offering a benefit. And we can't even offer a weak "oh we always make an attempt to go to the latest version" because we don't. 

Asking users who don't want to keep up to use an older version of chainladder is also a viable mechanism in general. However, I can't see how that's a good idea at the moment as we are introducing a slew of doc and QoL improvements. By pushing for Pandas 3 in the middle of these improvements, we are implicitly requiring users to upgrade to Pandas 3 or not able to enjoy those improvements. Yes, we'd be tying some benefit to the Pandas 3 migration, but entirely unnecessarily. It would cost us literally nothing to wait till a stable version where most of the proposed doc improvements are completed. Why are we in a rush to push for Pandas 3 now? What is the benefit to the package, the collaborators, or the users?

Now onto the more technical topics

>Adding a nano- or microsecond time adjustment in several locations depending on which installation of Pandas the user has

We can refactor the code to take the default from the Pandas version in use. It might be tedious, but probably a good practice anyway. 

>Doubling the number of GitHub actions workflows by testing each combo of pandas 2 and 3 with with Python 3.10 thru 3.14. To ensure compatibility, we would need to run this matrix of workflows for as long as we plan on supporting both versions of Pandas due to the breaking changes. How long would we plan to do this for? If it's just until the deprecation of Python 3.10 in 6 months, it might not be worth it.

I don't think this is necessary. Most of our users are on Pandas 2. So that's what we test with, till majority of the users are on Pandas 3. As early adopters of Pandas 3 run into issues, they can fix those issues in the package in a compatible way. 

>Any new features we add to Chainladder will have to be checked for both Pandas 2 and Pandas 3 compatibility until we drop Pandas 2 support.

We shouldn't be dropping Pandas 2 for as long as the user base is still using Pandas 2. New features just need to be checked in Pandas 2. Then if early adopters of Pandas 3 have issues with the new features, they can fix.

EKtheSage:
Maybe I’m missing something about the expected user workflow here, but I thought tools like uv generally encourage installing into an isolated project environment from `pyproject.toml` / `uv.lock`, rather than forcing users to mutate their global Python environment.

Would that reduce the practical burden of supporting Pandas 3? In other words, users who want the latest chainladder could install it into a fresh isolated environment without necessarily disrupting their existing projects.

Maybe Henry's concern is this:

A user may have an existing project like:
```
dependencies = [
  "chainladder",
  "pandas<3",
  "some-other-package-that-requires-pandas<3"
]
```

If new chainladder says:
```
dependencies = ["pandas>=3"]
```
then uv cannot satisfy both:
```
pandas < 3
pandas >= 3
```
So the user has to choose:

  - upgrade Pandas / fix compatibility issues
  or
  - stay on older chainladder


And maybe we should seriously think about using `polars` as an alternative backend as Gene alludes to earlier where new python users are slowly moving towards.

henrydingliu:
>In other words, users who want the latest chainladder could install it into a fresh isolated environment without necessarily disrupting their existing projects.

of course they can. But why are we asking them to do that? What's in it for them? What's in it for us? What are some tangible benefits of requiring Pandas 3?

EKtheSage:
I agree that we should be thoughtful about increasing dependency requirements, especially if there is no immediate benefit to users.

That said, I think part of the disconnect here is the assumed workflow. Modern Python workflows are increasingly based on isolated, project-specific environments. In that setup:

- environments are explicit and reproducible
- dependencies are pinned per project
- upgrading a dependency does not disrupt other work

So the practical burden of moving to something like Pandas 3 is lower than it would be in a shared or global environment.

That does not mean we should force the upgrade without reason. I agree that if we are not leveraging Pandas 3 features yet, the benefit is unclear. In fact, I think we should move to `polars` instead 😅 

My main point is that with the expectation of reproducible, isolated environments becoming the norm, user burden is probably lower than we're assuming.

henrydingliu:
>That said, I think part of the disconnect here is the assumed workflow. Modern Python workflows are increasingly based on isolated, project-specific environments. In that setup:

When we did the uv demo a few meetings back, did we get a good sense that everyone in the room was already fully leveraging it or something like it? Not saying that a workflow for pandas 3-based chainladder would be cumbersome to set up as an additional environment for someone who knows how to set up additional environments. But do we know for a fact that most of our users, even most of our collaborators, actually have workflow management ready to go?

kennethshsu:
Is it ok to close this PR for now? Let's focus on other lower hanging fruits for now for our v1.0 milestone?

genedan:
Let me see if I can get some backwards compatibility with the dates in. The longer we keep the date fixes unmerged, the amount of merge conflicts we'll get on the pandas 3.0 branch will accumulate and that'll be more work for us. This branch already has several from the recent wave of PRs that I'll need to sort out.

For example, I think I can fix the ultimate date by doing something like `ULT_VAL = str(pd.Timestamp("2262-01-01") - pd.Timedelta(1, unit=pd.Timestamp.max.unit))`, the second term of which will be version-agnostic.

As a compromise, we can remove the strict upper bound on the pandas version, but keep the lower bound as-is at `pandas >= 2.2.3` this would encourage people to report issues they might find with pandas 3 without discouraging them from installing it in their environments.

genedan:
This one's ready for another review. Summary of changes:

- Reintroduces Python 3.10 support.
- Removed hardcodings of datetime precisions in the codebase.
- Chainladder now extracts default precision based on user's Pandas installation
- ULT_VAL is now dynamically calculated based on the default precision
- Compatible with both Pandas 2 and Pandas 3
- Pandas version specification is now `pandas >= 2.3.3` with no strict lower bound on 3.0
- Added a utility function to adjust dates based on precision. Code examples are suppressed for doctest due to having both Pandas 2 and Pandas 3 outputs, but there is a unit test in the usual spot that tests for both versions to make up for it.

Since it's fully compatible with pandas 2, this should be safe to merge, and I think moving those hardcodings into `__init__.py` and away from the rest of the code would be good for maintenance going forward.

henrydingliu:
thanks for the extensive and monumental work on this. my recommendations:

- per our general guide on collaboration, let's keep the PR size manageable. you already went through the hassle of setting up individual sub-issues, would you mind breaking this up into smaller PRs, one for each sub-issue?
- then in this main pandas 3 PR, we add one run to the test suite that uses pandas 3 on a latest python version we support

genedan:
I can do that. I'll adjust the scope on some of those sub-issues that we changed our minds on too, like Pandas 3.10 support.

Issue #737, opened 2026-05-06
[ENH] Refactor sparse.py type hierarchy


genedan:
### Description

I don't think we'll be able to cleanly solve #736 without reorganizing the type hierarchy of `chainladder/utils/sparse.py`.

What I had in mind was a dual typing protocol of `ArrayModule` (kind of like an ArrayLibraryLike type) and `BackendArray` (like numpy's ArrayLike type, but with modifications for interop between the 4 current backend array types to match our use case). 

However, since the frequently-executed `xp = get_array_module(...)` returns either a ModuleType (numpy, dask, cupy) or an array class (sparse.COO) it's not clear which one it'll be when it's called dynamically. That is, it's unknown whether the type hint should be:

```python
# Case when backend is numpy, sparse, or dask.
xp: ArrayModule = obj.get_array_module(...)
```

or

```python
# Case when backend is sparse.
xp: BackendArray = obj.get_array_module(...)
```

Furthermore, when the backend is sparse, `xp` is also callable, but it's not in the case of the other backends. In order to use xp to create an array, you need to do it at different levels of hierarchy depending on the backend:

```python
# Case sparse
sparse_array = xp(...)
```

```python
# Case numpy
numpy_array = xp.ndarray(...)
```

A more consistent implementation of case sparse would be:

```python
sparse_array = xp.COO(...)
```

If we manage to get a general `ArrayModule` class defined, we could do something like:

```python
array = xp.BackendArray(...)
```

where the type of array to be produced is automatically determined by `Triangle.array_backend`. This would also avoid the various checks that are in place to make sure the backend is a certain type (usually sparse) before creating the array, and avoid calling the backend-specific `ndarray` and `COO`.

### Is your feature request aligned with the scope of the package?

- [x] Yes, absolutely!
- [ ] No, but it's still worth discussing.
- [ ] N/A (this request is not a codebase enhancement).

### Describe the solution you'd like, or your current workaround.

I believe this can be improved by reorganizing the objects in `chainladder/utils/sparse.py`. The library `sparse` can be augmented by assigning it the missing top-level functions that are either in numpy, or the ones that are currently defined in the file, whereas the augmented COO-level method calls can be mapped to `sparse.COO`.

I was able to play around with this approach and get the tests to pass, I would like to double-check later and open a PR for feedback.

### Do you have any additional supporting notes?

[...]

Issue #736, opened 2026-05-06
[ENH] Superclass array modules


genedan:
### Description

There are many places in Chainladder where we assign an array module via:

```python
xp: ModuleType = obj.get_array_module()
```

The return value of `get_array_module` varies depending on the backend. It can be:

- The cupy module (or numpy upon import failure)
- A sparse COO type
- The numpy module
- The dask.array module (or numpy upon import failure)

I have marked the return type of this as ModuleType, which is problematic for the following reasons:

- One of these (sparse COO) isn't ModuleType, it's just `type`
- Even if we fix the return type  of `get_array_module` to `ModuleType | type`, these are too general for tools like IDEs to inspect the methods that we call, such as:

https://github.com/casact/chainladder-python/blob/265e151f723bb7ad3d4269a3dbc0e2f4f600fad7/chainladder/development/development.py#L170

The IDE is unaware of which array module `xp` happens to be and thus cannot pull up any information on documentation, arguments, return types, etc.

`xp` is used as if it belonged to a general array backend class with a common set of methods such `nan_to_num`, with specific implementations depending on whether it happens to be cupy, numpy, sparse, or dask, but since it's not defined, coding tools get lost when trying to inspect it.

### Is your feature request aligned with the scope of the package?

- [x] Yes, absolutely!
- [ ] No, but it's still worth discussing.
- [ ] N/A (this request is not a codebase enhancement).

### Describe the solution you'd like, or your current workaround.

Declaring a class such as `ArrayBackend` with a common set of abstract methods, e.g., `nan_to_num`, `ones`, `nanmean` etc., can solve this, as well as providing a unified return type for `get_array_module`.

### Do you have any additional supporting notes?

This issue may become a moot point if we decide to go the route of creating new unified backend. There's been talk of adopting something like `Arrow`, or `Polars`, but I don't recall if it was going to be *the* single backend or just another option alongside `numpy` and `sparse`. But so long as we are switching back and forth between `numpy` and `sparse`, a general backend type could solve the typing and inspection issues.

genedan:
We might be able to do this without inheriting, by using `typing.Protocol`, and updating `chainladder/core/typing.py`:

```python
from typing import Protocol

class ArrayModule(Protocol):
    """Structural type for array backend modules: numpy, cupy, dask, sparse."""

    ...

    def nan_to_num(self, x: ArrayLike, **kwargs: Any) -> ArrayLike: ...

    ...
```

[...]

PR #739, opened 2026-05-06
REFACTOR: Reorganize type hierarchy of sparse.py. Move array-level fu…

genedan:
## Summary of Changes  

- Reorganize type hierarchy of sparse.py. Move array-level functions to sparse.COO, and keep module-level functions to sp.
- This isn't always possible, so I kept certain overrides like `COO.cumprod`, and added `sp.sum = COO.sum`.
- Replaces xp() with xp.COO() to create sparse arrays when xp used to be callable and was used as a constructor. This keeps the hierarchy consistent across the 4 backends.
- Removes unused args and kwargs from `sparse.py`

## Related GitHub Issue(s)  

#737
#736

Paves the way towards creating a consistent abstract, maintainable backend which may be used to swap specific backends as fashions change. Inches closer towards accomplishing our ambitious backend dreams.

I'm anticipating some hesitation on calling `xp.COO` where `xp` means the result of `Triangle.get_array_module`. However, this replaces instances when `xp` was called directly, which could only be done if the backend was sparse anyway. Therefore, the risk level is the same with this replacement, and there are already existing mechanisms in place that mention that the backend is sparse prior to the call. This reorganization of `sparse.py` is needed before we can tackle replacing `xp.COO` with something more generalized like `xp.BackendArray`.

---

> [!NOTE]
> **Medium Risk**
> Touches core sparse-backend construction paths (`Triangle` init, slicing/assignment, arithmetic, and regression) and changes how `sparse.COO` is instantiated/default fill values are handled, which could affect numerical results or edge cases. Scope is contained to the sparse backend and is backed by new unit tests for key wrappers.
> 
> **Overview**
> Standardizes sparse-backend array construction to always use `sp.COO`/`xp.COO` (instead of treating `sp`/`xp` as a callable constructor), and updates several call sites (`Triangle` initialization, `.loc/.at` sparse assignment, arithmetic validation, and `WeightedRegression`) accordingly.
> 
> Refactors `utils/sparse.py` to clarify the type hierarchy by exposing `COO` explicitly, moving/aligning defaults like `COO.nan`, and trimming unused wrapper signatures; also adds a focused `test_sparse.py` suite covering `array` fill-value behavior plus `where` and in-place `floor` semantics.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 07157937e7c600abb2881f7cb771c72f6835663a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/chainladder/core/dunders.py
+++ b/chainladder/core/dunders.py
@@ -45,7 +45,7 @@ class TriangleDunders:
             if isinstance(other, np.ndarray) and self.array_backend != 'numpy':
                 obj = self.copy()
                 other = obj.get_array_module().array(other)
-            elif isinstance(other, sp) and self.array_backend != 'sparse':
+            elif isinstance(other, sp.COO) and self.array_backend != 'sparse':
                 obj = self.set_backend('sparse')
             else:
                 obj = self.copy()
--- a/chainladder/core/slice.py
+++ b/chainladder/core/slice.py
@@ -103,7 +103,7 @@ class _LocBase:
                 (self.obj.values.coords, np.array(key)[:, None]), 1)
             self.obj.values.data = np.concatenate(
                 (self.obj.values.data, np.array([values])), 0)
-            self.obj.values = self.obj.get_array_module()(
+            self.obj.values = self.obj.get_array_module().COO(
                 self.obj.values.coords, self.obj.values.data, prune=True,
                 has_duplicates=False, shape=self.obj.shape,
                 fill_value=self.obj.values.fill_value)
@@ -238,7 +238,7 @@ class TriangleSlicer:
                 value.values.coords[1] = i
                 coords = np.concatenate((before.coords, value.values.coords), axis=1)
                 data = np.concatenate((before.data, value.values.data))
-                self.values = xp(
+                self.values = xp.COO(
                     coords, data, shape=self.shape, prune=True, fill_value=xp.nan
                 )
             else:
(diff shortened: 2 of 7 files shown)


genedan:
@jbogaardt, we decided that we need your review on this one, for your Sparse knowledge.

I wrote a bunch of text, but the quick 1-sentence summary is that I'm trying to get `get_array_module` to return a single return type (ModuleType) rather than two possible return types (ModuleType | COO) that would be hard to predict when `get_array_module` is called dynamically.

jbogaardt (approved):
chainladder uses the xp = numpy if dense else “sparse shim” pattern. This requires xp to expose a numpy-like surface (xp.where, xp.nansum, xp.array, xp.nan, etc.). At the time I wrote this, the sparse package's module-level API was thin — it didn't have direct equivalents for a lot of these, so I had to patch the gaps in somewhere.
 
Why did I patch them onto COO (the class) rather than onto the sparse module. I'd guess my chain of thought was:

  1. "I want xp(coords=..., data=...) to construct an array in one call, the way np.array(...) does for dense." → alias COO as sp.
  2. "Now I need somewhere to hang nansum, where, etc. so xp.nansum(...) works." → the thing already named sp is right there, so attach to it.

It worked, so I moved on and haven’t thought about it since. The sparse package has matured since then and has proper module-level functions for most of what we are patching. So your choice to flip to the more conventional layout (module is module, class is class) is a step in the right direction.

PR #811, opened 2026-05-18
#272 added geometric averaging type to the Development() estimator

### LINKED ISSUE #272: Medial and geometric averages
Hey guys,

Do you think it would be possible to add medial and geometric average options to the Development class? I've been working on a FASLR feature where the user can select which types of averages they want displayed on their screen. So far I've included the chainladder supported options (simple, volume, regression) and would like to add these two:

image

image

kennethshsu:
## Summary of Changes  
Added geometric LDFs in the `Development()` estimator.

## Related GitHub Issue(s)  
- Fixes #272 
- Removed blocker for #588 

Because geometric averaging cannot be easily implemented in the regression framework, it is calculated outside.  We then have to carefully `nan` out all relevant params related to the estimated values, such as `sigma_`, `std_err_`, and `std_residuals_`.

- [ ] I passed tests locally for both code (`uv run pytest`) and documentation changes (`uv run jb build docs --builder=custom --custom-builder=doctest`)

---

> [!NOTE]
> **Medium Risk**
> Introduces a new averaging mode and rewires weighted regression/std-error calculations, which can change LDFs and stochastic outputs (`sigma_`, `std_err_`, residuals) across workflows if behavior differs from prior methods.
> 
> **Overview**
> Adds **geometric averaging** as a supported `average` option for `Development`, including per-development-period mixing of `simple`/`geometric` methods.
> 
> Refactors `WeightedRegression` to accept the chosen averaging method, compute regression weights internally (tracking `_w_reg`/`exponent_`), handle geometric averaging via a log-based coefficient calculation with a new warning for zero values, and centralizes `std_err_` filling via a new `std_err_fill()` helper. Updates `Development.fit()` to use these new regression outputs when computing standardized residuals.
> 
> Expands the test suite to validate geometric/simple averaging results and to assert expected `sigma_`, `std_err_`, and `std_residuals_` values.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 81cef7e8abd11a2c8a084c2b8cda40bd2d2e6af3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chainladder/development/development.py
```diff
@@ -211,13 +213,53 @@ def fit(
...
+
+            # zero out the sigma, std_err, and std_residuals_ for the whole array whenever
+            # geometric average is used (becuase everything is wrong and useless)
+            nan_params = params.copy()
+            nan_params[..., 1, :] = xp.nan
+            nan_params[..., 2, :] = xp.nan
+
+            self.sigma_ = self._param_property(obj, nan_params, 1)
+            self.std_err_ = self._param_property(obj, nan_params, 2)
+            self.std_residuals_ = self.std_residuals_ * np.nan
```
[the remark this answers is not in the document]
kennethshsu: This is wrong? All `sigma_`, `std_err_`, and `std_residuals_` need to be errored out if even one of the LDFs is `geometric`.

henrydingliu:
geomean is the best estimator of a log-linked linear regression where the slope is hardcoded to -1. 
see this deconstruction in Excel [geomean.xlsx](https://github.com/user-attachments/files/27978477/geomean.xlsx)
this can folded into either the `WeightedRegression` class or the `DevelopmentML` estimator. `WeightedRegression` is superior because it deals with multiple triangles

for the `WeightedRegression` class, we can add a `link` string parameter and a `neg-diag` bool (similar to the `thru-orig` bool). then we just need to write a new `_fit_ols_neg_diag` function and slightly modify `_fit_ols` and `_fit_ols_thru_orig` to apply the `link` function to x and y.

kennethshsu:
Wow how come you didn't share this on the thread lol, I was trying to find texts that would do this in the regression framework. Let me rework this then, does the other variable (`sigma_`, `std_err_`, and `std_residuals_` )still make sense, or do they need to be errored out?

henrydingliu:
sorry, not negative diagonal slope. flat slope. 

sigma_ and std_err_ are both derived from residual. the formula for residual will be a bit different. but i think the rest is still applicable

kennethshsu:
Ya I figured it out. Ok will implement it this way instead.

Where did you find the formula from? Just curious?

kennethshsu:
Do you have the formula for the residuals? If they are different?

henrydingliu:
there's two ways of calculating residual in the package. i'm not sure which one is more consistent with the theoretical underpinning behind geomean. standard development has a set of [calculations](https://github.com/casact/chainladder-python/blob/72b270c3499992ce5fd0c0b06889680f7cfe467a/chainladder/utils/weighted_regression.py#L84). barnzehn has its own [calculations ](https://github.com/casact/chainladder-python/blob/72b270c3499992ce5fd0c0b06889680f7cfe467a/chainladder/development/barnzehn.py#L72)on the log scale. i say we just implement whatever is the easiest. then if someone who actually uses both geomean in a stochastic workflow finds an issue, they can propose a change.

chainladder/utils/weighted_regression.py
```diff
@@ -26,16 +27,81 @@ def infer_x_w(self):
...
+
+        std_err = xp.sqrt(mse / denominator)
+        std_err = std_err[..., None]
+
+        coef = coef[..., None]
+        sigma = xp.sqrt(mse)[..., None]
+
+        self.slope_ = coef
+        self.sigma_ = sigma
+        self.std_err_ = std_err
```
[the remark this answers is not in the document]
kennethshsu: Won't fix this, we will let the errors terms flow through using the original regression framework calculations.

kennethshsu:
@henrydingliu ok ready for review!

kennethshsu:
In case anyone wants to look at the Excel replication. 
[geomean.xlsx](https://github.com/user-attachments/files/28084177/geomean.xlsx)

chainladder/development/tests/test_development.py
```diff
@@ -393,6 +391,264 @@ def test_new_drop_10():
...
+                "geometric",
+            ],
+        )
+        .fit_transform(tri)
+        .ldf_.to_frame()
+        .values.flatten(),
+        6,
+    )
+
+    def sim_lastn(s, n):
```
henrydingliu, line 463: should these be centrally defined?

chainladder/development/development.py
```diff
@@ -92,8 +92,8 @@ class Development(DevelopmentBase):
     def __init__(
         self,
         n_periods: int = -1,
-        average: Literal['volume', 'simple', 'regression'] = "volume",
-        sigma_interpolation: Literal['log-linear', 'mack'] = "log-linear",
+        average: Literal["volume", "simple", "regression", "geometric"] = "volume",
```
henrydingliu, line 95: small nit. technically speaking, average could be list?

chainladder/utils/weighted_regression.py
```diff
@@ -26,16 +27,86 @@ def infer_x_w(self):
...
+            self.exponent_ = exponent
+            w = num_to_nan(w / (x**exponent))
+            denominator = num_to_nan(xp.nansum((y * 0 + 1) * w * x * x, axis))
+            reg_coef = num_to_nan(xp.nansum(w * x * y, axis)) / denominator
+
+            # special case for geometric average, still using the framework,
+            # but using the log link function and taking the differences
+            w_geo = num_to_nan(self.w)
+            geo_coef = xp.exp(
+                xp.nanmean(w_geo * xp.log(y), axis)
```
henrydingliu: do we need any extra error catching around negative values ?

chainladder/utils/weighted_regression.py
```diff
@@ -26,16 +27,86 @@ def infer_x_w(self):
...
+        xp = self.xp
+
+        if average_ is None:
+            self.exponent_ = xp.nan_to_num(y * 0)
+            denominator = num_to_nan(xp.nansum((y * 0 + 1) * w * x * x, axis))
+            coef = num_to_nan(xp.nansum(w * x * y, axis)) / denominator
+
+        else:
+            # calculate the coef using regression framework
+            exponent_map = {"regression": 0, "volume": 1, "simple": 2, "geometric": 1}
```
henrydingliu, line 63: confirming my understanding. `"geomtric":1` doesn't actually do anything

chainladder/development/tests/test_development.py
kennethshsu: Nah, I don't see how that can be reused outside calculating the LDFs... 

Unless you are saying you want a function of to calculate a metric (mean, sum, min, max) using only the last n values?

chainladder/development/development.py
kennethshsu: Like to do something like `average=["simple", "simple", "geometric", "volume"]`? Yes, you can.

kennethshsu: Just looked at the docstring and it doesn't look like it's super clear there, I can improve that after merge.

chainladder/utils/weighted_regression.py
kennethshsu: I don't think a cumulative triangle can ever be negative. But I do a potential problem in an edge case where losses are 0 for the first few periods for a super long tailed business. In that case, the geometric LDFs will error out...

chainladder/utils/weighted_regression.py
kennethshsu: Ya the coefficients calculated using the traditional regression framework with the wrong link function for geometric are simply thrown out. 

There's no simple way to consolidate the calculations I don't think, let me know if you have ideas on how to generalize this.

chainladder/development/tests/test_development.py
henrydingliu: i meant inside the test_development.py. it's currently repeated multiple times inside various tests

chainladder/utils/weighted_regression.py
henrydingliu: i like this. it cleaner than having to do a where on x, a where on y, and then a 3rd where on slope

chainladder/utils/weighted_regression.py
henrydingliu: geometric ldf is calculated regardless of whether it's selected, correct? so this could error out even when the user doesn't want geomean at all

chainladder/development/tests/test_development.py
kennethshsu: Ahh I see, hmm I think it's probably ok to leave inside because when a test fail you can easily see what failed. This is probably not that big of a deal. It's such a small helper function.

chainladder/utils/weighted_regression.py
kennethshsu: Yes, that's a good point, though not likely, but ya I can add a guard to skip if geometric is not present so it runs faster.

chainladder/utils/weighted_regression.py
```diff
@@ -26,16 +27,98 @@ def infer_x_w(self):
...
+            w = num_to_nan(w / (x**exponent))
+            denominator = num_to_nan(xp.nansum((y * 0 + 1) * w * x * x, axis))
+            reg_coef = num_to_nan(xp.nansum(w * x * y, axis)) / denominator
+
+            # special case for geometric average, still using the framework,
+            # but using the log link function and taking the differences
+            is_geo = xp.array([a == "geometric" for a in average_[0, 0, 0]])
+            if is_geo.any():
+
+                if xp.any((y == 0) & (x == 0)):
```
henrydingliu, line 77: should we be checking for 0 or for NaN?

kennethshsu: No, it should be 0. For 2 reasons:
1) `np.log(np.nan)` will just return `np.nan`. So this is not a problem.
2) the `NaN`s are actually useful, they are used in `n_periods`, so we actually want the `NaN` to flow through.

henrydingliu (approved):
harmonic mean next?

kennethshsu:
lol if you have the formulas we can do it, it's been refactored so should be pretty easy

PR #812, opened 2026-05-19
docs(triangle): document array_backend and pattern parameters (#521)

### LINKED ISSUE #521: [ENH] Need to add array_backend and pattern to Triangle docstring
The docstring for the triangle class is missing a few parameters in the __init__, specifically array_backend and pattern.

SaguaroDev:
Closes #521.

`array_backend` and `pattern` are real parameters of `Triangle.__init__` (chainladder/core/triangle.py:350-351) but the class docstring's Parameters section didn't list them. Added descriptions consistent with the existing `is_pattern` attribute documentation and the `options.ARRAY_BACKEND` fallback logic.

---

> [!NOTE]
> **Low Risk**
> Docstring-only update that adds missing `Triangle` constructor parameter documentation; no runtime behavior changes.
> 
> **Overview**
> Adds missing documentation for `Triangle`'s `array_backend` constructor parameter, including supported backends and the fallback to `cl.options.ARRAY_BACKEND`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d3124aacc21950cde2fcc05cd2930e74198a880e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/chainladder/core/triangle.py
+++ b/chainladder/core/triangle.py
@@ -71,15 +71,10 @@ class Triangle(TriangleBase):
         (i.e., calendar fiscal periods). When True, the period end is inferred
         from the data itself. This is useful when origin dates do not align
         with calendar period boundaries.
-    array_backend: str
+    array_backend: str, optional (default = None)
         Backend used to store the underlying values array. One of
         ``'numpy'``, ``'sparse'``, or ``'cupy'`` (if installed). If
-        ``None`` (default), falls back to ``cl.options.ARRAY_BACKEND``.
-    pattern: bool
-        Indicates whether the Triangle holds development patterns
-        (e.g. link ratios) rather than aggregated loss values. Defaults
-        to ``False``. Mirrors the ``is_pattern`` attribute and affects
-        array-backend selection and certain downstream methods.
+        ``None``, falls back to ``cl.options.ARRAY_BACKEND``.
 
     Attributes
     ----------
--- a/chainladder/core/triangle.py
+++ b/chainladder/core/triangle.py
@@ -71,6 +71,15 @@ class Triangle(TriangleBase):
         (i.e., calendar fiscal periods). When True, the period end is inferred
         from the data itself. This is useful when origin dates do not align
         with calendar period boundaries.
+    array_backend: str
+        Backend used to store the underlying values array. One of
+        ``'numpy'``, ``'sparse'``, or ``'cupy'`` (if installed). If
+        ``None`` (default), falls back to ``cl.options.ARRAY_BACKEND``.
+    pattern: bool
+        Indicates whether the Triangle holds development patterns
+        (e.g. link ratios) rather than aggregated loss values. Defaults
+        to ``False``. Mirrors the ``is_pattern`` attribute and affects
+        array-backend selection and certain downstream methods.
 
     Attributes
     ----------


henrydingliu:
thanks for catching. 

i'm supportive of adding array_backend to the documentation, because it's something we intend for a user to manually set occasionally. 

i am conflicted over adding pattern. i'm not sure we intend for a user to ever construct a pattern triangle. i couldn't find any example in the documentation. 

what i propose is an experiment, if you don't mind. manually construct a pattern triangle, then try to create a development estimator and a chainladder estimator with it. if both succeeds, then we know the package has minimal functionality around a manually constructed pattern triangle, and can list pattern in the public documentation. if either fails, we probably should treat pattern as an internal variable.

SaguaroDev:
I ran the experiment you suggested. 

**Setup:** built a manual pattern Triangle from a long-form DataFrame, matched to the shape of the canonical `Development().fit(raa).ldf_` (shape `(1,1,1,9)`, `is_pattern=True`, `is_cumulative=False`, identical LDF values).

```python
import chainladder as cl, pandas as pd
raa = cl.load_sample("raa")
ref_ldf = cl.Development().fit(raa).ldf_
ldfs = ref_ldf.values.flatten().tolist()  # 9 values

rows = []
for i, age in enumerate([12, 24, 36, 48, 60, 72, 84, 96, 108]):
    org = pd.Timestamp("1990-12-31")
    val = org + pd.DateOffset(months=age) + pd.offsets.MonthEnd(0)
    rows.append({"origin": org, "valuation": val, "ldf": ldfs[i]})

pat = cl.Triangle(
    data=pd.DataFrame(rows),
    origin="origin", development="valuation", columns="ldf",
    cumulative=False, pattern=True,
)
# pat.shape == (1, 1, 1, 9), pat.is_pattern == True, values match ref_ldf exactly
```

**Results:**

1. `cl.Development().fit(pat)` — returns a Development object without raising, but the resulting `ldf_` values are `[0.333, 0.616, 0.787, 0.853, 0.898, 0.960, 0.968, 0.983]`. It's treating the link ratios as losses and computing ratios-of-ratios. Silent semantic failure.
2. `cl.Chainladder().fit(pat)` — hard fail: `ValueError: could not broadcast input array from shape (1,1,1,8) into shape (1,1,1,9)` (raised in `_get_ultimate` → `_align_cdf`).
3. Control: `cl.DevelopmentConstant(patterns={12: 2.999, ...}, style="ldf").fit_transform(raa)` piped into `cl.Chainladder().fit(...)` works correctly — ultimate matches the reference fit. So the supported "user-supplied LDFs" path is healthy; it just isn't the `Triangle(pattern=True)` path.

By the rule ("if either fails, treat `pattern` as internal"), it sounds like the right call here is to drop `pattern` from this PR and ship `array_backend` only. I can push that revision whenever you want, or if you'd rather keep `pattern` documented with a caveat pointing users to `DevelopmentConstant`, I'm happy to do that instead.

henrydingliu:
thanks for indulging me. i think we provided sufficient evidence here that manually declaring a pattern wouldn't actually allow the user to perform any additional estimation, and therefore should not be a behavior that we encourage through the docstring. and as you pointed out, developmentconstant is already available for hardcoding patterns. please keep pattern as internal. 

if the python police asks why we are not changing the name to _pattern, i plead the fifth.

kennethshsu:
@SaguaroDev can you remove the `pattern` from the docstrings so those are hidden from the user?

SaguaroDev:
Hey Kenneth, sorry for the delay — on it now. Pushing the docstring cleanup shortly.

SaguaroDev:
Pushed d3124aa: dropped `pattern` from the docstring, kept `array_backend` only. Also normalized `array_backend` to the `optional (default = None)` form used by the Development class.

kennethshsu (approved):

chainladder/core/triangle.py
```diff
@@ -71,6 +71,10 @@ class Triangle(TriangleBase):
         (i.e., calendar fiscal periods). When True, the period end is inferred
         from the data itself. This is useful when origin dates do not align
         with calendar period boundaries.
+    array_backend: str, optional (default = None)
+        Backend used to store the underlying values array. One of
+        ``'numpy'``, ``'sparse'``, or ``'cupy'`` (if installed). If
+        ``None``, falls back to ``cl.options.ARRAY_BACKEND``.
```
henrydingliu, line 77: upon another look, i believe the actual behavior is to check for package constant auto_sparse and then tries to convert to sparse, ignoring the constant array_backend.

kennethshsu: Uh oh, did I jump the gun here?

I don't think the package unconditionally try to convert to sparse.

`auto_sparse` (default is `True`) does not automatically convert everything to `sparse`, only when necessary. Trying to recall my memory here: but I think it's something like

numpy -> sparse only if size is large and low density
sparse ->  numpy if above is wrong

To be honest, I don't except a regular user to change this setting at all, which is probably why it was never written in the first place. 

On second thought, I actually think we should take this off. Sorry for jumping the gun here.

kennethshsu:
@henrydingliu see reply above, maybe a new ticket to revert the changes if we want to?

henrydingliu:
no need to revert. i logged a new issue to add documentation for the package options

Issue #575, opened 2025-06-22
[ENH] Make empty Triangle behavior consistent with Pandas


genedan:
So when defining an empty triangle:

```
foo = cl.Triangle()
```

The way it works is that if there is no value provided to the `data` argument, `__init__()` terminates near the beginning and skips over the rest of the initialization code. This leaves the attributes such as `Triangle.values` undefined.

Accessing such attributes leads to a traceback:

```
foo.values
Traceback (most recent call last):
  File "/home/ubuntu/anaconda3/envs/cl_dev/lib/python3.12/site-packages/IPython/core/interactiveshell.py", line 3549, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-48-beca352ff75c>", line 1, in <module>
    foo.values
AttributeError: 'Triangle' object has no attribute 'values'
```

On the other hand, initializing a pandas DataFrame has some placeholder values for the analogous attributes:

```
bar = pd.DataFrame()
bar.values
Out[40]: array([], shape=(0, 0), dtype=float64)
```

If users are like me (who haven't thoroughly gone through all the documentation), they sometimes try pandas methods on a triangle just to see if they'd work like how they would on a DataFrame, and in lots of cases it works.

I think, making such behavior consistent with pandas as much as we can would be a good enhancement, such as creating a `.empty` attribute for an empty Triangle, which can be used in conditional statements:

```
if foo.empty:
    print("I'm empty!")
    
Traceback (most recent call last):
  File "/home/ubuntu/anaconda3/envs/cl_dev/lib/python3.12/site-packages/IPython/core/interactiveshell.py", line 3549, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-49-da5cfd440d07>", line 1, in <module>
    if foo.empty:
       ^^^^^^^^^
AttributeError: 'Triangle' object has no attribute 'empty'
```

vs.

```
if bar.empty:
    print("I'm empty!")
I'm empty!
```

I realize this is a big request, since there are a lot more other features like appending records and assigning columns that you can perform on an empty DataFrame. But for starters I'd say we just handle the initialization attributes for this ticket and then create other tickets as we encounter those other features in the field.

I know even doing this will break some code such as `Triangle.__repr__()`  which defines an empty triangle as one that does not have a `.value` attribute defined. Defining it as `None` or an empty ndarray will break this method.

jbogaardt:
[...]

PR #813, opened 2026-05-19
docs(triangle): clarify Triangle constructor docstring (#490, #538)

### LINKED ISSUE #490: More descriptive docstring for Triangle constructor
## Problem

I have a hard time understanding the exact supported usage patterns for the Triangle constructor based on the source code + docstring. 

## Proposed solution

Make it clear in the source code what valid ways to construct a Triangle are. Pandas includes examples in their docstring: https://github.com/pandas-dev/pandas/blob/v2.1.4/pandas/core/frame.py#L541

## Things that confuse me

**what is optional or not?** The default value is None for all fields, but the type indicated in the docstring is not optional for the data field.

```
    data: DataFrame
        A single dataframe that contains columns represeting all other
        arguments to the Triangle constructor
```

**How do origin and development behave when passed in as lists?** 

```
    origin: str or list
         A representation of the accident, reporting or more generally the
         origin period of the triangle that will map to the Origin dimension
    development: str or list
        A representation of the development/valuation periods of the triangle
        that will map to the Development dimension
```

**development is optional or not?**: The docstring says `str or list` but there are [tests that do not pass it in](https://github.com/casact/chainladder-python/blob/master/chainladder/core/tests/test_triangle.py#L239), is it optional?

### LINKED ISSUE #538: [BUG] origin and development parameters of Triangle class cannot handle lists
**Describe the bug**
The Triangle class docstring states that origin and development parameters can take str or list inputs, but it does not seem to work with lists

**To Reproduce**
Steps to reproduce the behavior. Code should be self-contained and runnable against publicly available data. For example:
```python
import pandas as pd
import numpy as np
np.NINF = -np.inf
import chainladder as cl

raa_df = pd.read_csv(
    "https://raw.githubusercontent.com/casact/chainladder-python/master/chainladder/utils/data/raa.csv"
)

raa = cl.Triangle(
    raa_df,
    origin=["development", "origin"],
    development="development",
    columns="values",
    cumulative=True,
)
raa
```

**Expected behavior**
triangle object with a triangle created for both development year and origin year as development period (In theory would just be a column when development period is used for the origin parameter)

**Desktop (please complete the following information):**
 - Numpy Version [e.g. 2.0.1]
 - Pandas Version [e.g. 2.2.2]
 - Chainladder Version [e.g. 0.8.23]

SaguaroDev:
Closes #490. Closes #538.

#490 asked for a clearer docstring around what's optional, what list types mean, etc. #538 separately noted that origin/development do not actually accept lists in practice and proposed updating the docstring rather than adding list support (per @henrydingliu's [comment](https://github.com/casact/chainladder-python/issues/538#issuecomment-3166285630)).

This PR addresses both by rewriting the Parameters section:

- Mark each parameter as optional/required and state the default where applicable, in line with numpy-style conventions used elsewhere in the package.
- Clarify that `origin`/`development`/`columns`/`index` name a column (or columns) in the supplied DataFrame, instead of the prior abstract 'representation of …' wording.
- Per #538, drop `or list` from `origin` and `development` — lists remain meaningful for `index` and `columns` only.
- Note that omitting `data` returns an empty Triangle and ignores all other arguments, which is what `__init__` actually does today.

The `array_backend` and `pattern` parameters are documented separately in PR #812 (#521); this PR is rebase-friendly with that one.

---

> [!NOTE]
> **Low Risk**
> Low risk: documentation-only changes that clarify parameter expectations and defaults without altering runtime behavior.
> 
> **Overview**
> Clarifies the `Triangle` constructor docstring to better match actual behavior and numpy-style parameter conventions.
> 
> Updates the `Parameters` section to describe `origin`/`development` as *column names* (not lists), document defaults/optional behavior (including that missing `data` yields an empty Triangle and missing `development` implies a single development period), and add clearer wording for `columns`, `index`, formats, and `trailing`’s default.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e45123f18cf6ab241ecf77ee25042dc56f2d3553. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/chainladder/core/triangle.py
+++ b/chainladder/core/triangle.py
@@ -37,34 +37,38 @@ class Triangle(TriangleBase):
     data: DataFrame or DataFrameXchg, or dict
         A single dataframe that contains columns representing all other
         arguments to the Triangle constructor. One may supply a DataFrame-like
-        object (referred to as DataFrameXchg) supporting the __dataframe__ protocol,
-        which will then be converted to a pandas DataFrame. If supplying a dict,
-        it must be structured such that a pandas DataFrame created from it will be
-        accepted by the constructor.
-    origin: str or list
-         A representation of the accident, reporting or more generally the
-         origin period of the triangle that will map to the Origin dimension
-    development: str or list
-        A representation of the development/valuation periods of the triangle
-        that will map to the Development dimension
+        object (referred to as DataFrameXchg) supporting the __dataframe__
+        protocol, which will then be converted to a pandas DataFrame. If
+        supplying a dict, it must be structured such that a pandas DataFrame
+        created from it will be accepted by the constructor. If omitted (or
+        if all arguments are omitted), an empty Triangle is returned.
+    origin: str
+         Name of the column in ``data`` representing the accident, reporting,
+         or more generally the origin period. Maps to the Origin dimension.
+    development: str
+        Name of the column in ``data`` representing the development or
+        valuation period. Maps to the Development dimension. If omitted, the
+        Triangle is treated as having a single development period (e.g. a
+        latest-diagonal-only view).
     columns: str or list
-        A representation of the numeric data of the triangle that will map to
-        the columns dimension.  If None, then a single 'Total' key will be
-        generated.
-    index: str or list or None
-        A representation of the index of the triangle that will map to the
-        index dimension.  If None, then a single 'Total' key will be generated.
-    origin_format: optional str
-        A string representation of the date format of the origin arg. If
-        omitted then date format will be inferred by pandas.
-    development_format: optional str
-        A string representation of the date format of the development arg. If
-        omitted then date format will be inferred by pandas.
+        Name(s) of the column(s) in ``data`` holding the numeric values that
+        will map to the columns dimension. If omitted, a single ``'Total'``
+        key is generated.
+    index: str or list
+        Name(s) of the column(s) in ``data`` that will map to the index
+        dimension. If omitted, a single ``'Total'`` key is generated.
+    origin_format: str
+        A string representation of the date format of the origin column
+        (e.g. ``'%Y-%m-%d'``). If omitted, the date format is inferred by
+        pandas.
+    development_format: str
+        A string representation of the date format of the development column.
+        If omitted, the date format is inferred by pandas.
     cumulative: bool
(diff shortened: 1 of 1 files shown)


henrydingliu:
small nitpick. every parameter is optional after gene's work to allow blank declaration. the [pandas](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.html) documentation isn't bothering with the 'optional' tag.

henrydingliu:
actually, doesn't look like much has been done on #575. @genedan can you confirm that declaring an empty triangle today basically gives you a useless triangle that you cannot edit further? if so, we should note in the docstring

genedan:
I think it's de-facto useless. There are some methods like `Triangle.append()` which I guess could in theory append a Triangle to an empty Triangle with some creative programming. 

Comparing to Pandas, I think an empty DataFrame could be used for static methods, but I don't know if the `Triangle` class is mature enough for those kinds of use cases. @kennethshsu if you can't think of anything off the top of your head then I suggest we just say that filling out none of the arguments just leads to an empty triangle.

I don't think I'd explicitly say it's useless since I'd be open to allowing updates to an empty triangle if someone found a legit use case for it.  One I can think of is:

```python
tri = cl.Triangle()

for i in range(n):
    temp_tri = make_triangle()
    tri.append(temp_tri)
```

But let me know if that's noob programming. I think it is because you could declare triangles the regular way into a list and then use `cl.concat`.

kennethshsu:
I don't know that we should encourage the usage of `.append()` like that, isn't it better to do all that data wrangling in `pandas` and the declare one `Triangle()` object once the data frame is nearly ready and complete?

SaguaroDev:
Pushed e45123f: rebased on main and dropped the `optional` tag from every parameter. Each line is now bare `name: type` per pandas convention. The `data` body note covers the empty-Triangle case ("If omitted, or if all arguments are omitted, an empty Triangle is returned") per the thread above.

On the de-facto-useless question: went with the neutral phrasing rather than calling it useless, since Gene flagged he'd be open to legitimate empty-Triangle use cases later. Want me to tighten it?

henrydingliu (approved):

Issue #824, opened 2026-05-20
[Test] Low test coverage on core.display.py


genedan:
### Description

`core.display.py` has coverage (67.11%) that's much lower than the project threshold of 85%.

<img width="3494" height="1113" alt="Image" src="https://github.com/user-attachments/assets/309308a9-2221-46e4-8c04-287174b02cb8" />

### Is your feature request aligned with the scope of the package?

- [x] Yes, absolutely!
- [ ] No, but it's still worth discussing.
- [ ] N/A (this request is not a codebase enhancement).

### Describe the solution you'd like, or your current workaround.

Write some tests for the missing lines.

genedan:
Closed via #825.

PR #825, opened 2026-05-20
[TEST] Add tests for TriangleDisplay.

genedan:
## Summary of Changes  

Adds unit tests for `core.display.py`

## Related GitHub Issue(s)  

#824

Adds a fixture called empty_triangle because it's used multiple times.

---

> [!NOTE]
> **Low Risk**
> Low risk: primarily adds test coverage and only changes runtime behavior by consistently raising `ImportError` when IPython/`HTML` is unavailable for `heatmap()`.
> 
> **Overview**
> Adds extensive unit tests for `TriangleDisplay` rendering and formatting, including `_dimensionality`, `__repr__`, `_repr_html_`, `_get_format_str`, semi-annual origin formatting, and `heatmap()` behavior.
> 
> Adjusts `heatmap()` to *always* raise `ImportError` (with updated message) when IPython’s `HTML` helper isn’t available, and adds an `empty_triangle` pytest fixture to reuse across tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b146e62f009a3a893ddb8994bfa169b4d3d4e98a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/chainladder/core/display.py
+++ b/chainladder/core/display.py
@@ -205,8 +205,8 @@ class TriangleDisplay:
             raise ValueError("heatmap() only works with a single triangle.")
         if HTML:
             return HTML(output_xnan)
-        elif HTML is None:
-            raise ImportError("heatmap requires IPython.")
+        else:
+            raise ImportError("heatmap() requires IPython.")
 
     @property
     def _dimensionality(self) -> str:
--- a/chainladder/core/tests/test_display.py
+++ b/chainladder/core/tests/test_display.py
@@ -1,20 +1,59 @@
 from __future__ import annotations
 
+import chainladder as cl
 import importlib
-import sys
-import pytest
 import numpy as np
 import pandas as pd
+import pytest
+import sys
 
-import chainladder as cl
 
-from unittest import mock
 from chainladder.core.display import TriangleDisplay
-
+from lxml import etree, html as lxml_html
+from unittest import mock
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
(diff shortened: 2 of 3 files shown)


henrydingliu:
My first impression is that `test_repr_multi` `test_repr_html_single` `test_repr_html_multi` doesn't test a whole lot. Maybe they don't need to. Just wanted to ask the question of, Is there some common mode of failure that we should be testing instead?

genedan:
`_repr_html()` is responsible for printing the triangles in Jupyter. The tests you listed only check whether there is some semblance of an HTML tag in there. I'll see if we can go one step further and validate the HTML syntax without bringing in additional dependencies.

I think we should keep it at that (for now), to be sure that the Triangle is displayed. The next step would be to actually see if the values in the representation are correct, but other functions handle the data, and fixing issues with those functions are more likely if the values are incorrect.

If we actually do start seeing issues in the Jupyter output caused by `_repr_html()`, we can take the tests further.

henrydingliu:
i'm good with that. we should circle back before we migrate all our docs to rst.

henrydingliu (dismissed):

genedan:
Added some testing for malformed HTML. I felt like even the parsing function would need its own test, so I added that too. `test_repr_multi` is the same though, it's just a quick and dirty check to see if we get a summary instead of a single-dim triangle.

henrydingliu (approved):
love the new tests for html integrity

Issue #839, opened 2026-05-23
[BUG] cl.Model_Diagnostics() does not show the correct Latest


henrydingliu:
### Description

john added a helper utility function for outputting chainladder results. however, it currently has a bug

```python
raa = cl.load_sample('raa')
print(raa)
print(cl.model_diagnostics(cl.Chainladder().fit(raa))['Latest'])
```
```python
         12       24       36       48       60       72       84       96       108      120
1981  5012.0   8269.0  10907.0  11805.0  13539.0  16181.0  18009.0  18608.0  18662.0  18834.0
1982   106.0   4285.0   5396.0  10666.0  13782.0  15599.0  15496.0  16169.0  16704.0      NaN
1983  3410.0   8992.0  13873.0  16141.0  18735.0  22214.0  22863.0  23466.0      NaN      NaN
1984  5655.0  11555.0  15766.0  21266.0  23425.0  26083.0  27067.0      NaN      NaN      NaN
1985  1092.0   9565.0  15836.0  22169.0  25955.0  26180.0      NaN      NaN      NaN      NaN
1986  1513.0   6445.0  11702.0  12935.0  15852.0      NaN      NaN      NaN      NaN      NaN
1987   557.0   4020.0  10946.0  12314.0      NaN      NaN      NaN      NaN      NaN      NaN
1988  1351.0   6947.0  13112.0      NaN      NaN      NaN      NaN      NaN      NaN      NaN
1989  3133.0   5395.0      NaN      NaN      NaN      NaN      NaN      NaN      NaN      NaN
1990  2063.0      NaN      NaN      NaN      NaN      NaN      NaN      NaN      NaN      NaN
          1990
1981  139826.0
1982   98203.0
1983  129694.0
1984  130817.0
1985  100797.0
1986   48447.0
1987   27837.0
1988   21410.0
1989    8528.0
1990    2063.0
```

genedan:
What do you think of adding it to the `__repr__` of the estimator? Printing it doesn't have any meaningful information:

```python
est = cl.Chainladder().fit(raa)
print(est)
```

```python
Chainladder()
```

I'm not all that familiar sci-kit learn syntax though, or if they have summaries of any kind. But it doesn't look like sklearn's estimators print meaningful information either:

```python
from sklearn.datasets import make_regression

X, y, true_weights = make_regression(
    n_samples=100,
    n_features=100,
    n_informative=10,
    noise=8,
    coef=True,
    random_state=42,
)


[...]

PR #852, opened 2026-05-26
Method summary

henrydingliu:
## Summary of Changes  
adding a summary property/attribute to ibnr methods,

a private function to assemble the exhibit in methodbase
a property in methodbase that points to the function above
an attribute to the predicted triangle (only needed for chainladder and benktander), again using the function above
added summary to the attribute section of the docstring of various methods
added a test across the various methods

## Related GitHub Issue(s)  
#839 

i went through various approaches:

- directly assembling the underlying values in numpy. didn't want to deal complications with the other backends
- using concat to bring together various triangles (ldf, ultimate, etc). main issue is that ultimate is not in a diagonal.
- ultimately used join on outputs of .to_frame(). .to_frame() eliminates rows where the value is null. this is deeply embedded in the implementation of .to_frame() via the sparse conversion. had to resort to join (which is very sparingly used in the codebase) to work around no null rows.

also streamlined some tests while i was in there, so i didn't have to write a bunch of identical tests for each method

- [X] I passed tests locally for both code (`uv run pytest`) and documentation changes (`uv run jb build docs --builder=custom --custom-builder=doctest`)

---

> [!NOTE]
> **Medium Risk**
> Renaming MackChainladder’s `summary_` to `mack_summary_` is a breaking API change for existing callers; the new join/melt summary logic affects all listed IBNR methods.
> 
> **Overview**
> Adds a shared **`summary_`** exhibit on IBNR estimators: a pandas table of **latest**, **ldf**, **cdf**, and **ultimate** per origin/column, built in **`MethodBase._get_summary`** from melted triangle frames and left joins (so ultimates without a reported amount still appear).
> 
> **`predict`** on **Chainladder** and **Benktander** attaches **`summary_`** on the returned triangle; docstrings for Cape Cod, BF, Expected Loss, etc. document the attribute. **Benktander** now **requires `sample_weight`** on **`predict`** (raises **`ValueError`** if missing).
> 
> **`MackChainladder`**: the old origin table (Latest / IBNR / Ultimate / Mack Std Err) is renamed to **`mack_summary_`**; **`summary_`** is the new generic exhibit. Tests and notebooks/gallery updated accordingly.
> 
> Predict tests are **parametrized** across several estimators to assert **`summary_`** matches **`latest_diagonal`** / **`ultimate_`** and enforces **`sample_weight`** where required.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1c3db5c95e47c21c0679e9e34e9d869fdb18e7cc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/chainladder/methods/base.py
+++ b/chainladder/methods/base.py
@@ -73,7 +73,7 @@ class MethodBase(BaseEstimator, EstimatorIO, Common):
         #columns for melt
         ids = X.key_labels + ['origin','development','valuation']
         #create dataframe for amount
-        amount = X.latest_diagonal.to_frame(implicit_axis=True,keepdims=True).reset_index().melt(id_vars=ids,var_name = 'column',value_name='actual')
+        amount = X.incr_to_cum().latest_diagonal.to_frame(implicit_axis=True,keepdims=True).reset_index().melt(id_vars=ids,var_name = 'column',value_name='actual')
         amount_index = X.key_labels + ['origin','column']
         amount = amount[amount_index + ['development','actual']]
         amount.set_index(amount_index,inplace=True)
--- a/chainladder/methods/base.py
+++ b/chainladder/methods/base.py
@@ -92,7 +92,7 @@ class MethodBase(BaseEstimator, EstimatorIO, Common):
         cdf.set_index(dev_factor_index,inplace=True)
         #assemble full summary. start from ultimate, as some methods (e.g. BF) return an ultimate without any actual amount
         output = ultimate.drop(columns=['development','valuation'])
-        output = output.join(amount,amount_index,how='left')
+        output = output.join(amount,on=amount_index,how='left')
         output = output.join(ldf,on=dev_factor_index,how='left')
         output = output.join(cdf,on=dev_factor_index,how='left')
         return output[X.key_labels + ['column','origin','development','actual','ldf','cdf','ultimate']]
--- a/chainladder/methods/base.py
+++ b/chainladder/methods/base.py
@@ -66,16 +66,16 @@ class MethodBase(BaseEstimator, EstimatorIO, Common):
             return self.X_.sum('development')
 
     @property
-    def summary(self):
+    def summary_(self):
         return self._get_summary(self.X_,self.ultimate_)
 
     def _get_summary(self,X,ult):
         #columns for melt
         ids = X.key_labels + ['origin','development','valuation']
         #create dataframe for amount
-        amount = X.incr_to_cum().latest_diagonal.to_frame(implicit_axis=True,keepdims=True).reset_index().melt(id_vars=ids,var_name = 'column',value_name='actual')
+        amount = X.incr_to_cum().latest_diagonal.to_frame(implicit_axis=True,keepdims=True).reset_index().melt(id_vars=ids,var_name = 'column',value_name='latest')
         amount_index = X.key_labels + ['origin','column']
-        amount = amount[amount_index + ['development','actual']]
+        amount = amount[amount_index + ['development','latest']]
         amount.set_index(amount_index,inplace=True)
         #create dataframe for ultimate
         ultimate = ult.to_frame(implicit_axis=True,keepdims=True).reset_index().melt(id_vars=ids,var_name = 'column',value_name='ultimate')
@@ -95,7 +95,7 @@ class MethodBase(BaseEstimator, EstimatorIO, Common):
         output = output.join(amount,on=amount_index,how='left')
         output = output.join(ldf,on=dev_factor_index,how='left')
         output = output.join(cdf,on=dev_factor_index,how='left')
-        return output[X.key_labels + ['column','origin','development','actual','ldf','cdf','ultimate']]
+        return output[X.key_labels + ['column','origin','development','latest','ldf','cdf','ultimate']]
 
     def fit(self, X, y=None, sample_weight=None):
         """Applies the chainladder technique to triangle **X**
--- a/chainladder/methods/base.py
+++ b/chainladder/methods/base.py
@@ -65,6 +65,38 @@ class MethodBase(BaseEstimator, EstimatorIO, Common):
         else:
             return self.X_.sum('development')
 
+    @property
+    def summary(self):
+        return self._get_summary(self.X_,self.ultimate_)
+
+    def _get_summary(self,X,ult):
+        #columns for melt
+        ids = X.key_labels + ['origin','development','valuation']
+        #create dataframe for amount
+        amount = X.latest_diagonal.to_frame(implicit_axis=True,keepdims=True).reset_index().melt(id_vars=ids,var_name = 'column',value_name='actual')
+        amount_index = X.key_labels + ['origin','column']
+        amount = amount[amount_index + ['development','actual']]
+        amount.set_index(amount_index,inplace=True)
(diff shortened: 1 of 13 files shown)


chainladder/methods/base.py
```diff
@@ -65,6 +65,38 @@ def latest_diagonal(self):
         else:
             return self.X_.sum('development')
 
+    @property
+    def summary(self):
```
kennethshsu: I think it should be `summary_` with the `_` to denote that the summary includes estimated values.

chainladder/methods/base.py
```diff
@@ -65,6 +65,38 @@ def latest_diagonal(self):
...
+        amount.set_index(amount_index,inplace=True)
+        #create dataframe for ultimate
+        ultimate = ult.to_frame(implicit_axis=True,keepdims=True).reset_index().melt(id_vars=ids,var_name = 'column',value_name='ultimate')
+        #columns for melt for dev factors
+        dev_ids = X.ldf_.key_labels + ['origin','development','valuation']
+        #create dataframe for ldf
+        ldf = X.ldf_.to_frame(implicit_axis=True,keepdims=True).reset_index().melt(id_vars=dev_ids,var_name = 'column',value_name='ldf')
+        dev_factor_index = X.ldf_.key_labels + ['development','column']
+        ldf = ldf[dev_factor_index + ['ldf']]
+        ldf.set_index(dev_factor_index,inplace=True)
```
kennethshsu, line 88: Is the LDFs useful? I think the CDFs are fine.

chainladder/methods/base.py
```diff
@@ -65,6 +65,38 @@ def latest_diagonal(self):
...
+        #create dataframe for cdf
+        cdf = X.cdf_.to_frame(implicit_axis=True,keepdims=True).reset_index().melt(id_vars=dev_ids,var_name = 'column',value_name='cdf')
+        cdf = cdf[dev_factor_index + ['cdf']]
+        cdf.set_index(dev_factor_index,inplace=True)
+        #assemble full summary. start from ultimate, as some methods (e.g. BF) return an ultimate without any actual amount
+        output = ultimate.drop(columns=['development','valuation'])
```
kennethshsu, line 94: I would also add the `ibnr_` in there.

chainladder/methods/base.py
```diff
@@ -65,6 +65,38 @@ def latest_diagonal(self):
...
 
+    @property
+    def summary(self):
+        return self._get_summary(self.X_,self.ultimate_)
+
+    def _get_summary(self,X,ult):
+        #columns for melt
+        ids = X.key_labels + ['origin','development','valuation']
+        #create dataframe for amount
+        amount = X.incr_to_cum().latest_diagonal.to_frame(implicit_axis=True,keepdims=True).reset_index().melt(id_vars=ids,var_name = 'column',value_name='actual')
```
kennethshsu: Can we call "Actual" "Latest" instead? Let's match the `MackChainladder` method. Here's an [example](https://github.com/casact/chainladder-python/blob/9e833f664b67296d8e85e729b9fe7cc8c8164efe/chainladder/methods/mack.py#L61-L68).

chainladder/methods/base.py
henrydingliu: doesn't have to be useful. we are trying to save the user extra work. manipulating ldf separately is not trivial.

chainladder/methods/base.py
henrydingliu: will do

chainladder/methods/base.py
henrydingliu: this actually came up a couple of weeks earlier. bugbot saw my friedland code and made the comment that ult - paid isn't IBNR. i think that's a valid point. ibnr_ isn't guaranteed to be IBNR.

chainladder/methods/base.py
henrydingliu: will do

chainladder/methods/base.py
kennethshsu: Oh ya that's another thing that bugs me, a better name would be unmerged or something like that. But this will be deprecate and move to a new name.

chainladder/methods/base.py
kennethshsu: When would the user ever need LDFs in the summary?

henrydingliu: easy example, if they want to calculate the expected emergence over the next calendar period

chainladder/methods/base.py
henrydingliu: also, unlike ldf, "ibnr" can just be df["ultimate"] - df["latest"], whenever the user wants.

chainladder/methods/base.py
kennethshsu: I still think it's not needed since in my mind `summary_` is a summary, it's the end of the workflow and it should tell me everything I want from the model/method. If I want the expected emergence it shouldn't come out of the summary. But we can keep it if you think it's useful.

chainladder/methods/base.py
kennethshsu: It depends, because most actuaries care about the "ibnr", or the "unemerged". 

I do think we should rename the column from `ibnr_` to 'unemerged_` or something like that. Similar in the Triangle class, `development` should be `valuation`, because `development` are actual ages - but this is a separate topic.

chainladder/methods/mack.py
```diff
@@ -27,8 +27,8 @@ class MackChainladder(Chainladder):
     full_triangle_:
         The ultimates back-filled to each development period in **X** retaining
         the known data
-    summary_:
-        summary of the model
+    mack_summary_:
```
kennethshsu, line 30: I understand the intend here, but I don't think we should change this right away as it will break the workflow. If we want to go this route, we need to throw a warning and deprecate appropriately. I think it's best to just leave as `summary_`.

kennethshsu:
@henrydingliu I am good with the `summary_` implementation as is, after you respond to my final round of feedback?

I don't like change of `mack_summary_`, can you consider reverting back?

chainladder/methods/base.py
henrydingliu: I agree that actuaries care about IBNR. I'm only not including it because I don't have a good name that generalizes between IBNR and unpaid. my reservation with unemerged is that no one uses it. 

https://www.casact.org/search?query=unemerged

chainladder/methods/mack.py
henrydingliu: i need clearer instructions then. these methods are inherited. i can't make mack have a different name than chainladder. we either go back to summary for this new dataframe, or change the old summary_ from mack to something else.

henrydingliu: what if i add st err to the summary_ just for mack, essentially merging the old summary_

kennethshsu: Ya that might be good...

So we have `summary_` for everything, and `std_err_` with actual numbers from Mack, and NaNs for deterministic models.

henrydingliu:
closing due to finding [this](https://github.com/casact/chainladder-python/blob/449b5c12db33c854a81906f3dabc2bb0cd5d0de1/chainladder/utils/utility_functions.py#L768). will rework and resubmit.

PR #868, opened 2026-05-27
#864 development constant tail bug

### LINKED ISSUE #864: DevelopmentConstant drops the tail factor

### Describe the bug in words

The `DevelompentConstant()` discards the last CDF to ultimate when the last CDF is not 1.0 (it sets it to 1.0).

```python
reported_patterns = {
    12: 4.0, 24: 2.9, 36: 1.8, 48: 1.4, 60: 1.2, 72: 1.1, 84: 1.03, 96: 1.02, 108: 1.005,
}
auto_bi = cl.load_sample("friedland_auto_bi_insurer")
reported_BI_claim = cl.DevelopmentConstant(
    patterns=reported_patterns, style="cdf"
).fit_transform(auto_bi["Reported Claims"])
reported_BI_claim.cdf_
```

The` 108-Ult` factor of 1.005 should not be discarded.

kennethshsu:
## Summary of Changes  
Addressed two bugs in the DevelopmentConstant()
1. When the supplied pattern is longer than the triangle, everything from the end of the tail is incorrectly discarded.
2. When the supplied pattern is a LDF (instead of CDF), the algorithm fails to convert it to CDF first.

## Related GitHub Issue(s)  
Fixes #864

This PR fixes both bugs, even though only 1 is reported on #864.
There was also an old bug(?) in `test_constant_callable_axis1`, not sure why `patterns.values` had `patterns.values[:, :-1]` dropped the last value. This is corrected.

---

> [!NOTE]
> **Medium Risk**
> Changes actuarial development-pattern fitting used before reserving; behavior shifts for long/short/LDF patterns but impact is confined to DevelopmentConstant with broad new tests.
> 
> **Overview**
> **`DevelopmentConstant.fit`** is reworked so external CDF/LDF patterns align correctly with triangle development length, including tails beyond the triangle.
> 
> A new **`_prepare_cdf_patterns`** converts LDF inputs to CDFs, splits an extra-age **tail factor**, and rebases in-triangle CDFs. **`fit`** uses that tail on the last LDF, chooses whether to keep the final dev period, fills short patterns with **1.0** (with a **`UserWarning`**), and applies the same preparation per row for **callable** patterns on index or columns.
> 
> Tests add coverage for no-tail vs tail, exact/short/long patterns (CDF and LDF), incremental triangles, and fix **`test_constant_callable_axis1`** to compare full CDF values.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5fdf5d8567ceefc483f3b4157fd342e8fb7083f6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


henrydingliu:
@kennethshsu there is some extensive refactoring in this PR, which is a little surprising given that both bugs appear to be fairly minor at a glance. could you please provide some commentary on the rationale and how the new code works?

kennethshsu:
Ya sure, I still don't really like the way it works currently, and I've certainly introduced some tech debt here, but I couldn't figure out how to fix this without messing up the fundamental design, which have two major flaws:

1) The method was written to work in mostly LDFs, and not CDFs. It should be obvious that CDFs would've worked a lot better, especially when the provided pattern is too short or too long, CDFs would work with much less manipulation.
2) The method assumes that the oldest origin year's latest valuation is ultimate. Not only does it assume that this latest val does not develop to ultimate, it also assumes there's no tail. Initially I solved this with `TailConstant()`, but that didn't work out in some edge cases so I decided to just rewrite.

I'll add some commentary in the diffs to help with the review, but it can probably be improved further.

chainladder/development/constant.py
```diff
@@ -39,56 +41,161 @@ def __init__(self, patterns=None, style="ldf", callable_axis=0, groupby=None):
...
+            Returns the instance itself.
         """
         from chainladder import options
-        if X.is_cumulative == False:
+
+        # convert to cumulative triangle
+        if not X.is_cumulative:
             obj = self._set_fit_groups(X).incr_to_cum().val_to_dev().copy()
         else:
             obj = self._set_fit_groups(X).val_to_dev().copy()
```
kennethshsu, line 89: This is mostly the same.

chainladder/development/constant.py
```diff
@@ -39,56 +41,161 @@ def __init__(self, patterns=None, style="ldf", callable_axis=0, groupby=None):
...
+            # on columns
             elif self.callable_axis == 1:
-                ldf = obj.columns.to_frame(index=False).apply(self.patterns, axis=1)
-                ldf = (
-                    pd.concat(ldf.apply(pd.DataFrame, index=[0]).values, axis=0)
-                    .fillna(1)[obj.ddims].values)
-                ldf = xp.array(ldf[None, :, None, :])
+                rows = obj.columns.to_frame(index=False)
             else:
-                raise ValueError('callable axis needs to be 0 or 1')
```
kennethshsu, line 76: On this left side, it jumps straight into working with the LDFs, this is not going to work if the pattern supplied does not match the triangles. This is the major flaw of issue 1 noted on the main thread.

chainladder/development/constant.py
```diff
@@ -39,56 +41,161 @@ def __init__(self, patterns=None, style="ldf", callable_axis=0, groupby=None):
...
+                raise ValueError("callable axis needs to be 0 or 1")
+
+            patterns = self.patterns(rows.iloc[0])
         else:
-            ldf = xp.array([float(self.patterns[item]) for item in obj.ddims])
+            # force the patterns to a dictionary
+            patterns = dict(self.patterns)
```
kennethshsu, line 107: On the right side, most cases is in the else case, where we are not working on the axis, it's just making sure that the patterns is in dict form.

In the callable case, I left most of it alone except I renamed to `rows`.

chainladder/development/constant.py
```diff
@@ -39,56 +41,161 @@ def __init__(self, patterns=None, style="ldf", callable_axis=0, groupby=None):
...
+
+        # seperate cdf_patterns and tail_cdf
+        cdf_patterns, tail_cdf = self._prepare_cdf_patterns(patterns, tri_dev_periods)
```
kennethshsu, line 111: This is the most important part of the function, is to make sure I get the pattern is CDF form, and separate out the tail. See the function above.

chainladder/development/constant.py
```diff
@@ -39,56 +41,161 @@ def __init__(self, patterns=None, style="ldf", callable_axis=0, groupby=None):
         self.callable_axis = callable_axis
         self.groupby = groupby
 
+    def _prepare_cdf_patterns(self, patterns, n_dev_periods):
+
+        patterns = dict(patterns)
+
+        sorted_keys = sorted(patterns.keys())
+        pattern_values = np.array([float(patterns[k]) for k in sorted_keys])
```
kennethshsu, line 49: Bugbot suggested this, but basically to sort the keys and make sure that the pattern supplied is reordered. It's a good suggestion, but I don't think a user would intentionally supplied an unordered pattern.

chainladder/development/constant.py
```diff
@@ -39,56 +41,161 @@ def __init__(self, patterns=None, style="ldf", callable_axis=0, groupby=None):
...
+
+        # build the cdf patterns, up to the needed periods
+        if self.style == "ldf":
+            cdf_patterns = dict(
+                zip(sorted_keys, np.cumprod(pattern_values[::-1])[::-1])
+            )
+        else:
+            cdf_patterns = {int(k): float(patterns[k]) for k in sorted_keys}
```
kennethshsu: Per comment, convert the patterns into CDF form.

chainladder/development/constant.py
```diff
@@ -39,56 +41,161 @@ def __init__(self, patterns=None, style="ldf", callable_axis=0, groupby=None):
...
+        if len(cdf_patterns) > n_dev_periods:
+            tail_cdf = float(cdf_patterns[sorted_keys[n_dev_periods]])
+
+            for k in sorted_keys[:n_dev_periods]:
+                cdf_patterns[int(k)] = cdf_patterns[int(k)] / tail_cdf
+
+            return cdf_patterns, tail_cdf
+
+        else:
+            return cdf_patterns, 1.0
```
kennethshsu: This is to separate out the tail factor, I needed to divide off the tail_cdf from the prior or we would "double-count" the tails.

If the supplied pattern is the same or less than the triangle development periods, set the tail to 1.

chainladder/development/constant.py
```diff
@@ -39,56 +41,161 @@ def __init__(self, patterns=None, style="ldf", callable_axis=0, groupby=None):
...
+            warnings.warn(
+                "Supplied patterns are shorter than the triangle development "
+                "periods. Missing ages will be filled with a factor of 1.0.",
+                UserWarning,
+                stacklevel=2,
+            )
+            tail_cdf = 1
+
+        elif len(cdf_patterns) == tri_dev_periods:
+            tail_cdf = 1
```
kennethshsu: This warning is new, I thought about not giving one and just do it quietly, but I think this is better since I did the same in throwing a warning when I put in additional assumptions for the users in the `cl.Development()` method.

kennethshsu: On second look I think I can discard the `tail_cdf = 1`. That's already set above.

chainladder/development/constant.py
```diff
@@ -39,56 +41,161 @@ def __init__(self, patterns=None, style="ldf", callable_axis=0, groupby=None):
...
+
+        pattern_dev_periods = len(cdf_patterns)
+
+        if pattern_dev_periods < tri_dev_periods:
+            include_last = False
+        elif pattern_dev_periods == tri_dev_periods:
+            include_last = True
```
kennethshsu: This is to address problem 2 in the main thread. Usually, the last age to age factor is not included (it assumes that the oldest origin row is at ultimate, e.g. if it's age 120, the last age to ultimate would be 102-ult, NOT 120-ult. Rewriting this broke a bunch of other things, so I kept that here.

If we need to somehow include 120-ult, because they are the same length, we'll do that here.

chainladder/development/constant.py
```diff
@@ -39,56 +41,161 @@ def __init__(self, patterns=None, style="ldf", callable_axis=0, groupby=None):
...
+        else:
+            include_last = tail_cdf != 1
+
+        dev_slice = slice(None) if include_last else slice(None, -1)
```
kennethshsu, line 137: This is to slice that development, depends on if we drop the last factor (e.g. 120-ult).

chainladder/development/constant.py
```diff
@@ -39,56 +41,161 @@ def __init__(self, patterns=None, style="ldf", callable_axis=0, groupby=None):
...
+
+        # this is the object to fill out the patterns, skeleton frame
+        obj = obj.iloc[..., :1, dev_slice] * 0 + 1
```
kennethshsu, line 140: Now, we are back to the old method, obj is just the blank age to age (NOT age to ult) to fill out.

chainladder/development/constant.py
```diff
@@ -39,56 +41,161 @@ def __init__(self, patterns=None, style="ldf", callable_axis=0, groupby=None):
...
+                .values
+            )
+            tail_cdfs = xp.array([item[1] for item in prepared])
+
+            if self.callable_axis == 0:
+                ldf = xp.array(ldf[:, None, None, :])
+                tail_cdfs = tail_cdfs[:, None, None]
+            else:
+                ldf = xp.array(ldf[None, :, None, :])
+                tail_cdfs = tail_cdfs[None, :, None]
```
kennethshsu, line 168: This is similar to the old stuff, I didn't touch.

chainladder/development/constant.py
```diff
@@ -39,56 +41,161 @@ def __init__(self, patterns=None, style="ldf", callable_axis=0, groupby=None):
...
+
+        if callable(self.patterns):
+
+            def _callable_row(row):
+                raw_patterns = self.patterns(row)
+                cdf_row, row_tail_cdf = self._prepare_cdf_patterns(
+                    raw_patterns, tri_dev_periods
+                )
+                fit_row = raw_patterns if self.style == "ldf" else cdf_row
+                return dict(fit_row), row_tail_cdf
```
kennethshsu, line 150: The `rows` stuff is from the above, when it's callable. This is super annoying is because the pattern needs to stay the same in either LDF or CDF form, but just get the tail out.

chainladder/development/constant.py
```diff
@@ -39,56 +41,161 @@ def __init__(self, patterns=None, style="ldf", callable_axis=0, groupby=None):
...
+                tail_cdfs = tail_cdfs[None, :, None]
+
+        else:
+            for ddim in obj.ddims:
+                if not any(ddim == k or int(ddim) == int(k) for k in cdf_patterns):
+                    cdf_patterns[int(ddim)] = 1.0
+                if self.style == "ldf" and not any(
+                    ddim == k or int(ddim) == int(k) for k in patterns
+                ):
+                    patterns[int(ddim)] = 1.0
```
kennethshsu: Fill in the cdf patterns with 1.0 if something is missing (i.e. blank spaces).

chainladder/development/constant.py
```diff
@@ -39,56 +41,161 @@ def __init__(self, patterns=None, style="ldf", callable_axis=0, groupby=None):
...
+
+            fit_patterns = patterns if self.style == "ldf" else cdf_patterns
```
kennethshsu: Still keeping track of both LDFs and CDFs.

chainladder/development/constant.py
```diff
@@ -39,56 +41,161 @@ def __init__(self, patterns=None, style="ldf", callable_axis=0, groupby=None):
...
+            ldf = xp.array([float(fit_patterns[int(item)]) for item in obj.ddims])
             ldf = ldf[None, None, None, :]
```
kennethshsu, line 179: Same as before.

chainladder/development/constant.py
```diff
@@ -39,56 +41,161 @@ def __init__(self, patterns=None, style="ldf", callable_axis=0, groupby=None):
...
+            tail_cdfs = tail_cdf
+
         if self.style == "cdf":
             ldf = xp.concatenate((ldf[..., :-1] / ldf[..., 1:], ldf[..., -1:]), -1)
+
+        # apply tail_cdf to the last ldfs of the triangle
+        ldf[..., -1] = ldf[..., -1] * tail_cdfs
```
kennethshsu, line 186: This is after following through everything, and just need to multiply the last element with the tail.

kennethshsu:
Hopefully this helps a bit. I can try to improve the readability to reduce tech debt a bit more tomorrow.

kennethshsu:
@henrydingliu do you have Claude code?

chainladder/development/constant.py
henrydingliu: i like this. by default, cdfs in a series is actually stored from oldest to newest.

chainladder/development/constant.py
henrydingliu: i can see why the existing code focuses on ldf. the entire package is written this way. cdf_ is a [property](https://github.com/casact/chainladder-python/blob/449b5c12db33c854a81906f3dabc2bb0cd5d0de1/chainladder/core/common.py#L94) that's derived off of ldf_ at runtime.

henrydingliu:
>Hopefully this helps a bit. I can try to improve the readability to reduce tech debt a bit more tomorrow.

it's a lot of catching of corner cases. i actually just had an idea after reviewing your other PR. you know how @genedan created a fake 2nd latest diagonal in all the friedland data? what if we just take that idea to implement `DevelopmentConstant()`? then we'd be leveraging all the corner case catching that's built into `Development()` itself. basically, when calling the `DevelopConstant().fit()`, you just create a triangle with all 1s, then override the latest diagonal with the pattern (and override the second latest diagonal with the pattern if cdf), and run a `Development(n_periods=1)` to get the `ldf_`. 

>do you have Claude code?

i develop in databricks. so i just use the built-in AI for basic syntax type acceleration/debugging. this is what the databricks AI tells me when I ask what model it is. 

I’m an OpenAI-provided assistant accessed through Databricks, but the specific model identifier may be abstracted away by the platform. If you need the exact deployment/model name, the best place to check is your Databricks or API configuration where this assistant was set up.

I also have a paid copilot 365. i do a lot of ideation and high-level solution design through that. Claude Opus is one of the models available.

kennethshsu:
> > do you have Claude code?
> 
> i develop in databricks. so i just use the built-in AI for basic syntax type acceleration/debugging. this is what the databricks AI tells me when I ask what model it is.
> 
> I’m an OpenAI-provided assistant accessed through Databricks, but the specific model identifier may be abstracted away by the platform. If you need the exact deployment/model name, the best place to check is your Databricks or API configuration where this assistant was set up.
> 
> I also have a paid copilot 365. i do a lot of ideation and high-level solution design through that. Claude Opus is one of the models available.

Hmm ok, I tried a few times to ask Cursor to refactor my code and it worked pretty terribly. I am going to go through another pass to do more cleanup but feel free to give it a try!

chainladder/development/tests/test_constant.py
```diff
@@ -23,36 +23,382 @@ def test_constant_ldf(raa):
...
+        120: 1.1,
+        132: 1.1,
+    }
+
+    result = cl.DevelopmentConstant(
+        patterns=reported_patterns, style="cdf"
+    ).fit_transform(raa)
+    assert np.all(
+        np.round(result.cdf_.to_frame().values.flatten(), 6)
+        == np.array([1.1, 1.1, 1.1, 1.1, 1.1, 1.1, 1.1, 1.1, 1.1, 1.1])
```
henrydingliu, line 412: isn't this the same problem as what you originally wanted to address? i.e. fed 11 cdf into developmentconstant but only get 10 back

henrydingliu: ah, okay. the original code only gave you 9 back

chainladder/development/constant.py
```diff
@@ -39,56 +41,153 @@ def __init__(self, patterns=None, style="ldf", callable_axis=0, groupby=None):
...
-        if X.is_cumulative == False:
+
+        # convert to cumulative triangle
+        if not X.is_cumulative:
             obj = self._set_fit_groups(X).incr_to_cum().val_to_dev().copy()
         else:
             obj = self._set_fit_groups(X).val_to_dev().copy()
+
         xp = obj.get_array_module()
-        obj = obj.iloc[..., :1, :-1]*0+1
```
henrydingliu, line 61: i think this is the only line you have to change. 

```python
if self.style == "cdf":
    obj = obj.iloc[..., :1, :]*0+1
else:
    obj = obj.iloc[..., :1, :-1]*0+1

Comment (kennethshsu) [reply chainladder/development/constant.py:61]: Nah, that's the lazy way. You aren't going to catch all the edge cases. Just bring my tests in and try your code. You'll fail a bunch.

Comment (kennethshsu) [reply chainladder/development/constant.py:61]: `obj = obj.iloc[..., :1, :]*0+1` is fundamentally changing the structure, this is now saying all development period will now develop one more, including the oldest origin period. 

And if it's LDF style, then you don't? This is wrong.

Comment (kennethshsu) [reply chainladder/development/constant.py:61]: Try to bring my tests in, and you can see if you can catch all the edge cases more cleanly. I'm sure there's a way.

I think for this PR, reviewing the tests is actually more important than the code itself.

Comment (henrydingliu) [reply chainladder/development/constant.py:61]: >Nah, that's the lazy way. You aren't going to catch all the edge cases. Just bring my tests in and try your code. You'll fail a bunch.

it definitely is the lazy way. it's also done in this PR, literally just a few lines down
```
obj = obj.iloc[..., :1, dev_slice] * 0 + 1
```

>Try to bring my tests in, and you can see if you can catch all the edge cases more cleanly

good idea. i can do that

Comment (kennethshsu) [reply chainladder/development/constant.py:61]: Oh and if you want to try to use the tests and let AI solve it, you can give that a try if you have access to a good AI agent.

Cursor couldn't do it and just kept iterating itself until I killed it.

File: chainladder/development/tests/test_constant.py [hunk 20]
```diff
@@ -23,36 +23,382 @@ def test_constant_ldf(raa):
...
+                2.593742,
+                2.357948,
+                2.143589,
+                1.948717,
+                1.771561,
+                1.61051,
+                1.4641,
+                1.331,
+                1.21,
+            ]
```
henrydingliu, line 473: i think this test actually shows a bug. the ldf_ would show 120-132 to be 1.21, which is not what was originally supplied.

chainladder/development/tests/test_constant.py
henrydingliu: on third thought, i think this test actually shows a bug. the ldf_ would show 120-132 to be 1.1, which is not what was originally supplied.

chainladder/development/constant.py
henrydingliu: I implemented an alternative approach in this [branch](https://github.com/casact/chainladder-python/tree/%23864_alt_imp). you can find the diff [here](https://github.com/casact/chainladder-python/pull/878/changes). I couldn't figure out how to pass all of your tests. So instead, I'm choosing to declare two of your tests defective :P

chainladder/development/tests/test_constant.py
kennethshsu: Sorry, you think the RHS is wrong? Why? The cdf from 120-ult is 1.1, as supplied?

chainladder/development/tests/test_constant.py
kennethshsu: The CDF, not LDF, is 1.21. Line 365 checks the CDF.

chainladder/development/constant.py
kennethshsu: Ha, it's very possible I made mistakes on the tests. So let's make sure the tests are right. 

Your implementation is much cleaner, let's just make sure it can catch all the edge cases. I think we are super close.

chainladder/development/tests/test_constant.py
henrydingliu: i'm not saying that the rhs is wrong. i'm saying this test isn't capturing an error in the untested ldf_. the ldf_ at 120 coming out of `DevelopmentConstant` is 1.1, but the actual ldf at 120 that was supplied to the estimator was 1.

chainladder/development/tests/test_constant.py
henrydingliu: yes, the cdf is 121. that is correct. but the underlying ldf_ at 120 is also 1.21, which is not what was supplied to the estimator

chainladder/development/constant.py
henrydingliu: i think each test needs to test both the ldf_ and the cdf_. we'd just have to manually calculate the cdf/ldf from the supplied vector.

chainladder/development/tests/test_constant.py
kennethshsu: Let me check... I will add tests to check both the `cdf_` and `ldf_`

chainladder/development/tests/test_constant.py
```diff
@@ -340,6 +411,10 @@ def test_constant_pattern_long_cdf(raa):
         np.round(result.cdf_.to_frame().values.flatten(), 6)
         == np.array([1.1, 1.1, 1.1, 1.1, 1.1, 1.1, 1.1, 1.1, 1.1, 1.1])
     )
+    assert np.all(
+        np.round(result.ldf_.to_frame().values.flatten(), 6)
+        == np.array([1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.1])
```
henrydingliu, line 416: this doesn't match "reported_pattern"

chainladder/development/tests/test_constant.py
```diff
@@ -363,18 +438,38 @@ def test_constant_pattern_long_ldf(raa):
...
+                1.1,
+                1.1,
+                1.1,
+                1.1,
+                1.1,
+                1.1,
+                1.1,
+                1.1,
+                1.1,
+                1.1**2,
```
henrydingliu, line 472: this doesn't match `reported_patterns`

kennethshsu: It does? The only thing that's different is the last LDF, those need to be grouped, or it will be incorrectly discarded.

chainladder/development/tests/test_constant.py
kennethshsu: It's not supposed to be, the `reported_pattern` is in CDF form, but the LHS/RHS check here is in LDF form.

kennethshsu: Let me know if you agree lol. 

I had to be super careful, made many mistakes before. Just double check to see if you are aligned.

henrydingliu: the 1.1 comes at the 10th in the returned ldf_, instead of the 11th element implied by the original pattern

chainladder/development/tests/test_constant.py
henrydingliu: there's no 1.21 in the original ldf pattern.

kennethshsu: Yes because the pattern extends beyond what is needed by the data object. Are you suggesting that you just discard the last one?

chainladder/development/tests/test_constant.py
kennethshsu: Because the data object doesn't have the 11th origin period?

henrydingliu: right. so we have an issue here because the ldf_ returns changes, beyond filling with additional 1.0's, depending on how large the triangle is. in theory, we would want `DevelopmentConstant().fit(trI_9x9).ldf_[:-1]` to be equal to `DevelopmentConstant().fit(trI_8x8).ldf_` but that's not happening

chainladder/development/tests/test_constant.py
henrydingliu: no, the PR is what's discarding the last ldf. i'm saying we need to keep the last ldf, that would make it consistent with `Development`

```python
raa = cl.load_sample('raa')
dev = cl.Development().fit(raa)
raa_1987 = raa[(raa.valuation <= '1988-01-01')]
print(dev.transform(raa).ldf_)
print(dev.transform(raa_1987).ldf_)
```
```python
          12-24     24-36     36-48     48-60     60-72     72-84     84-96    96-108   108-120
(All)  2.999359  1.623523  1.270888  1.171675  1.113385  1.041935  1.033264  1.016936  1.009217
          12-24     24-36     36-48     48-60     60-72     72-84     84-96    96-108   108-120
(All)  2.999359  1.623523  1.270888  1.171675  1.113385  1.041935  1.033264  1.016936  1.009217
```

chainladder/development/tests/test_constant.py
kennethshsu: Let me reply you on the main thread.

kennethshsu:
> right. so we have an issue here because the ldf_ returns changes, beyond filling with additional 1.0's, depending on how large the triangle is. in theory, we would want `DevelopmentConstant().fit(trI_9x9).ldf_[:-1]` to be equal to `DevelopmentConstant().fit(trI_8x8).ldf_` but that's not happening

So I think the above is only true if the pattern supplied is shorter than a length of 7. Try this example, and comment/uncomment the pattern length, everything is working as expected to me. 

When pattern is `ldf` style:
```python
raa10x10 = cl.load_sample("raa")
raa9x9 = raa10x10[raa10x10.valuation<'1990']

reported_patterns = {
    12: 1.1,
    24: 1.1,
    36: 1.1,
    48: 1.1,
    60: 1.1,
    72: 1.1,
    84: 1.1,
    96: 1.1,
    108: 1.1,
    # 120: 1.1,
    # 132: 1.1,
}

raa10x10_ldf_ = cl.DevelopmentConstant(
    patterns=reported_patterns, style="ldf"
).fit_transform(raa10x10).ldf_
raa9x9_ldf_ = cl.DevelopmentConstant(
    patterns=reported_patterns, style="ldf"
).fit_transform(raa9x9).ldf_
print(raa10x10_ldf_)
print(raa9x9_ldf_)
```

In `cdf` style:
```python
raa10x10 = cl.load_sample("raa")
raa9x9 = raa10x10[raa10x10.valuation<'1990']

reported_patterns = {
    12: 1.1,
    24: 1.1,
    36: 1.1,
    48: 1.1,
    60: 1.1,
    72: 1.1,
    84: 1.1,
    96: 1.1,
    108: 1.1,
    # 120: 1.1,
    # 132: 1.1,
}

raa10x10_ldf_ = cl.DevelopmentConstant(
    patterns=reported_patterns, style="cdf"
).fit_transform(raa10x10).ldf_
raa9x9_ldf_ = cl.DevelopmentConstant(
    patterns=reported_patterns, style="cdf"
).fit_transform(raa9x9).ldf_
print(raa10x10_ldf_)
print(raa9x9_ldf_)
```

kennethshsu:
Sorry, but I am not following your question/feedback, but can you give me an example of where you think my implementation doesn't give the correct answer, and what the correct answer should be?

henrydingliu:
> > right. so we have an issue here because the ldf_ returns changes, beyond filling with additional 1.0's, depending on how large the triangle is. in theory, we would want `DevelopmentConstant().fit(trI_9x9).ldf_[:-1]` to be equal to `DevelopmentConstant().fit(trI_8x8).ldf_` but that's not happening
> 
> So I think the above is only true if the pattern supplied is shorter than a length of 7. Try this example, and comment/uncomment the pattern length, everything is working as expected to me.
> 
> When pattern is `ldf` style:
> 
> ```python
> raa10x10 = cl.load_sample("raa")
> raa9x9 = raa10x10[raa10x10.valuation<'1990']
> 
> reported_patterns = {
>     12: 1.1,
>     24: 1.1,
>     36: 1.1,
>     48: 1.1,
>     60: 1.1,
>     72: 1.1,
>     84: 1.1,
>     96: 1.1,
>     108: 1.1,
>     # 120: 1.1,
>     # 132: 1.1,
> }
> 
> raa10x10_ldf_ = cl.DevelopmentConstant(
>     patterns=reported_patterns, style="ldf"
> ).fit_transform(raa10x10).ldf_
> raa9x9_ldf_ = cl.DevelopmentConstant(
>     patterns=reported_patterns, style="ldf"
> ).fit_transform(raa9x9).ldf_
> print(raa10x10_ldf_)
> print(raa9x9_ldf_)
> ```
> 
> In `cdf` style:
> 
> ```python
> raa10x10 = cl.load_sample("raa")
> raa9x9 = raa10x10[raa10x10.valuation<'1990']
> 
> reported_patterns = {
>     12: 1.1,
>     24: 1.1,
>     36: 1.1,
>     48: 1.1,
>     60: 1.1,
>     72: 1.1,
>     84: 1.1,
>     96: 1.1,
>     108: 1.1,
>     # 120: 1.1,
>     # 132: 1.1,
> }
> 
> raa10x10_ldf_ = cl.DevelopmentConstant(
>     patterns=reported_patterns, style="cdf"
> ).fit_transform(raa10x10).ldf_
> raa9x9_ldf_ = cl.DevelopmentConstant(
>     patterns=reported_patterns, style="cdf"
> ).fit_transform(raa9x9).ldf_
> print(raa10x10_ldf_)
> print(raa9x9_ldf_)
> ```

right. when the supplied ldf or cdf is longer, the resulting ldf_ is distorted. right now you are forcing those tests to pass by comparing the resulting ldf_ to something other than what was supplied.

henrydingliu:
> Sorry, but I am not following your question/feedback, but can you give me an example of where you think my implementation doesn't give the correct answer, and what the correct answer should be?

this implementation doesn't give the right answer when the supplied pattern is longer. the correct answer should replicate the supplied ldf or the implied ldf of the suppled cdf **exactly** (not counting factors of 1 to fill the space). 

basically, if elsewhere in the package we want stuff like `read_json(to_jason)` to give us back the original. then DevelopmentConstant(pattern).ldf_ should always give back the pattern. whether the fitted triangle is longer or shorter is irrelevant. this is evidenced by 

```python
raa = cl.load_sample('raa')
dev = cl.Development().fit(raa)
raa_1987 = raa[(raa.valuation <= '1988-01-01')]
print(dev.transform(raa).ldf_)
print(dev.transform(raa_1987).ldf_)
```

kennethshsu:
> > Sorry, but I am not following your question/feedback, but can you give me an example of where you think my implementation doesn't give the correct answer, and what the correct answer should be?
> 
> this implementation doesn't give the right answer when the supplied pattern is longer. the correct answer should replicate the supplied ldf or the implied ldf of the suppled cdf **exactly** (not counting factors of 1 to fill the space).

I don't think I agree with this. You are saying to just discard the extra LDF pattern beyond the triangle object? Look at this example:

```python
raa10x10 = cl.load_sample("raa")
raa9x9 = raa10x10[raa10x10.valuation<'1990']

reported_patterns = {
    12: 1.1,
    24: 1.1,
    36: 1.1,
    48: 1.1,
    60: 1.1,
    72: 1.1,
    84: 1.1,
    96: 1.1,
    108: 1.1, #9th LDF, or if this is a CDF, it would've been 1.21
    120: 1.1, #10th's LDF, not CDF
    # 132: 1.1,
}

raa10x10_ldf_ = cl.DevelopmentConstant(
    patterns=reported_patterns, style="ldf"
).fit_transform(raa10x10).ldf_
raa9x9_ldf_ = cl.DevelopmentConstant(
    patterns=reported_patterns, style="ldf"
).fit_transform(raa9x9).ldf_
print(raa10x10_ldf_)
print(raa9x9_ldf_)
```
Returns:
```python
       12-24  24-36  36-48  48-60  60-72  72-84  84-96  96-108  108-120  120-132
(All)    1.1    1.1    1.1    1.1    1.1    1.1    1.1     1.1      1.1      1.1
       12-24  24-36  36-48  48-60  60-72  72-84  84-96  96-108  108-120
(All)    1.1    1.1    1.1    1.1    1.1    1.1    1.1     1.1     1.21
                                                                  # ^ this 9th's LDF should 
                                                                  # be 1.1 instead of 1.21
```
You are saying, we should just discard the remaining `1.1` ldf from `120-132` (`120-ult`) if the triangle data is shorter?

> basically, if elsewhere in the package we want stuff like `read_json(to_jason)` to give us back the original. then DevelopmentConstant(pattern).ldf_ should always give back the pattern. whether the fitted triangle is longer or shorter is irrelevant. this is evidenced by
> 
> ```python
> raa = cl.load_sample('raa')
> dev = cl.Development().fit(raa)
> raa_1987 = raa[(raa.valuation <= '1988-01-01')]
> print(dev.transform(raa).ldf_)
> print(dev.transform(raa_1987).ldf_)
> ```

This is different, in this example, you are estimating patterns using a set of data, and estimating another pattern using a subset of that data, it should be clear that the pattern estimated using the subset lacks something (i.e. the tail).

Let me try to se if I can convince you with another example:

Do you agree that these two patterns are the same?
```python
DC_LDF = cl.DevelopmentConstant(
    patterns={
    12: 1.1,
    24: 1.1,
    36: 1.1,
    48: 1.1,
    60: 1.1,
    72: 1.1,
    84: 1.1,
    96: 1.1,
    108: 1.1,
    120: 1.1,
}, style="ldf"
)
DC_CDF = cl.DevelopmentConstant(
    patterns={
    12: 1.1**10,
    24: 1.1**9,
    36: 1.1**8,
    48: 1.1**7,
    60: 1.1**6,
    72: 1.1**5,
    84: 1.1**4,
    96: 1.1**3,
    108: 1.1**2,
    120: 1.1,
}, style="cdf"
)
```
If so, then you should get the same CDF, not LDF, no matter what triangle object you fit on.

Again, in my opinion, one of the current implementation flaw that I think is that if the pattern provided is shorter, and it is in LDF form, the extra pattern is basically discarded. My PR fixes that.

kennethshsu:
@henrydingliu do you agree with my latest post? 

I am trying to resolve the last bug bot comment.

kennethshsu:
@henrydingliu do you have notifications muted? 😜

henrydingliu:
sorry, everything sends a notification and this got buried in the pile.

henrydingliu:
> @henrydingliu do you agree with my latest post?
> 
> I am trying to resolve the last bug bot comment.

i think we both want the same outcome. it's actually this implementation that isn't following your latest post. this implementation is discarding LDFs . it's discarding LDFs [here](https://github.com/casact/chainladder-python/pull/868#discussion_r3326889701) and [here](https://github.com/casact/chainladder-python/pull/868#discussion_r3326887731). i'm trying to say that it shouldn't be discarding anything.

kennethshsu:
Ok I think I finally understood where the misalignment is after re-reading many times because I definitely said some weird stuff from misreading lol... Sorry! But let me restate the problem first, and we can decide where to go.

I actually don't think there's **ANY** bug in this implementation, but the question is about how it's implemented. **Let me know if you disagree.**

The different view is actually simple, it's how we treat `patterns` in `DevelopmentConstant()` that goes beyond the triangle object.

My proposal and the current PR is that everything that goes beyond the triangle object are consolidated into a single tail CDF, and this CDF is assigned to the last age.

@henrydingliu's proposal is that we just assign the `patterns` as is, and let it extend beyond the triangle object as needed. 

Consider this example:

```python
raa10x10 = cl.load_sample("raa")
raa5x5 = raa10x10[raa10x10.valuation<'1986']

patterns_len_10 = {
    12: 1.1,
    24: 1.1,
    36: 1.1,
    48: 1.1,
    60: 1.1,
    72: 1.1,
    84: 1.1,
    96: 1.1,
    108: 1.1,
    120: 1.1, 
}

raa10x10_factors = cl.DevelopmentConstant(
    patterns=patterns_len_10, style="ldf"
).fit_transform(raa10x10)
raa5x5_factors = cl.DevelopmentConstant(
    patterns=patterns_len_10, style="ldf"
).fit_transform(raa5x5)

print("Full tri")
print(raa10x10_factors.cdf_)
print(raa10x10_factors.ldf_)

print("\nSmall tri")
print(raa5x5_factors.cdf_)
print(raa5x5_factors.ldf_)
```

My implemntation returns:
```python
Full tri
         12-Ult    24-Ult    36-Ult    48-Ult    60-Ult   72-Ult  84-Ult  96-Ult  108-Ult  120-Ult
(All)  2.593742  2.357948  2.143589  1.948717  1.771561  1.61051  1.4641   1.331     1.21      1.1
       12-24  24-36  36-48  48-60  60-72  72-84  84-96  96-108  108-120  120-132
(All)    1.1    1.1    1.1    1.1    1.1    1.1    1.1     1.1      1.1      1.1

Small tri
         12-Ult    24-Ult    36-Ult    48-Ult    60-Ult
(All)  2.593742  2.357948  2.143589  1.948717  1.771561
       12-24  24-36  36-48  48-60     60-72
(All)    1.1    1.1    1.1    1.1  1.771561
```
Note the `60-72` LDF is actually the CDF from `72-Ult`.

But I think @henrydingliu you wanted
```python
Full tri
         12-Ult    24-Ult    36-Ult    48-Ult    60-Ult   72-Ult  84-Ult  96-Ult  108-Ult  120-Ult
(All)  2.593742  2.357948  2.143589  1.948717  1.771561  1.61051  1.4641   1.331     1.21      1.1
       12-24  24-36  36-48  48-60  60-72  72-84  84-96  96-108  108-120  120-132
(All)    1.1    1.1    1.1    1.1    1.1    1.1    1.1     1.1      1.1      1.1

Small tri
         12-Ult    24-Ult    36-Ult    48-Ult    60-Ult   72-Ult  84-Ult  96-Ult  108-Ult  120-Ult
(All)  2.593742  2.357948  2.143589  1.948717  1.771561  1.61051  1.4641   1.331     1.21      1.1
       12-24  24-36  36-48  48-60  60-72  72-84  84-96  96-108  108-120  120-132
(All)    1.1    1.1    1.1    1.1    1.1    1.1    1.1     1.1      1.1      1.1
```

Basically, the `patterns` in `DevelopmentConstant()` is copied directly to the object, and extended the development ages as needed.

I don't think there's any other disagreement except this one.

@henrydingliu do you agree with the summary above? Once we are aligned here, we can discuss which one to go with. I don't think there's any other misalignment.

By the way, I think it is time we move forward with test based development or whatever it's called. We can break PRs into two parts, the tests, and the implementation. That way we don't have misalignment on the implementation and having to go back and redo. This is basically the problem of not having good spec sheet. I'll add this to

henrydingliu:
>I don't think there's any other disagreement except this one.

yessir

>I think it is time we move forward with test based development or whatever it's called

i do think test-based development is a good idea in general. a couple of concerns based on our current scenario

- test-based development wouldn't have helped with this particular PR. you found more corner cases to test as you developed, which then spun into our protracted discussion here. but maybe that's okay. 
- i'm not sure our contributor community is large enough at the present to truly support this operating model. maybe we just make an edit to the issue template to ask for a test (it's already asking for desired behavior, so a small wording change)?

kennethshsu:
> > I think it is time we move forward with test based development or whatever it's called
> 
> i do think test-based development is a good idea in general. a couple of concerns based on our current scenario
> 
> * test-based development wouldn't have helped with this particular PR. you found more corner cases to test as you developed, which then spun into our protracted discussion here. but maybe that's okay.
> * i'm not sure our contributor community is large enough at the present to truly support this operating model. maybe we just make an edit to the issue template to ask for a test (it's already asking for desired behavior, so a small wording change)?

Ya all good point, I might just try to do this personally myself. It's something I learned in school but never practiced. I might give it a try.
> > I don't think there's any other disagreement except this one.
> 
> yessir
> 

Ok perfect!! So my implementation, even though it looks complex in terms of the algorithm, it's actually simpler in implementation. Your suggestion that we extend beyond the `obj` is actually harder because we would need expand the triangle `obj` to something that doesn't exist yet. This is different than your example:

```python
raa = cl.load_sample('raa')
dev = cl.Development().fit(raa)
raa_1987 = raa[(raa.valuation <= '1988-01-01')]
print(dev.transform(raa).ldf_)
print(dev.transform(raa_1987).ldf_)
```

In the above example, the "triangle" (pattern) is already built on the full grid, and we are just bringing that full grid into the smaller triangle, so the ddim is already there.

I do think your implementation has one benefit though, is that we can get the next 12 month expected emergence. But even with that, I still think we should merge this as is given that there's not a bug here (so I can unblock my Friedland recon work) and we can consider opening a new ticket for this feature. Thoughts?

henrydingliu:
you know, this whole thing started when you said that you weren't entirely happy with this, and it's probably introducing future tech debt. 

>In the above example, the "triangle" (pattern) is already built on the full grid, and we are just bringing that full grid into the smaller triangle, so the ddim is already there.

my preference is to not use the result as the rationale to compromise on a principle. i provided this example to illustrate that the existing behavior within the development estimators is to let the full pattern ride. we can definitely deviate from that behavior if that behavior doesn't make sense in a given situation. but we don't seem to be making an argument for that. so do you feel good justifying a deviation in package behavior with "well cuz it'd be hard otherwise..."

>But even with that, I still think we should merge this as is given that there's not a bug here (so I can unblock my Friedland recon work) and we can consider opening a new ticket for this feature.

unfortunately this is the one place I disagree with you on. to me, this truncation of the LDF is indeed a bug, because it's altering the input data. _you cannot get back the original input pattern using the newly created `DevelopmentConstant` object. `pattern = cl.DevelopmentConstant(pattern).fit(tri).ldf_` should be true regardless of size of `tri`. it's not actually passing all of the tests. we are just shoehorning it through by doctoring a couple of the asserts. 

you can proceed with your friedland work by adding a tailconstant to the existing developmentconstant. 

>I do think your implementation has one benefit though, is that we can get the next 12 month expected emergence.

that was just to show that it's possible. my [preferred](https://github.com/casact/chainladder-python/pull/868#issuecomment-4561238157) implementation would be a complete rewrite, and pivoting to creating dummy triangles using the supplied pattern and strategically placed 1's to leverage all the robustness this package has around `Triangle` and the base `Development` estimator.

kennethshsu:
Ya I think I agree with you now, you convinced me.

What's the idea of using the fake triangles? So you create a new triangle with the right shape with the latest diagonal just being 1s? And then the 2nd to last is whatever works out to the correct LDF/CDF?

henrydingliu:
> Ya I think I agree with you now, you convinced me.
> 
> What's the idea of using the fake triangles? So you create a new triangle with the right shape with the latest diagonal just being 1s? And then the 2nd to last is whatever works out to the correct LDF/CDF?

something like that. the cdf case is easier, just set the whole triangle to the cdf based on age. then the ldf case is just cumprod into the cdf case.

Issue #866, opened 2026-05-27
Update Options.describe_option()


genedan:
`Options.describe_option()` currently doesn't do anything. It looks like it was planned to be implemented:

https://github.com/casact/chainladder-python/blob/23340299500236af6dd0423c1c23aa9393a43dbb/chainladder/__init__.py#L88-L89

This issue can be completed by mirroring the behavior of `pandas.describe_option`.

Issue #874, opened 2026-05-28
Move options functions to top level


genedan:
Discovering more work to do...options functions like `get_option` and `set_option` are top-level in pandas:

```python
In [5]: pd.get_option("display.chop_threshold")

In [6]: pd.set_option("display.chop_threshold", 2)

In [7]: pd.get_option("display.chop_threshold")
Out[7]: 2

In [8]: pd.set_option("chop", 4)

In [9]: pd.get_option("display.chop_threshold")
Out[9]: 4
```

They aren't in chainladder:

```python
cl.options.get_option(...)
```

Let's see if we can improve the consistency of the API.

Issue #855, opened 2026-05-26
Governing doc, contribution guidelines, and enhanced PR template


kennethshsu:
### Description

My goal is for me personally is to make sure this package stays robust and maintains consistent, high-quality PR reviews. With the rise of AI generated slop projects on GitHub, I want to make sure this clearly isn’t one of them.

I’d like to hear thoughts on what guardrails we should put in place around PRs and AI usage.

I opened a [PR](https://github.com/casact/chainladder-python/pull/827/changes) on this a while back, but I’d love feedback on what others would like to see. Some of the language, especially around credentials, may be stronger than needed, and I’m certainly not attached to any specific wording. The goal is not to limit PRs or participation, but to maintain a high standard of quality for contributions and reviews.

### Is your feature request aligned with the scope of the package?

- [ ] Yes, absolutely!
- [ ] No, but it's still worth discussing.
- [x] N/A (this request is not a codebase enhancement).

### Describe the solution you'd like, or your current workaround.

- Implement new contributing guidelines, especially regarding AI tools usage
- Improve the PR template 
- (maybe) How does one become a code owner (and therefore an approved PR reviewer)

genedan:
Bringing in my thoughts from other discussions. I think the contribution guidelines should be modeled after those of other big open source projects such as:

- [Python Enhancement Proposals (PEPs)](https://peps.python.org/)
- [Pandas Contributing Guide](https://pandas.pydata.org/docs/development/contributing.html)
- [Pandas Enhancement Proposals (PDEPs)](https://pandas.pydata.org/pdeps/0001-purpose-and-guidelines.html)

Our project is smaller, so we should just take what we need and then grow the guide as needed. I had proposed a set of CEPs (Chainladder Enhancement Proposals):

- CEP 1: Purpose and Guidelines (contains info about codeowners, continuity planning, voting rules for resolving disputes, adopting new features, etc.)
- CEP 1: AI Usage
- CEP 2: Pull Request review guidelines
- CEP 3: Docstring conventions

Other sub-pages for the contributing guides would include things for:

- Documentation formatting
- Unit test conventions
- Class design & deep-dive internals
- Deprecation guidelines
- Release schedule checklist & guidelines

Our current contributing page is a "here's the gist of what you need to know to get started", which I think should stay that way, but we should move it to a top-level page and add subpages, which can get into the details. Our practices have grown organically, and a lot of them aren't written down, or at least not in a central location.

[...]

PR #877, opened 2026-05-28
FEAT: Implement cl.options.describe_option.

genedan:
## Summary of Changes  

Implement `cl.options.describe_option()` to mirror the behavior of `pd.describe_option`.

Also renamed `options` parameter to `pat` in the options functions to match the pandas API.

## Related GitHub Issue(s)  

#866

The regex is complicated. I did ask Claude to write it and explain (see below). I summarized the match within the function but I can also add this detailed piece-by-piece explanation if you think it's needed.

```shell
  - ^{option}: — matches the option name (e.g. ARRAY_BACKEND:) at the start of a line, thanks to re.MULTILINE
  - \s* — allows optional whitespace between the colon and the type
  - (\S+) — captures the type annotation (e.g. str, bool, list) — one or more non-whitespace characters, in capture group 1
  - \n — matches the newline ending that first line
  - ((?:[ \t]+.+\n?)+) — captures the description block in group 2:
    - [ \t]+ — one or more spaces or tabs (the indentation that marks a description line)
    - .+ — one or more characters (the description text)
    - \n? — optional newline at the end of the line
    - (?:...)+ — the whole thing repeats one or more times to capture multi-line descriptions, with (?:...) being a non-capturing group so it doesn't interfere with capture group 2 
```

Here's what the output should look like to help:

```python
cl.options.describe_option("AUTO_SPARSE")
```

```shell
AUTO_SPARSE : bool
      Controls whether chainladder automatically converts a triangle's backing array to a sparse representation
      when it would be memory-efficient to do so.
      [default: True] [currently: True]
```

```python
cl.options.describe_option('AUTO_SPARSE|ARRAY_BACKEND')
```

```shell
ARRAY_BACKEND : str
    The default array backend for chainladder.
    [default: numpy] [currently: numpy]
AUTO_SPARSE : bool
    Controls whether chainladder automatically converts a triangle's backing array to a sparse representation
    when it would be memory-efficient to do so.
    [default: True] [currently: True]
```

---

> [!NOTE]
> **Low Risk**
> Configuration and documentation API only, with a deprecation path for the old `option` keyword; no changes to reserving or data handling logic.
> 
> **Overview**
> Adds **`cl.options.describe_option()`** in the style of pandas: it prints or returns option metadata (type and description from the `Options` class docstring, plus **default** and **current** values). **`pat`** accepts an empty string for all options, a single name, or a **regex** (e.g. `AUTO_SPARSE|ARRAY_BACKEND`); invalid names or bad regex raise **`ValueError`**.
> 
> **`get_option`**, **`set_option`**, and **`reset_option`** now take **`pat`** instead of a positional **`option`** argument. The old **`option=`** keyword still works with a **`FutureWarning`** via **`_resolve_pat`**, which also errors if both **`pat`** and **`option`** are set or if required arguments are missing (including **`value`** on **`set_option`**). **`set_option`** only treats **`ARRAY_PRIORITY`** cupy warnings when **`value`** is a list.
> 
> Tests in **`test_utilities.py`** cover describe output, deprecation, and error paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0c4e7615d1b2c112a0b02e62efbbc8254f541b34. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


kennethshsu:
Do we need proper deprecation cycle to go from `options` to `pat`? I've personally never used this, but I think we should be cautious.

genedan:
Yeah that's a public API change. So would #874. I'll rename the parameters back to `options` and then we can mark the argument name change and #874 for milestone 1.0, and then add a warning. For #855 we can decide on how long those warnings will stick around.

I wonder if not-so-big changes like the name of a parameter would warrant a major bump (let's say, the next time we do this, does the package have to be 2.0?). I think it would actually, if we wanted to be strict about things, since backwards-incompatible changes are what warrant a major bump.

henrydingliu:
we could also be verbose and make old names into properties that return the new name.

kennethshsu:
I know we haven't quite figure out deprecation cycle rules, but I think we should at least give warning for a cycle (a quarter?), then move on at the next release (6 mos after?) at the minimum. I also don't want to overcomplicate this but I think this is good practice.

The semantic version is not a big issue, but probably safer to go up in minor (2nd digit) instead of fix (last digit).

kennethshsu:
@genedan @henrydingliu how are we moving forward with this? Who are we waiting on?

genedan:
You're waiting on me lol. Still working on getting the deprecation warning in there.

genedan:
I added some deprecation logic, via the `_resolve_pat()`  function which still allows `option` to be used but warns the user if they do it. It's more involved than I'd like, the reason why is because temporarily allowing two parameters to do the same thing leads to lots of combos that need to be handled (triggering errors when none are supplied but are required, or when both are supplied accidentally, or when both or optional, etc.).

I don't think we should discard `_resolve_pat()` after we change the parameter name for good. We could generalize it to handle deprecation/renaming of argument names.

kennethshsu:
That's probably how I would've done it too. 

And sorry for the merge conflict, can you resolve? 

@henrydingliu can you take a look at this one too and see if you agree/approve the implementation? I'd like a second set of eyes on this one just because I don't feel my review is good enough and you'll probably have more comments.

henrydingliu:
i have a more meta question before reviewing the code in more detail. did we already screw the pooch with the previous PR on this campaign? i.e. users, like our own [docs](#889), are already having to make (very slight) changes.

genedan:
I think it's revealed some weaknesses in maintaining the docs. #906 will help a lot. If the warnings in #881 work, we should see them the next time we publish the docs in the place where we set cupy:

https://chainladder-python.readthedocs.io/stable/user_guide/triangle.html#other-parameters

And then when once the deprecation happens, the notebook will fail to execute and we'll discover that we need to update that portion of the docs.

I think the other place where we set the options:

https://chainladder-python.readthedocs.io/stable/user_guide/methods.html

Doesn't use keyword arguments, so I don't think it'll be affected.

We don't have an API ref page for Options in the docs, should I go ahead and make one? The docstrings have deprecation warnings in them that are supposed to render warning boxes, so we'll be able to test whether they show up in the docs.

henrydingliu:
> I think it's revealed some weaknesses in maintaining the docs. https://github.com/casact/chainladder-python/pull/906 will help a lot. If the warnings in https://github.com/casact/chainladder-python/pull/881 work, we should see them the next time we publish the docs in the place where we set cupy:

what i'm saying is, if we have already introduced breaking changes in that PR, do we need to revert prior to the 0.9.3. or, do we just power through and allow a bit more breaking changes to make this PR simpler?

kennethshsu:
We really need to get the CEPs drafted, I think it will help us in reviewing and approving PRs. That said, I don't know many people are actually updating their packages to the latest. In fact, most have no reason to. 

The notebooks are rebuild at every release though, in fact they are rebuilt on every commit on `main`, so if something does break, the tests will fail.

> We don't have an API ref page for Options in the docs, should I go ahead and make one? The docstrings have deprecation warnings in them that are supposed to render warning boxes, so we'll be able to test whether they show up in the docs.

Yes, please do that, but I would recommend creating a new Issue and keep out of this PR.

henrydingliu:
> The notebooks are rebuild at every release though, in fact they are rebuilt on every commit on main, so if something does break, the tests will fail.

This is not entirely true. The notebooks are attempted to rebuild on every commit on main. When there is an error, the docbuild will stop at that spot. Simultaneously, depending on the severity of the error in the notebook, tests do not always fail. 

Look at this [error](https://chainladder-python.readthedocs.io/main/user_guide/triangle.html#other-parameters) on main.  It doesn't error out on doctest, only a warning. 

When it's [fixed](https://chainladder-python.readthedocs.io/experimental/user_guide/triangle.html#other-parameters), the next error (a few lines down) shows up. Then if you fixes that next error, THEN doctest will [bomb](https://github.com/casact/chainladder-python/actions/runs/26727702521/job/78765734682?pr=895) on the incomplete Frieland chapter 8. 

Anyway, my point wasn't about the notebook, but the fact that this campaign to streamline options has already introduced breaking changes. To be consistent with ourselves, we should either revert the previous PR from this campaign, or we forge ahead with pat

genedan:
That was actually a bug that became apparent rather than a breaking change. The Options class attributes were upper case, so what that docs example did was create a new lowercase attribute and assign it the value `cupy`. So you'd have 2 attributes, `ARRAY_BACKEND` being the same value as it was before, and `array_backend` being `cupy`. So the real backend never really changed.

This was a semantic bug that wasn't caught until I put the validation in there. That would also apply to the lowercase `auto_sparse`.

I think we should try putting these deprecation warnings in there, since it's a good practice and would enhance all our skills. It's not really until we try that we find ways to improve the process. If we hold off, we'll have to try again and go through the same growing pains, but on a more complex code base with more accumulated tech debt.

chainladder/__init__.py
```diff
@@ -118,32 +206,176 @@ def set_option(
...
         Returns
         -------
         None
 
         """
-
-        if option is not None:
-            self._validate_option(option)
-            setattr(self, option, copy.deepcopy(self._defaults[option]))
+        pat = _resolve_pat(pat=pat, option=option, required=False)
```
henrydingliu, line 239: is this the sole reason for the overloading on _resolve_pat?

genedan: It's the other 2 calls:

https://github.com/casact/chainladder-python/pull/877/changes/BASE..af01a1a0c003ccfa7632105360e736719ed1cd74#diff-825b6ab153366f76f74e172a1bc0f7c35b99ba01cb725bb29dc0d25bf58b8ff3R152

https://github.com/casact/chainladder-python/pull/877/changes/BASE..af01a1a0c003ccfa7632105360e736719ed1cd74#diff-825b6ab153366f76f74e172a1bc0f7c35b99ba01cb725bb29dc0d25bf58b8ff3R183

In these two cases, option/pat is required, so `_resolve_pat()` will return `str` rather than `str | None`.

henrydingliu:
gene, thanks so much for putting all of this together. 

i'm mainly reviewing this from the perspective of, how apparent will this be to future maintainer of this package (may not be this group) to know which bits to delete to deprecate `option`. there's a decent amount of complexity built into `_resolve_pat` with the overloading (which i'm not sure is necessary) and the attributeerrors (which are untested). my personal preferences aside, as long as the intention is that they can just clean delete `_resolve_pat` and everything will still work, i'm good with this PR.

genedan:
I added a note to the docstring of `_resolve_pat()`. Basically says that once `option` is removed for good, you can delete or generalize the function. I'd recommend generalizing since renaming a parameter seems likely to happen again and we can reuse the pattern the next time.

henrydingliu (approved):

henrydingliu:
great. i'm good with it.

Issue #601, opened 2025-09-25
Simplify internals to prep for 1.0 release


jbogaardt:
On the path towards 1.0 there are several refactors worth considering that would simplify the code base and bring it more into software development best practices.

# Backend support
To migrate to an arrow backend, we should minimize existing backend support to just numpy/sparse. 

cupy and dask were always elective and not well maintained. They add unnecessary complexity to the project and should be removed.

# Clarify Triangle public API
The Estimators occasionally manipulate the private internals of the Triangle class. This makes the Estimator code heavily coupled to the internals of the Triangle making refactors of Triangle internals difficult. We need to force etimators to only use the public API.  This may include adding additional support properties and methods to the Triangle class that the estimators can call on as opposed to direct manipulation of the underlying triangle structure.

None of these should change any of the exiting public API of the package and all test should continue to pass.

henrydingliu:
yo, I think 2D is the solution to all our problems. I'm so sick of all the grain issues. i'm one more grain issue away from rewriting the triangle internals to force storing at OMDM, and only regraining for estimation and display. In 2D, we can keep claim-level accident date and dates of transactions. then only map them to o and d grain at run time and aggregate for estimation or display. basically 2D enables sparse (if I'm understanding the purpose of sparse correctly) support by default.

henrydingliu:
Alas we don't have to choose between actuarial techniques or data management. Here is an excerpt from Clark 2003, which we reference in the package but is also on the exam 7 syllabus currently. According to him, Arrow is integral to more techniques. 

<img width="447" height="292" alt="Image" src="https://github.com/user-attachments/assets/d5855610-f211-4d9e-97b1-24115f539bed" />

Let me think about how we can do some performance testing next.

genedan:
I need close out a bunch of browser windows full of ideas so I'm recording them here:

Have any of you explored [JAX](https://docs.jax.dev/en/latest/notebooks/thinking_in_jax.html)? It looks like an all-in-one library with CPU, GPU, parallel processing, and sparse support, with a numpy-compatible API, which I thought could cut down on trying to maintain 4 separate backends. The main downside I see is that their sparse support is still listed as experimental. I think we can keep an eye on this one to see where it goes.

Since Polars is a dataframe and not a multidimensional array library, I suppose it's off the table as a backend option. I think there might be some performance improvements in select areas, such as I/O like when data are ingested into a DataFrame during the `Triangle` `__init__`, or when they are output as DataFrames.

[...]

Issue #842, opened 2026-05-24
Deprecate Dask


genedan:
### Description

Sub-task for #601.

### Is your feature request aligned with the scope of the package?

- [x] Yes, absolutely!
- [ ] No, but it's still worth discussing.
- [ ] N/A (this request is not a codebase enhancement).

priyam0k:
I'd like to take this one. I'll use the same approach as #843 and make the priority warning depend on where dask appears before numpy or sparse.

genedan:
Flagging these lines for review:

https://github.com/casact/chainladder-python/blob/1dd5f1d0f93343daa4a67f56ae65597d08a9f169/chainladder/core/base.py#L164-L172

priyam0k:
thanks, good catch. that constructor path was the one extra spot cupy didn't have. covered it in #937 along with set_option + set_backend.

genedan:
I created a sub-issue for the warning, which will be credited to @priyam0k. This parent should stay open just to remind us that we need to actually remove the feature at some point.

genedan:
Found another instance where dask is used:

https://github.com/casact/chainladder-python/blob/1eac4f8a349c165b486b2d59e5a3f69fcb8b1f39/chainladder/core/pandas.py#L578-L587

priyam0k:
thanks for flagging this. the bag usage is a separate thing from the array backend, so the earlier warnings didn't catch it. i added a one time deprecation warning to all three spots that use it: the groupby agg in pandas.py, incr_to_cum in triangle.py, and the arithmetic mapper in dunders.py. it's gated on the sparse backend so it only fires when dask would actually run, and it warns once per process so it doesn't spam.

opened #1008 for it (draft for now). still just warnings, no removal, so this issue can stay open to track taking it out later.

genedan:
Since the warnings went out in 0.10.0, we should be good to start removing the modules, right?

henrydingliu:
do we feel good about leaving existing dask backend users (which there could be none) to essentially live on 0.10.0 forever?

also, i don't understand semantic versioning very well. @kennethshsu can we deprecate between minor releases?

genedan:
Oh yeah, this is a breaking change, but this feature has been so unmaintained/tested, I'm not really sure how well it works with recent releases. I would be more surprised if dask/cupy actually worked than if it didn't.

henrydingliu:
[...]

PR #881, opened 2026-05-29
Deprecate cupy array backend (#843)

### LINKED ISSUE #843: Deprecate Cupy
### Description

Sub-task for #601.

- [x] Yes, absolutely!

_No response_

_No response_

priyam0k:
## Summary of Changes

Emits a `DeprecationWarning` when the cupy array backend is selected, without removing any functionality. cupy is elective and not well maintained, so it's slated for removal in a future release (#601). This warns users now so they can migrate before it's gone.

Three entry points warn:

- `set_option("ARRAY_BACKEND", "cupy")` warns
- `set_option("ARRAY_PRIORITY", ...)` warns only when cupy is ranked ahead of numpy or sparse, so valid priority lists don't get noisy
- `Triangle.set_backend("cupy")` warns once at the user's call site, even with `deep=True`. Internal recursive calls are suppressed via a private `_warn` flag.

Also documented the `_warn` and `deep` params on `set_backend`.

Added tests in `chainladder/utils/tests/test_utilities.py` covering each warning path plus the no-warn cases (non-cupy backends, and cupy ranked last in priority).

## Related GitHub Issue(s)

Closes #843.

Per @henrydingliu's call on the issue, this only deprecates (warns) and does not remove cupy. Removal can follow later under #601.

The current default priority `["dask", "sparse", "cupy", "numpy"]` has cupy ahead of numpy, so explicitly re-setting that default would warn. Happy to switch the check to only fire when cupy leads both numpy and sparse if you'd prefer the default never warns.

- `uv run --extra dev pytest -k "deprecat or cupy"` passed, 6 tests
- `uv run jb build docs --builder=custom --custom-builder=doctest` clean

### DIFF
--- a/chainladder/__init__.py
+++ b/chainladder/__init__.py
@@ -16,6 +16,7 @@ valuation defaults, as well as package metadata such as version number.
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 import copy
+import warnings
 import numpy as np
 import pandas as pd
 from importlib.metadata import version
@@ -95,6 +96,20 @@ class Options:
 
         """
         self._validate_option(option)
+        if option == "ARRAY_BACKEND" and value == "cupy":
+            warnings.warn(
+                "The 'cupy' array backend is deprecated and will be removed in a "
+                "future release. See https://github.com/casact/chainladder-python/issues/843.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+        elif option == "ARRAY_PRIORITY" and "cupy" in value:
+            warnings.warn(
+                "The 'cupy' array backend is deprecated and will be removed in a "
+                "future release. See https://github.com/casact/chainladder-python/issues/843.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
         setattr(self, option, value)
 
     def reset_option(self, option: str | None = None) -> None:
--- a/chainladder/__init__.py
+++ b/chainladder/__init__.py
@@ -104,12 +104,20 @@ class Options:
                 stacklevel=2,
             )
         elif option == "ARRAY_PRIORITY" and "cupy" in value:
-            warnings.warn(
-                "The 'cupy' array backend is deprecated and will be removed in a "
-                "future release. See https://github.com/casact/chainladder-python/issues/843.",
-                DeprecationWarning,
-                stacklevel=2,
-            )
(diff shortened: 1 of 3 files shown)


henrydingliu:
@genedan I'd like to defer this review to you.

chainladder/core/common.py
```diff
@@ -147,7 +149,7 @@ def pipe(self, func, *args, **kwargs):
         return func(self, *args, **kwargs)
 
     def set_backend(
-        self, backend: str, inplace: bool = False, deep: bool = False, **kwargs
+        self, backend: str, inplace: bool = False, deep: bool = False, _warn: bool = True, **kwargs
```
genedan, line 152: Could you add a parameter entry for _warn in the docstring?

chainladder/__init__.py
```diff
@@ -95,6 +96,20 @@ def set_option(
...
         """
         self._validate_option(option)
+        if option == "ARRAY_BACKEND" and value == "cupy":
+            warnings.warn(
+                "The 'cupy' array backend is deprecated and will be removed in a "
+                "future release. See https://github.com/casact/chainladder-python/issues/843.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+        elif option == "ARRAY_PRIORITY" and "cupy" in value:
```
genedan, line 106: Can you change this to warning the user when "cupy" occurs before "numpy" or "sparse" in the ordering? `ARRAY_PRIORITY` would expect a list of all 4 backends. So the warning should be issued based on where "cupy" is in the ordering rather than whether it occurs in the list.

priyam0k:
gentle bump on this one whenever you get a chance @genedan — it follows the three hookpoints you outlined on #843 (the two set_option paths + set_backend), warning-only per @henrydingliu's note, with a small test confirming the warnings fire. ci's green. if it looks good i'm happy to do the matching dask deprecation (#842) next so they stay consistent.

genedan:
I've added my review notes, could you address them?

priyam0k (commented):
Thanks @genedan! On my end I only see this comment, not the review notes yet. Might still be pending...

genedan (approved):
You'll find that some of the tests will fail once you get around to deprecating dask. Approving this because we can get around to adjusting the tests as #842 is addressed.

priyam0k (commented):
Thanks @genedan! Picking up #842 (deprecate Dask) next.

PR #879, opened 2026-05-29
docs: remove autosummary-generated API stubs

priyam0k:
## Summary of Changes
- Deletes the 36 `.rst` stub files under `docs/library/generated/`.
- Adds `docs/library/generated/` to `.gitignore`, mirroring the existing `docs/modules/generated/` entry one line above.

These files are emitted by `sphinx.ext.autosummary` on every docs build (`autosummary_generate = True` in `docs/conf.py`, driven by the eight `:toctree: generated/` directives in `docs/library/api.md`), so committing them is redundant and the committed copies drift out of sync with what Sphinx actually produces.

Docs-only; no runtime, config, or public API changes. Rendered HTML site is unchanged — the stubs are regenerated automatically on `jb build .`.

## Related GitHub Issue(s)
- Closes #846
- Refs parent #841
- Companion to merged #848 (which explicitly called out this follow-up: *"The companion cleanup for the redundant generated stubs is in #846."*)
- As documented in #848's `docs/README.md`, the contributor flow for adding a new API symbol is to edit `docs/library/api.md` + the package docstring — never to hand-write a `.rst` stub.

## Additional Context for Reviewers
- Verified `docs/library/generated/*.rst` is re-emitted by the build after deletion and is now ignored by git.
- The `.gitignore` addition is a single line placed directly under the existing `docs/modules/generated/` entry for consistency.
- No notebooks, conf, or `_toc.yml` references point at the deleted files; all references go through `library/api.md`'s autosummary toctrees.

- [x] I passed tests locally for both code (`uv run pytest`) and documentation changes (`uv run jb build docs --builder=custom --custom-builder=doctest`)

---

> [!NOTE]
> **Low Risk**
> Documentation and git hygiene only; API HTML is unchanged and stubs are recreated on `jb build`.
> 
> **Overview**
> Stops tracking Sphinx **autosummary** stub files under `docs/library/generated/` and adds that directory to **`.gitignore`**, alongside the existing `docs/modules/generated/` entry.
> 
> The removed `.rst` files are regenerated on every docs build from `library/api.md`; keeping them in git was redundant and let committed copies drift from what Sphinx produces. **No runtime or public API changes**—contributors still document symbols via `api.md` and docstrings.
> 
[...]

PR #886, opened 2026-05-31
feat(data): centralize sample-dataset metadata in a manifest (#774)

### LINKED ISSUE #774: Centralize sample-dataset metadata (load_sample / tests / docs / MANIFEST.in)
### Description

In #771 (review of #745), @henrydingliu observed that sample-dataset column names are duplicated across three places:

- the `load_sample` branch in `chainladder/utils/utility_functions.py`
- the test that exercises the load (e.g. `test_load_sample_clrd2025`)
- the dataset table in `docs/library/sample_data.md`

He suggested storing the metadata of available sample datasets and fields in one place, and using it to:

- run `load_sample()`
- key off tests
- generate `MANIFEST.in`
- generate `sample_data.md`

This applies to every sample dataset in `chainladder/utils/data/`, not just `clrd2025` — the same triple-duplication exists today for `clrd`, `berqsherm`, `xyz`, the friedland family, etc. The fix is general:

1. Define one manifest file (YAML or a Python dict in `chainladder/utils/data/_manifest.py`) keyed by sample name with `origin`, `development`, `index`, `columns`, `cumulative`, and any per-sample flags.
2. Refactor `load_sample` to look up its config from the manifest rather than the long `if key.lower() == ...` chain.
3. Generate the `sample_data.md` table at docs-build time (or commit the generated file with a regen script under `scripts/`).
4. Have `MANIFEST.in` include `chainladder/utils/data/*.csv` via the manifest's listed files, or just keep the existing wildcard — whichever the maintainers prefer.
5. Update tests to iterate over the manifest, so adding a new sample is a one-line change.

- [x] Yes, absolutely!

See above. Current workaround is the existing pattern — every new sample dataset (including #745's `clrd2025`) updates three files by hand.

Filed at @henrydingliu's suggestion in https://github.com/casact/chainladder-python/pull/771#issuecomment-thread. Keeping #771 / #745 scoped to landing the data; this refactor is its own piece of work.

SaguaroDev:
Closes #774.

Sample-dataset metadata was duplicated across four places: the `load_sample` if/elif chain, the tests, `docs/library/sample_data.md`, and the per-file `include` list in `MANIFEST.in`. They had already drifted: `sample_data.md` listed 23 of 46 datasets (with names like `frieldand_uspp_auto_steady_state` and `friedland_us_industy_auto_case`), and `MANIFEST.in` listed 22, so the sdist shipped incomplete.

This adds `chainladder/utils/data/_manifest.py` as the single source of truth (a plain Python dict, no new dependency per the issue discussion) and keys everything off it.

**load_sample** now looks up its `Triangle` config from the manifest instead of the if/elif chain. This is behavior-preserving: I captured the resolved `origin`/`development`/`index`/`columns`/`cumulative` for all 46 bundled samples before and after the refactor and they are byte-identical.

**cl.list_samples()** is the new utility @henrydingliu asked for. Per the design confirmed on the issue, it returns a DataFrame of name, index, columns, cumulative, and (when `include_grain=True`, the default) origin/development grain plus period counts. `include_grain=False` skips loading the data for a fast metadata-only listing. It doubles as the source for the docs table.

**MANIFEST.in** collapses the 22 hand-listed `include` lines to one `recursive-include chainladder/utils/data *.csv`. I built the sdist both ways and confirmed it now contains all 46 CSVs (was 22). Wheels were already complete via `pyproject.toml`'s `package-data` wildcard; this fixes the sdist.

**Tests** `test_load_sample` now iterates the manifest rather than globbing the data directory, so adding a sample is a one-entry change and stray non-CSV files (`__init__.py`, `_manifest.py`) can't be mistaken for datasets. Added a both-ways sync assertion (manifest set == CSVs on disk) and a `test_list_samples` test.

**docs/library/sample_data.md** is regenerated from `cl.list_samples()` via `scripts/regen_sample_data_docs.py`. The table is now complete and accurate; rerun the script after adding a dataset.

Full test suite: 717 passed, 12 xfailed (pre-existing), 0 failures.

---

> [!NOTE]
> **Low Risk**
> Refactor of sample-loading and packaging metadata with tests guarding manifest/CSV parity and sdist contents; intended behavior parity for `load_sample` with no auth or production runtime path changes.
> 
> **Overview**
> Introduces **`chainladder/utils/data/_manifest.py`** as the single registry for all bundled sample CSV metadata (`origin`, `development`, `index`, `columns`, `cumulative`). **`load_sample`** now validates keys and builds **`Triangle`** instances from that dict instead of a long per-dataset `if/elif` chain (intended to be behavior-preserving).
> 
> Adds **`cl.list_samples()`** (exported from **`chainladder.utils`**) to return a manifest-driven catalog, optionally loading each sample for grain/period counts when **`include_grain=True`**.
> 
> **`MANIFEST.in`** switches from ~22 explicit CSV includes to **`recursive-include chainladder/utils/data *.csv`**, fixing incomplete sdists; tests assert manifest ↔ on-disk CSV sync, cover **`list_samples`**, and optionally build an sdist to verify all CSVs ship.
> 
> Docs replace the static **`sample_data.md`** table with **`sample_data.ipynb`** (live **`cl.list_samples()`**) and document **`list_samples`** in the API reference.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 724b45518cfebdbbaf51e2e6ee31fd08b3ac7e03. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


henrydingliu:
@SaguaroDev thanks for all this! will the new utility function list_sample get added to the api reference section of the doc automatically? or do we need to manually add it?

SaguaroDev:
It's manual, not automatic. The API reference (`docs/library/api.md`) hand-lists each function in an autosummary block, so I added `list_samples` there plus a matching generated stub, and dropped a duplicate `load_sample` entry that was already in that block.

On the sdist: added `test_sdist_ships_all_samples`, which builds a source distribution and asserts every sample CSV is present. It self-skips when the `build` package or a source checkout isn't available, so it proves completeness in CI without weighing down the fast suite.

Pushed 577fc842.

docs/library/generated/chainladder.list_samples.rst
```diff
@@ -0,0 +1,6 @@
+﻿chainladder.list\_samples
```
henrydingliu: we've been deleting these stubs in #879. can you confirm if creating list_samples_rst is still necessary?

docs/library/sample_data.md
```diff
@@ -4,54 +4,55 @@ Below is the list of all datasets that come included with the `chainladder` pack
 
 You can load any dataset with `cl.load_sample(...)` such as `cl.load_sample("abc")`.
```
henrydingliu: actually, now that we have list_samples, would it be cleaner to just change this markdown into a notebook? i.e. just call list_samples in the notebook

docs/library/generated/chainladder.list_samples.rst
SaguaroDev: Dropped it. The committed stub is redundant anyway since autosummary regenerates it at build, and with #879 removing these on `experimental` there's no reason to add a new one here. The `list_samples` entry stays in the `api.md` autosummary block, which is what actually wires it into the API reference.

docs/library/sample_data.md
SaguaroDev: Done in 724b4551. `sample_data.md` is now `sample_data.ipynb` with a single `cl.list_samples()` cell, so the table renders live from the manifest and can't drift. I kept the `# Sample Dataset` H1 so the existing `sample_data.html#sample-dataset` anchor (linked from the tutorials) still resolves, updated the `_toc.yml` entry, and removed `scripts/regen_sample_data_docs.py` since there's nothing static left to regenerate.

henrydingliu (approved):
reviewed sample_data.ipynb on [fork](https://hdlclp.readthedocs.io/en/latest/library/sample_data.html) to validate build. confirmed that [warning from list_sample is pre-existing](https://github.com/casact/chainladder-python/actions/runs/26624780474/job/78458797598#step:5:71).

Issue #847, opened 2026-05-25
[DOCS] create custom api doc templates


henrydingliu:
the api pages appear to be generated using sphinx default templates, which contains a method summary. the summary is meant to link to individual pages for each method. however, this is overkill for our repo. a custom template will allow us to forgo the method summary, and simply list all methods on the same page. 

this should be done before we get to api docstrings at the method level, as those aren't currently shown in the method summary tables.

henrydingliu:
@priyam0k would you be interested in tackling this?

priyam0k:
On it. I added real templates under _templates/autosummary/ so methods show inline instead of as a summary table.

priyam0k:
Opened #885 for this. Added custom autosummary templates so methods/properties show inline on the class page instead of a summary table. One open call in there: `Triangle` inherits a bunch of sklearn methods (`get_params`, `set_output`, etc.) — let me know if you want those shown or hidden.

henrydingliu:
@priyam0k  i'd like to avoid scope creep. since the default template we previously use was letting all the inherited members show, we can punt on exclusions to a future enhancement. 

two edits to bring things to parity and we can move on:

-let's get rid of show inheritance on all the templates. the inheritance hierarchy has a few internal classes (like Chainladder inheriting MethodBase) that we probably don't need to show
-we should also add inherited members to the class template. i was reading Development and saw useful methods like to_pickle missing. 

@kennethshsu i'm going to change this to a bug. the original campaign was to remove doc warnings. a custom template is necessary to get rid of the missing stub file warnings we got on all the methods. will limit template enhancement to a minimum

priyam0k:
got it, makes sense. keeping inherited members and skipping the exclusions for now. i'll drop :show-inheritance: from the templates and add :inherited-members: to the plain class template so things like to_pickle show up on Development. opened #900 for it.

PR #900, opened 2026-06-01
DOCS: Bring autosummary templates to parity (#847)

priyam0k:
Follow-up to #885, addressing Henry's two parity edits from #847.

## Summary of Changes
- Drop `:show-inheritance:` from the class templates so internal base classes (e.g. `MethodBase` behind `Chainladder`) are no longer listed.
- Add `:inherited-members:` to the plain `class` template so useful inherited methods like `to_pickle` now show up on pages such as `Development`.
- Guard `linkcode_resolve` against members not accessible on the class object. Adding `:inherited-members:` surfaces members like `Pipeline.steps` (present on instances, not on the class), which made linkcode's `getattr` walk raise `AttributeError` and abort the build; those members are now skipped.

After these edits `class.rst` and `class_inherited.rst` are identical; kept as two files to stay minimal and avoid touching `api.md`. Exclusions for sklearn boilerplate are intentionally punted to a future enhancement per the discussion on #847.

## Verification
`jb build .` succeeds (189 warnings, all pre-existing). HTML spot check of `Development`: `to_pickle` now present, no "Bases:" line, 12 inline method blocks; `Pipeline` page builds without the linkcode crash.

## Related GitHub Issue(s)
- Refs #847

---

> [!NOTE]
> **Low Risk**
> Documentation and Sphinx extension behavior only; no runtime library or security impact.
> 
> **Overview**
> Brings the Sphinx **autosummary** class templates in line: both drop **`:show-inheritance:`** (so internal bases like `MethodBase` no longer clutter class pages) and enable **`:inherited-members:`** on the default `class` template so inherited API members (e.g. `to_pickle` on `Development`) appear in generated docs.
> 
> **`linkcode_resolve`** now catches **`AttributeError`** while walking dotted names, so inherited or synthetic members that cannot be resolved for GitHub source links no longer break the doc build—they simply omit the link.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7f066fff6d411aae600f5f9d4e9655a80a102f9f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/_ext/linkcode.py
+++ b/docs/_ext/linkcode.py
@@ -49,9 +49,14 @@ def linkcode_resolve(domain: str, info: dict) -> str | None:
     mod: ModuleType = importlib.import_module(info["module"])
 
     obj: ModuleType = mod
-    # Extract the part.
+    # Extract the part. Some documented members (e.g. instance attributes pulled
+    # in via :inherited-members:) are not accessible on the object itself; skip
+    # linking those rather than crashing the build.
     for part in info["fullname"].split("."):
-        obj: type = getattr(obj, part)
+        try:
+            obj: type = getattr(obj, part)
+        except AttributeError:
+            return None
 
     # Resolve properties to their underlying getter function.
     if isinstance(obj, property):
--- a/docs/_templates/autosummary/class.rst
+++ b/docs/_templates/autosummary/class.rst
@@ -5,4 +5,4 @@
 .. autoclass:: {{ objname }}
    :members:
    :undoc-members:
-   :show-inheritance:
+   :inherited-members:
--- a/docs/_templates/autosummary/class_inherited.rst
+++ b/docs/_templates/autosummary/class_inherited.rst
@@ -6,4 +6,3 @@
    :members:
    :undoc-members:
    :inherited-members:
-   :show-inheritance:


henrydingliu:
>Guard linkcode_resolve against members not accessible on the class object. Adding :inherited-members: surfaces members like Pipeline.steps (present on instances, not on the class), which made linkcode's getattr walk raise AttributeError and abort the build; those members are now skipped.

how did the old, default template get around this bug?

priyam0k (commented):
it didn't really get around it, the bug was just never triggered before. the old class template didn't use :inherited-members:, so Pipeline.steps never got documented and linkcode was never asked to resolve it. after adding :inherited-members:, that instance attribute shows up, linkcode tries to getattr it from the class, and the build fails. the guard just skips members linkcode can't resolve instead of aborting the docs build.

henrydingliu:
> it didn't really get around it, the bug was just never triggered before. the old class template didn't use :inherited-members:, so Pipeline.steps never got documented and linkcode was never asked to resolve it. after adding :inherited-members:, that instance attribute shows up, linkcode tries to getattr it from the class, and the build fails. the guard just skips members linkcode can't resolve instead of aborting the docs build.

so what did the old template use as the [scope](https://chainladder-python.readthedocs.io/main/library/generated/chainladder.Development.html) of the autosummary table?

priyam0k:
good point, my earlier wording was off. the old table did include inherited members, that's why `to_pickle` etc. show up. the difference is the old setup only listed member *names* in a summary table, which has no per-member `[source]` link, so linkcode only ran once for the class and never saw `Pipeline.steps`. the new template renders each member inline with its own `[source]` link, so linkcode finally hits `steps` (an instance attr, not on the class) and the existing bug surfaces. it was always there, inline rendering just exposed it.

henrydingliu:
is there a place where we can track al the instances of the link fail?

henrydingliu:
> is there a place where we can track al the instances of the link fail?

I take the question back. @priyam0k Are we supposed to have [failed links](https://chainladder-python.readthedocs.io/experimental/library/generated/chainladder.Pipeline.html#chainladder.Pipeline.classes_)? I thought the desired behavior of the AttributeError exception was to remove the link. But it's still giving a link, an invalid one at that.

henrydingliu:
Here is a better [link](https://chainladder-python.readthedocs.io/experimental/library/generated/chainladder.Pipeline.html#chainladder.Pipeline.set_score_request).

priyam0k:
good catch. the AttributeError guard only covered members not on the class (like `steps`). `classes_` and `set_score_request` are inherited from sklearn, so they resolve but their source is in site-packages, and linkcode built a github url as if it were in our repo. fixed it to only link members whose source is actually inside the chainladder package, otherwise no link. verified those three drop the link while our own `fit/to_pickle` still link, and the full build passes.

henrydingliu:
ah, okay. it's the injected methods from sklearn. we need to exclude them anyway. no need to fix the link

priyam0k:
got it, that makes sense. excluding them takes care of the links anyway.

Issue #928, opened 2026-06-05
UserWarning when executing cl.load_sample('quarterly')


genedan:
### Are you on the latest chainladder version?

- [x] Yes, this bug occurs on the latest version.

### Describe the bug in words

I'm working on this. Loading the quarterly sample triggers a warning, it's responsible for about 30 or so warnings in the test suite.

### How can the bug be reproduced?

```python
import chainladder as cl
cl.load_sample('quarterly')
```

```shell
UserWarning: Could not infer format, so each element will be parsed individually, falling back to `dateutil`. To ensure parsing is consistent and as-expected, please specify a format.
  datetime_mapping = dict(zip(datetime_arg, pd.to_datetime(**date_inference)))
```

### What is the expected behavior?

No warnings issued.

henrydingliu:
30 warnings means 29 instances of not using pytest fixture?

genedan:
It was still used, just initialized inefficiently. The Pandas warning was saying that the dates were being inferred element-by-element rather than all at once due to a lack of a format string. The fix should offer a performance boost if anyone's trying to initialize a really big quarterly triangle.

Issue #840, opened 2026-05-24
[TEST] Reduce pytest warnings


genedan:
### Are you on the latest chainladder version?

- [x] Yes, this bug occurs on the latest version.

### Describe the bug in words

It's looking like our push towards 1.0 is turning into a giant tech debt reduction exercise. I might as well add getting the number of pytest warnings down.

This issue is an open call to all contributors to have at it and see what they can knock down, and governs the overall subject of pytest warnings for the 1.0 release. If you do identify something you can fix, please create a sub-issue and ask to be assigned to it. That way you can receive credit for closing an issue via a PR.

### How can the bug be reproduced?

Run `uv run pytest`. 

```shell
=============================================================================================================================== test session starts ================================================================================================================================
platform linux -- Python 3.14.4, pytest-9.0.3, pluggy-1.6.0
rootdir: /home/ubuntu/Repos/chainladder-python
configfile: pyproject.toml
testpaths: chainladder, docs/_ext
plugins: nbmake-1.5.5, cov-7.0.0, xdist-3.8.0
collected 723 items                                                                                                                                                                                                                                                                

chainladder/adjustments/tests/test_berqsherm.py ..                                                                                                                                                                                                                           [  0%]
chainladder/adjustments/tests/test_bootstrap.py ...                                                                                                                                                                                                                          [  0%]
chainladder/adjustments/tests/test_parallelogram.py .......                                                                                                                                                                                                                  [  1%]
chainladder/adjustments/tests/test_trend.py ....                                                                                                                                                                                                                             [  2%]
chainladder/core/tests/test_arithmetic.py ....................xx....                                                                                                                                                                                                         [  5%]
chainladder/core/tests/test_correlation.py .....                                                                                                                                                                                                                             [  6%]
chainladder/core/tests/test_display.py ...........................                                                                                                                                                                                                           [ 10%]
chainladder/core/tests/test_grain.py ..............................................................................................................                                                                                                                          [ 25%]
chainladder/core/tests/test_slicing.py ..............................x..                                                                                                                                                                                                     [ 30%]
chainladder/core/tests/test_triangle.py ....................................................................................................xxxxxxxx...........................................                                                                              [ 50%]
chainladder/development/tests/test_barnzehn.py ....                                                                                                                                                                                                                          [ 51%]
chainladder/development/tests/test_constant.py ........                                                                                                                                                                                                                      [ 52%]
[...]

PR #929, opened 2026-06-05
#928 quarterly warning

genedan:
## Summary of Changes  

Handles quarterly edge case when conducting datetime inference. Python doesn't have a native datetime format string for quarterly dates, so I decided to intercept the UserWarning and handle the quarterly case separately.

Reduces the number of warnings from 1112 to 1080.

## Related GitHub Issue(s)  

#840
#928 

---

> [!NOTE]
> **Medium Risk**
> Changes core Triangle date parsing used at construction; quarterly fallback could affect edge-case origin/development labels, though behavior for known formats should stay the same.
> 
> **Overview**
> **`TriangleBase._to_datetime`** now treats **`UserWarning`** from **`pd.to_datetime`** as a failed inference attempt (via **`warnings.filterwarnings('error')`**) instead of letting quarterly strings warn and still partially parse. When the usual format list fails, it falls back to **`pd.PeriodIndex(..., freq='Q').to_timestamp`**, honoring **`period_end`** through a shared **`start_end`** flag for period timestamps.
> 
> A regression test asserts **`Triangle`** still raises **`ValueError`** with *"Unable to infer datetime"* when origin/development strings use incompatible formats (e.g. **`1995/Q1`** vs **`1995Q1`**). **`Triangle`** docs only clarify the **`data`** type as **`DataFrame | DataFrameXchg | dict`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 99682fe9c93db90aa3587d9363e00118250a7e30. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/chainladder/core/base.py
+++ b/chainladder/core/base.py
@@ -409,17 +409,17 @@ class TriangleBase(
                     try:
                         datetime_values = cast(DatetimeIndex, pd.to_datetime(**date_inference))
                         break
-                    except ValueError:
+                    except (ValueError, UserWarning):
                         pass
-                    # Quarterly edge case.
-                    except UserWarning:
-                        datetime_values = pd.PeriodIndex(datetime_arg, freq='Q').to_timestamp(how=start_end)
-                        break
 
             if datetime_values is None:
-                raise ValueError(
-                    "Unable to infer datetime for field(s): " + str(fields) +
-                    ". Please check the underlying data or any supplied format arguments."
+                # Quarterly edge case.
+                try:
+                    datetime_values: DatetimeIndex = pd.PeriodIndex(datetime_arg, freq='Q').to_timestamp(how=start_end)
+                except ValueError:
+                    raise ValueError(
+                        "Unable to infer datetime for field(s): " + str(fields) +
+                        ". Please check the underlying data or any supplied format arguments."
                     )
             datetime_mapping = dict(zip(datetime_arg, datetime_values))
             target: Series = target_field.map(arg=datetime_mapping)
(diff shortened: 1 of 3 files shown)


henrydingliu:
I had the entirely wrong idea of what this PR would look like. I was honestly thinking you would just change the underlying data. 

- switching from a warning to an error should probably take a deprecation cycle?
- i'm not too hot about adding these edge cases. sure, it's one of the sample datasets in the package. but [elsewhere](https://github.com/casact/chainladder-python/blob/main/chainladder/utils/data/friedland_auto_freq_sev.csv) in the repo, we have non-annual grains dataset working perfectly fine. What if someone now wants to open up an issue to add an edge case for H, or another date format? I think my response to any of those issues would be, "wrangle in pandas", which I believe is also sklearn's position.

genedan:
I can try that. It would be something like replacing data like `1995-Q2` with  `1995-04` to fix? I'll open up another PR if I can get it to work.

Is there anywhere in the docs that has best practice for the date formats? The docs are quite hard to search. If absent, we could add that the best way to format a quarterly triangle would be to have the month beginning of the quarter rather than using a Q, which could be inferred but will cause performance issues.

genedan:
I think the term I was looking for was [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601), we could update the docs to say the best practice would be to encode the dates as ISO 8601 compatible.

henrydingliu:
> Is there anywhere in the docs that has best practice for the date formats? The docs are quite hard to search. If absent, we could add that the best way to format a quarterly triangle would be to have the month beginning of the quarter rather than using a Q, which could be inferred but will cause performance issues.

There is [this](https://chainladder-python.readthedocs.io/stable/user_guide/triangle.html#date-inference). We can definitely beef it up.

Issue #486, opened 2024-01-18
Add type annotations where useful (constructors, methods, functions, etc.)


MatthewCaseres:
**Is your feature request related to a problem? Please describe.**
When typing out arguments I want to see what I should be passing in.

**Is your feature request at odds with the scope of the package?**
dont think so

**Describe the solution you'd like**
Put type annotations so that there are minimal "Any" in this screenshot

<img width="694" alt="Screenshot 2024-01-17 at 7 59 25 PM" src="https://github.com/casact/chainladder-python/assets/43053796/4f7c9c4f-4ef0-44ac-b430-c1b06f915d7a">

**Describe alternatives you've considered**
none

**Additional context**

Currently there is no minimal version of Python associated with this project, see setup.py `install_requires`. Type annotations are from 3.5, so it might be worthwhile to add this as requirement.

Ideally the type annotation for origin_grain is `Literal['Y', 'S', 'Q', 'M']`, and typing.Literal is from 3.8, and 3.8 is also the earliest version that is not end of life - https://devguide.python.org/versions/. So I want to add an earliest Python version to `install_requires` and make it 3.8.

genedan:
I'll take this one on.

genedan:
By the way, I found an interesting library called [Pyright](https://microsoft.github.io/pyright/#/) that should help out greatly with this. I went through [a tutorial](https://betterstack.com/community/guides/scaling-python/pyright-explained/) and it can do some cool things, such as type coverage (analogous to test coverage) and can check whether certain conditional statements will never be encountered due to the way the code is written, as well as other bug-detection features.

I also learned some things that we can use in chainladder, such as the `Literal` keyword for restricting inputs to a finite list (such as accepted types of grain or triangle dimensions). And also `Final` for certain constants that should never be changed (maybe certain time constants like year length?).

It's actively developed by Microsoft so it looks to be well-maintained. And since it doesn't impact any of the features, it can also be easily removed as a dependency if we ever need to.

Issue #939, opened 2026-06-07
Add type completeness report workflow


genedan:
### Description

Running `pyright --verifytypes chainladder` creates a type completeness coverage report:

```shell
...
Symbols exported by "chainladder": 1101
  With known type: 95
  With ambiguous type: 89
  With unknown type: 917

Other symbols referenced but not exported by "chainladder": 3156
  With known type: 2159
  With ambiguous type: 85
  With unknown type: 912

Symbols without documentation:
  Functions without docstring: 930
  Functions without default param: 53
  Classes without docstring: 207

Type completeness score: 8.6%

Completed in 4.375sec
```

The percentage % is steadily increasing. I think we should add a workflow to report it in the PRs. I don't think think we should add any patch or project minimums yet since it would be unfair to contributors to expect to know what types to use if most of the codebase is still untyped. So This is mostly just for informational purposes at the moment.

### Is your feature request aligned with the scope of the package?

- [x] Yes, absolutely!
- [ ] No, but it's still worth discussing.
- [ ] N/A (this request is not a codebase enhancement).

henrydingliu:
@genedan can close?

PR #948, opened 2026-06-10
Triangle slicer tests

genedan:
## Summary of Changes  

Adds tests for missing lines in `TriangleSlicer` class, and adds annotations.

## Related GitHub Issue(s)  

#486 (partial)
#736 (partial)

The pattern of how the `Triangle` class is composed of a bunch of sibling classes (TriangleSlicer, Common, etc.)  is called a [mixin](https://en.wikipedia.org/wiki/Mixin), i.e., `TriangleSlicer` is a mixin class. I've annotated this class using the practices found in this [Stack Overflow thread](https://stackoverflow.com/questions/51930339/how-do-i-correctly-add-type-hints-to-mixin-classes).

A new class called `TriangleProtocol` is defined in `typing.py` which includes missing attributes from `TriangleSlicer` that it needs to act upon (such as `vdims`). The reason why this was created is due to numerous unresolved attribute warnings, since attributes like `vidims` are defined in `Triangle` and not `TriangleSlicer`. This pattern is intended to be rolled out to the rest of the mixins.

Some type fixes are applied to other files so I could clean up `TriangleSlicer`. I tried to keep those to a minimum otherwise.

---

> [!NOTE]
> **Low Risk**
> Changes are mostly annotations and tests; runtime edits are localized to column assignment and exception handling with existing behavior preserved and covered by new tests.
> 
> **Overview**
> Introduces **`BackendArray`** and **`TriangleProtocol`** in `typing.py` so mixin methods on `TriangleSlicer` can be typed without duplicating `Triangle` attributes, and applies **`cast`** for `Triangle.values` in `triangle.py` and aggregation paths in `pandas.py`.
> 
> **`TriangleSlicer`** gains full signatures and docstrings on `__getitem__`, `__setitem__`, `_slice`, `_slice_valuation`, and `_set_slicers`, with `isinstance` checks replacing `type(...)`, valuation masks reshaped with **`order="F"`**, misaligned new-column assignment catching **`ValueError`/`AttributeError`** instead of a bare `except`, and scalar values broadcast when appending a column.
> 
> **`test_slicing.py`** adds tests for `.loc` assignment, virtual columns, cross-backend column replacement, raw array assignment, and misaligned origin grids when adding a column.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d585a32bc52a9016f7a3a356326de792bf03eb11. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chainladder/core/typing.py
```diff
@@ -23,4 +42,29 @@
...
+    """
+    @property
+    def shape(self) -> tuple[int, int, int, int]: ...
+
+    @property
+    def index(self) -> pd.DataFrame: ...
+
+    key_labels: list[str]
+    values: BackendArray
+    vdims: np.ndarray
```
henrydingliu: i think vdims is meant to be private

chainladder/core/typing.py
henrydingliu, line 1: this is excluded from test coverage, right? in general, what's your mental model for ensuring that everything in here correct?

chainladder/core/pandas.py
```diff
@@ -548,7 +551,7 @@ def agg_func(
         obj._set_slicers()
         if auto_sparse:
             obj = obj._auto_sparse()
-        obj.values: ArrayLike = num_to_nan(obj.values)
+        obj.values = cast("BackendArray", num_to_nan(obj.values))
```
henrydingliu, line 564: with stuff like `BackendArray`, how do we make sure that each underlying type gets tested? or maybe the better question is, do we need to care?

chainladder/core/typing.py
```diff
@@ -23,4 +42,29 @@
 # before index_key/other_key resolve it to a positional _AxisKey.
 _LabelKey: TypeAlias = IndexExpression | str | pd.Series | pd.DataFrame
 
+class TriangleProtocol(Protocol):
```
henrydingliu, line 45: what's the difference between triangleprotocol and trianglelike?

henrydingliu:
we are getting pretty sophisticated over here. while i have enormous confidence in the care and consideration you are putting into this, i'm also concerned that we are entering into a new orbit of abstractions that makes it harder for new collabortors to pick up on the internal architecture of the package. 

so is the introduction of `TriangleProtocol` a good milestone to also lay down some documentation of the inheritance structure? maybe [here](https://github.com/casact/chainladder-python/tree/main/chainladder/core#core)?

genedan:
I considered a few alternatives:

https://github.com/casact/chainladder-python/discussions/574

Leaving the classes as-is (mixin) and then typing them via a protocol was the least disruptive option and adheres to accepted practice. The last post in that thread (annotating each class with the missing attributes as stubs) was also something I tried, but it can lead to repetition and inconsistencies across the base classes.

2 other options I considered - abstract classes (using decorators) and cooperative multiple inheritance (using `super().__init__()`)  would require significant redesign of the Triangle class, so I opted not to pursue them.

There's a thing called [`ARCHITECTURE.md`](https://architecture.md/) which could be a good way to document what you'd suggested.

As far as the barrier-to-entry, I think the biggest hurdle is a combination of (1) moving from functions to classes and (2) moving from scripts to packages. I think once you've gotten that far, typing is definitely learnable, I think it's easier than the other 2 steps I mentioned since actuaries need to work with types in Excel and VBA. We should maintain:

- Several "Great First Contribution" issues
- A very warm welcome to invite the general public to contribute (I think the package README.md does an excellent job).
- Contributing tutorials that cover the package architecture, and a guide on how to turn the base estimator classes into a custom reserving method.

So far the number of active contributors has been higher than ever, so I'm feeling good about the community at the moment.

chainladder/core/tests/test_slicing.py
```diff
@@ -188,6 +188,31 @@ def test_at_setitem_triangle_value(raa: Triangle) -> None:
...
+    sub = tri.loc["Aegis Grp", "comauto"].copy()
+    if tri.array_backend == "sparse":
+        with pytest.raises(ValueError, match="Setting values with sparse backend requires .at or .iat"):
+            tri.loc["Aegis Grp", "comauto"] = sub * 2
+    else:
+        tri.loc["Aegis Grp", "comauto"] = sub * 2
+        assert tri.loc["Aegis Grp", "comauto"] == sub * 2
+
+
+@pytest.mark.xfail
```
henrydingliu: we upgraded the xfail in an earlier PR

henrydingliu:
>There's a thing called [ARCHITECTURE.md](https://architecture.md/) which could be a good way to document what you'd suggested.

that'd definitely work. 

@genedan in case you did leave some comment, i can't see any

chainladder/core/tests/test_slicing.py
genedan: The decorator must have gotten caught up in the merge. I've removed it.

genedan:
k, I saw that and responded, I see it visible in incognito too. Maybe last time I accidentally started a review, I'll keep an eye out for whether a comment says "pending" as a precaution.

kennethshsu:
Pending is basically staged, you need to submit to commit lol

henrydingliu:
> k, I saw that and responded, I see it visible in incognito too. Maybe last time I accidentally started a review, I'll keep an eye out for whether a comment says "pending" as a precaution.

i have maybe another 4 comments from when i first reviewed this a week ago. 

@kennethshsu can you see them?

kennethshsu:
I only see 1?

kennethshsu:
[This](https://github.com/casact/chainladder-python/pull/948#discussion_r3424519641) one.

genedan:
Ah, I see them now, let me respond.

chainladder/core/typing.py
genedan: They're public, although the methods used to define them are private. Should we convert them to private in 1.0? :

image

Or - John wants them out completely for 1.0: https://github.com/casact/chainladder-python/issues/601

How about this - I'll remove the 'dims stub with the anticipation that we'll be making either of these two changes in the near future. There will be residual warnings from the IDEs, but we'll know they'll be gone soon.

chainladder/core/typing.py
genedan: `TriangleLike` is just an alias for `Triangle`. I saw `TriangleLike` as prior work in one of the docstrings, so I made an alias for it since it's not actually defined as a class. `TriangleProtocol` contains the attribute and method stubs that are expected for the Triangle mixins.

If it's okay with you, I'd like to remove TriangleLike for now. It seemed like something that was planned, but not implemented, so I'd rather convert that into an issue/discussion instead of leaving it in the code. But I'm also ok with keeping it too.

chainladder/core/typing.py
genedan: Two ways:

1. Pyright workflow: https://github.com/casact/chainladder-python/issues/939, it checks the consistency of the types throughout the package. Takes just a few seconds, so it's a low cost. Right now it spits out an overwhelming number of warnings about objects not being typed, which makes the report hard to parse. I'd like to get the coverage % higher before we get stricter on consistency. We'll eventually get there as we chip away at it.

2. We can add 1 unit test for the TriangleProtocol. It will check to makes sure the `Triangle` class and `TriangleProtocol` APIs the same. That way we can detect breaking changes that would warrant a major release. We could eventually do something similar with estimators. I'd like to get closer to completion before writing a test though, since the signatures won't be in sync until then.

chainladder/core/pandas.py
genedan: I'd written the wrong comment here, although the response involves using the same tool - Pyright's type checker.

There's also several other tools like mypy, pyrefly, and I think I saw some blog posts saying you should run all of them although that seems like overkill at the moment. So far I've been satisfied with Pyright, maybe we can switch if we find something wrong with it.

henrydingliu:
Merging the comments across several threads.

henrydingliu:
sorry, wrong button

chainladder/core/typing.py
henrydingliu: >If it's okay with you, I'd like to remove TriangleLike for now.

no issue with removing TriangleLike since it's a placeholder anyway. so should we be go back to using triangle as the hinted type?

chainladder/core/typing.py
henrydingliu: there's gotta be a way for AI to get us caught up quickly

chainladder/core/typing.py
henrydingliu: if vdim is a one-of, i.e. you won't be adding the rest of the dims, i don't mind it as the one spot we need to come back and address later

genedan: I think if you hadn't mentioned it, I would have added a the rest. I opted to remove as that would be more in line with the goals in #601.

chainladder/core/typing.py
genedan: I've been using it, and it helps a lot. What I do is I ask the LLM to comb through the unit tests and find all the types that get passed to function arguments, that gives me a starting point, and then I go through the lines to check and refine.

It discovers bugs and inconsistencies that have to be fixed though, and also places where the code has to be refactored, this is one of the scenarios that can slow things down a bit. Another issue is sometimes Claude will suggest inconsistent type hints across functions that should take the same types, and Cursor bugbot also catches errors made by Claude. So in my experience, it wasn't a one-shot prompt, but overall it's been a net positive for me.

On the faster side, I can type hint a module in about a day, before that, it took me several days.

chainladder/core/typing.py
genedan: Yeah, we'd just replace it with Triangle. I asked Claude to comb through the repo history to see who came up with the idea of `TriangleLike`, it says it was me:

```
The first introduction of TriangleLike was commit f926c140 by you (Gene Dan) on June 7, 2026 at 16:47 CST — creating chainladder/core/typing.py with the TriangleLike: TypeAlias = "Triangle" definition. The second commit (bb523d08, ~6 hours later) by Henry Liu was
  the first consumer of it, importing and using it as a type annotation in triangle_weight.py. No variations (triangle-like, trianglelike, etc.) appear anywhere in the history.
```

But I still don't think it was my idea. There's no way this is true because that was a little more than a week ago. I'll create a discussion to give a chance for people to chime in. I'd leave the actual deletion for a separate PR.

chainladder/core/typing.py
henrydingliu: copilot gave me another idea. i'll approve this and raise in a discussion

henrydingliu (approved):

Issue #704, opened 2026-04-21
API Reference Examples


genedan:
**Latest Status**:

|Domain | Module | Status |
| ----------|------------ | ------------ |
Core Data Structures | Triangle | Done ✅
Core Data Structures | minimum | Done ✅
Core Data Structures | maximum | Done ✅
Core Data Structures | base.py | Done ✅
Core Data Structures | common.py | Done ✅
Core Data Structures | display.py | Done ✅
Core Data Structures | dunders.py | Done ✅
Core Data Structures | io.py | Done ✅
Core Data Structures | pandas.py | Done ✅
Core Data Structures | slice.py | Done ✅
Deterministic Reserving Models | Chainladder | Done ✅
Deterministic Reserving Models | BornhuetterFerguson | Done ✅
Deterministic Reserving Models | Benktander | Done ✅
Deterministic Reserving Models | CapeCod | Done ✅
Deterministic Reserving Models | ExpectedLoss | Done ✅
Development Pattern Estimators | Development | Done ✅
Development Pattern Estimators | DevelopmentCorrelation | Done ✅
Development Pattern Estimators | ValuationCorrelation | Done ✅
Development Pattern Estimators | DevelopmentConstant | Done ✅
Development Pattern Estimators | MunichAdjustment | Done ✅
Development Pattern Estimators | IncrementalAdditive | Done ✅
Development Pattern Estimators | ClarkLDF | Done ✅
Development Pattern Estimators | CaseOutstanding | Done ✅
Development Pattern Estimators | TweedieGLM | Done ✅
Development Pattern Estimators | DevelopmentML | Done ✅
Development Pattern Estimators | BarnettZehnwirth | Done ✅
Operational / Trend / Exposure Tools | BerquistSherman | Done ✅
Operational / Trend / Exposure Tools | Trend | Done ✅
Operational / Trend / Exposure Tools | ParallelogramOLF | Done ✅
Stochastic / Simulation | MackChainladder | Done ✅
Stochastic / Simulation | BootstrapODPSample | Done ✅
Tail Factor Models | TailConstant | Done ✅
Tail Factor Models | TailCurve | Done ✅
Tail Factor Models | TailBondy | Done ✅
Tail Factor Models | TailClark | Done ✅
Utility Functions | load_sample | Done ✅
Utility Functions | read_pickle | Done ✅
Utility Functions | read_json | Done ✅
Utility Functions | concat | Done ✅
Utility Functions | minimum | Done ✅
[...]

PR #1035, opened 2026-06-21
Pandas annotations

genedan:
## Summary of Changes  

Add type annotations to `pandas.py`.  Add API reference examples to `TrianglePandas.dropna()`. This method really needed examples to illustrate its complex behavior. I'll try to stick to just 1 or 2 methods next time if they turn out to be as complex.

## Related GitHub Issue(s) 

#486 
#704  

Removed line 303:

```
self.values = np.where((xp.nan_to_num(self.values) == 0) * (self.nan_triangle == 1), self.nan_triangle * 0, self.values)
```

This line didn't do anything, it simplifies to self.values = self.values, so I removed it.

---

> [!NOTE]
> **Low Risk**
> Changes are mostly annotations and documentation; fillna(None) now errors and fillna inplace is refactored but intended to be equivalent—worth a quick regression check on NaN filling.
> 
> **Overview**
> Adds **`TriangleProtocol`** coverage and **`TrianglePandas`** type hints (`self: TriangleProtocol`, return types, axis literals) so static checkers understand mixin methods on `Triangle`, with matching **`__getitem__` overloads** in `slice.py` and an expanded protocol in `typing.py`.
> 
> **`dropna()`** gains doctest-style **API examples** and inline comments; behavior is unchanged aside from typing/casts.
> 
> **`fillna()`** now **rejects `value=None`** with `TypeError`, and the **inplace** path no longer builds an intermediate `self + value * 0` frame—it applies the same mask logic directly on `self.values`.
> 
> **`fillzero()`** drops a redundant final `np.where` branch that was a no-op (`self.values = self.values`).
> 
> Minor **`to_frame()`** cleanup: sparse `COO` cast, valuation unstack via a named Series, and removal of invalid `out[col]: Series` assignment syntax.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8fc49bff76961101fc3c089bdf0d7c4335587f3f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


genedan:
I'll fix that codecov comment in another PR, I think this one's already kind of big.

chainladder/core/pandas.py
```diff
@@ -236,36 +280,256 @@ def _get_axis(axis: str | int| None) -> int:
...
         existing line in the same triangle.
+
+        Returns
+        -------
+        Triangle
+
+        Examples
+        --------
+
+        In a single-dimension case, an origin period will be dropped if it contains all NaN.
```
henrydingliu, line 301: does this happen in the interior of the triangle?

chainladder/core/pandas.py
```diff
@@ -236,36 +280,256 @@ def _get_axis(axis: str | int| None) -> int:
...
+
+        .. testoutput::
+
+                     24     36
+            1985  500.0  600.0
+            1986  600.0    NaN
+
+        If both the earliest origin and development periods are all NaN, both will be dropped.
+
+        .. testcode::
```
henrydingliu, line 376: dropna isn't called in this example

chainladder/core/pandas.py
henrydingliu: i just read the later example. looks like it wouldn't work in the interior. we should clarify earlier in the docstring that dropna doesn't work on the interior

chainladder/core/pandas.py
```diff
@@ -236,36 +280,256 @@ def _get_axis(axis: str | int| None) -> int:
...
+                    'lob': ['abc'] * 6 + ['xyz'] * 6,
+                    'paid': [np.nan, np.nan, np.nan, 500, 600, 500] * 2,
+                },
+                origin='origin',
+                development='development',
+                index='lob',
+                columns=['paid'],
+                cumulative=True
+            )
+            print(tri.loc['abc'])
```
henrydingliu, line 495: dropna isn't used.

chainladder/core/pandas.py
```diff
@@ -236,36 +280,256 @@ def _get_axis(axis: str | int| None) -> int:
...
+            print(tri.dropna())
+
+        .. testoutput::
+
+                     12     24
+            1985  500.0  600.0
+            1986  500.0  600.0
+            1987  500.0    NaN
+
+        In the case of a multi-dimensional Triangle, periods will only be dropped if their aggregate sum across
```
henrydingliu, line 477: does this mean if i have a ground-up triangle and a deductible triangle in the same triangle, i could end up with origins or developments dropped?

chainladder/core/pandas.py
```diff
@@ -236,36 +280,256 @@ def _get_axis(axis: str | int| None) -> int:
...
             ddim = list(
                 (xp.nansum(obj.values[0, 0, :], -2) != 0).astype("int"))
             ddim = obj.development[pd.Series(ddim).astype(bool)]
+            # Slice the Triangle by the development periods that have data.
             obj = self[
                 (self.development >= ddim.min()) & (
                     self.development <= ddim.max())
             ]
-            return obj[(self.origin >= min_odim) & (self.origin <= max_odim)]
+            obj = cast("TriangleProtocol", cast(object, obj))
```
henrydingliu, line 535: two casts (i guess technically 3) in 2 lines. why the differences?

- one upcasts to object but ones doesn't
- one casts to triangleprotocal one casts to triangle

chainladder/core/pandas.py
```diff
@@ -436,7 +709,7 @@ def astype(self, dtype, inplace=True):
         obj.values = obj.values.astype(dtype)
         return obj
 
-    def head(self, n=5):
+    def head(self: TriangleProtocol, n=5):
```
henrydingliu: is it too trivial to add n:int?

chainladder/core/pandas.py
genedan: Could you give a small example to illustrate this scenario? I wasn't sure if you meant losses above or below the deductible, and signs, etc. If you were to add up all the sub-triangles and wound up with a period of all NaNs, it'd get dropped.

chainladder/core/pandas.py
genedan: It is a couple blocks down, this initial setup shows the location of the NaNs.

chainladder/core/pandas.py
genedan: The double-cast pattern is used to eliminate a certain IDE warning. If I were to do a single cast:

`cast("TriangleProtocol", obj)`

I'd get this warning from my IDE: 

`Cast of type 'Triangle | Series' to type 'TriangleProtocol' may be a mistake because they are not in the same inheritance hierarchy. If this was intentional, cast the expression to 'object' first.`

The next cast doesn't throw the warning since we're casting to Triangle, which is one of `Triangle | Series`. I wonder if casting is a crutch to cover up for some weaknesses in the code. I have a hunch that if we had clearer branching in the functions, the return type would be unambiguous and the type checker would know which one of `Triangle | Series` it'd be. But that's just a hunch, I guess I've gotten used to dynamic typing for far too long...

chainladder/core/pandas.py
genedan: Added.

chainladder/core/pandas.py
genedan: Added - this example uses multiple blocks, you'll find it further down.

chainladder/core/pandas.py
genedan: Added.

chainladder/core/pandas.py
henrydingliu: hmmm, where is your IDE getting `Triangle|Series` from? self is typed to `TriangleProtocol` at the top. 

and when should we type to `Triangle` and when should we type to `TriangleProtocol`?

chainladder/core/pandas.py
henrydingliu:
```python
tri = cl.Triangle(
    data={
        'origin': [1985, 1985, 1985, 1986, 1986, 1987],
        'development': [1985, 1986, 1987, 1986, 1987, 1987],
        'gross_paid': [500, 600, np.nan, 500, 600, 500],
        'ceded_paid': [-500, -500, np.nan, -500, -500, -500],
    },
    origin='origin',
    development='development',
    columns=['gross_paid','ceded_paid'],
    cumulative=True
)
print(tri)
tri.dropna()
print(tri)
```
it does drop. bit of a corner case but annoying nonetheless.

chainladder/core/pandas.py
genedan: It's getting it from `TriangleSlicer.__getitem__()` since the preceding line executes a slice:

```
obj = self[
                (self.development >= ddim.min()) & (
                    self.development <= ddim.max())
            ]
```

I've typed `Self: TriangleProtocol` when I need the attributes that aren't defined in the class. For example, in this case we needed the slicer that's defined in `TriangleSlicer`, so I used the protocol. In methods where we're only using names defined in the class, we can use `self: Triangle`.

I also added overloads in `typing.py`. This explicitly states some of the return paths certain combos of arguments lead to, which allowed me to discard some of the casts.

henrydingliu: > I've typed Self: TriangleProtocol when I need the attributes that aren't defined in the class. 

so what's the end goal? will triangleprotocol be a mixin for triangle? like, will we permanently have functions with self: something thats not a triangle but returns a triangle? i think i'd feel better about this path we are taking if we have a model package that we are following, or a sketch in architecture.md

genedan: Since `Triangle` inherits from all the various mixin classes, I don't think we'll need to use `TriangleProtocol` once we get around to typing the `Triangle` class.

The model I'm using is the pattern in this StackOverflow thread (answer 1):

https://stackoverflow.com/questions/51930339/how-do-i-correctly-add-type-hints-to-mixin-classes

Answer 2 is also an option, where the mixin inherits the protocol instead. If we did that, we'd have `class TrianglePandas(TriangleProtocol)` with no need for `self: TriangleProtocol` for all the methods, it could lead to cleaner syntax.

henrydingliu (approved):

chainladder/core/pandas.py
genedan: I came across this snippet from Pandas that gives us some guidance on what to do. Indeed, we should avoid casts to the best of our ability:

image

We're kind of in this awkward situation where type checking tools haven't caught up with Python. So I don't think we can avoid casting perfectly, we just have to make a best effort to avoid it.

henrydingliu: should we put our typing efforts on pause? do we need to go back and refactor a bunch of the existing `cast` we've added to the codebase over the recent month?

genedan: I think we should keep going. A lot of the casting issues can also be caused by the low coverage we have (hard for the tools to infer types when so much of the library is uncovered), and we'll have a better idea of possible refactor strategies when the whole library is covered, rather than little portions of it.

henrydingliu: i would propose this. let's create a branch that has a single test that's just all-period volume-weighted development transformer then chainladder predictor on raa. then we type every line that's covered by that test. basically the experimentation we are taking to cover typing by .py file is not agile enough for us to quickly discover what doesn't work. if we can find a consistent typing approach for every line covered by a single raa test, that approach should work for the entire package.

genedan: I would recommend that approach for typing up the estimators. I actually do this for the `core` modules. When I type hint one of them, instead of going through the file from top-to-bottom, I first ask Claude to create a reprex that his the majority of the lines in the module in question.

Some things that slow me down are finding bugs during that process that need to be fixed. I also combine the tasks of filling out the docstring, doing code comments and linting as well.

PR #1038, opened 2026-06-21
TST: Add arithmetic compatibility tests.

genedan:
## Summary of Changes  

Add some tests for uncovered lines in `dunders.py`, dealing with arithmetic operations on Triangles with different grain and backends.

---

> [!NOTE]
> **Low Risk**
> Test-only changes with no production code modifications.
> 
> **Overview**
> Adds regression tests in `test_arithmetic.py` for `dunders.py` arithmetic behavior that was previously untested.
> 
> **Backend promotion:** sparse `Triangle` + `numpy.ndarray`, and numpy `Triangle` + sparse `COO`, should both yield a sparse backend and match `raa + 1` when converted to numpy.
> 
> **Grain guard:** adding triangles with mismatched grain (e.g. `raa` vs `qtr`) must raise `ValueError` with the message that both triangles need the same grain.
> 
> Existing arithmetic tests are unchanged aside from new imports and typing (`TYPE_CHECKING`, `Triangle` fixture annotations).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b024736bb4641a7a8e0f21048f91b6f110bf9c48. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/chainladder/core/tests/test_arithmetic.py
+++ b/chainladder/core/tests/test_arithmetic.py
@@ -1,4 +1,78 @@
+from __future__ import annotations
+import numpy as np
 import pytest
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from chainladder import Triangle
+
+from chainladder.utils.sparse import COO
+
+
+def test_arithmetic_ndarray_other_backend(raa: Triangle) -> None:
+    """
+    When the left side of an arithmetic operator has a sparse backend, and the right side is a numpy ndarray,
+    The result should have a sparse backend.
+
+    Parameters
+    ----------
+    raa: Triangle
+        The raa sample data set Triangle.
+
+    Returns
+    -------
+    None
+    """
+    sparse_raa = raa.set_backend("sparse")
+    other = np.ones(sparse_raa.shape)
+    result = sparse_raa + other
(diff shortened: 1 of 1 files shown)


henrydingliu:
ah, splendid. i've had an issue with sparse arithmetic forever. 

the behavior below seems counter-intuitive when adding any two sparse triangles return a sparse triangle. would you consider this to already covered by these new test? 

```python
clrd = cl.load_sample('clrd').set_backend('sparse')
clrd_gr = clrd.groupby('GRNAME').sum()
print(clrd.array_backend)
print(clrd_gr.array_backend)
```
```python
sparse
numpy
```

genedan:
Not for that example. I tried running it through the debugger and it doesn't hit those lines in `dunders.py`. Instead, the backend changes because the example triggers `TriangleBase._auto_sprase()`, which determines that the grouped-by triangle is too small to be sparse, so it changes the backend to numpy.

https://github.com/casact/chainladder-python/blob/468f11a4267e90308a5261e24a27a280fe8f53aa/chainladder/core/base.py#L546

henrydingliu (approved):

PR #1088, opened 2026-07-07
TEST: Add tests for triangle.py.

genedan:
## Summary of Changes  

Add tests for uncovered lines in `triangle.py`.

@henrydingliu, could you submit a PR testing for these lines for the disposal rate method? I think you'd be able to create a better test than I can:

https://github.com/casact/chainladder-python/blob/c45d10b3ae04a066e1825c7624bcf36e11e0f468/chainladder/core/triangle.py#L1193-L1195

I left any lines dealing with dask/cupy uncovered.

---

> [!NOTE]
> **Low Risk**
> Test-only changes plus documentation; no runtime logic changes in the diff.
> 
> **Overview**
> Adds broad **test coverage** in `test_triangle.py` for previously untested `Triangle` behaviors, plus a small **docstring clarification** on `array_backend` in `triangle.py`.
> 
> New tests lock in contracts for **`link_ratio`** (pattern flags/shape, zero ratios → NaN, idempotent on pattern triangles), **`trend`** / **`shift`** (invalid axes and no-op cases), **`sort_axis`** (values permute with labels on columns/origin/development), and **index reassignment** via the `index` setter and **`set_index`** (DataFrame-only, length checks, inplace vs copy).
> 
> Constructor / backend routing is covered with cases for defaulting to **`cl.options.ARRAY_BACKEND`**, **`set_backend`** when **`AUTO_SPARSE`** is off, and auto-sparse overriding an explicit or option-based backend when **`AUTO_SPARSE`** is on. **`dev_to_val(inplace=True)`** on an already valuation triangle is asserted to return the same object.
> 
> The rest of the diff is largely **formatting** (quotes, wrapping) in existing tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4a1a6365a6ad398ec1b28a2384fb6b1e3fa1bd53. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chainladder/core/tests/test_triangle.py
```diff
@@ -230,10 +325,151 @@ def test_origin_and_value_setters(raa):
...
+    """
+    tri = clrd.iloc[:3]
+    new_index = pd.DataFrame({"Company": ["A", "B", "C"]})
+
+    tri.index = new_index
+
+    assert tri.key_labels == ["Company"]
+    np.testing.assert_array_equal(tri.kdims, new_index.values)
+    # _set_slicers() must have rebuilt .loc against the new key label.
+    assert tri.loc["A"].kdims.tolist() == [["A"]]
```
henrydingliu, line 357: would a stronger test be `tri.loc["A"] == clrd.iloc[:1]`?

chainladder/core/tests/test_triangle.py
```diff
@@ -230,10 +325,151 @@ def test_origin_and_value_setters(raa):
...
+        The clrd sample dataset Triangle.
+
+    Returns
+    -------
+    None
+    """
+    tri = clrd.iloc[:3]
+    new_index = pd.DataFrame({"Company": ["A", "B", "C"]})
+
+    result = tri.set_index(new_index, inplace=True)
```
henrydingliu, line 418: should set_index be used instead of the index setter? what's the best practice?

chainladder/core/tests/test_triangle.py
```diff
@@ -230,10 +325,151 @@ def test_origin_and_value_setters(raa):
...
+        The qtr sample dataset Triangle.
+
+    Returns
+    -------
+    None
+    """
+    val_tri = qtr.dev_to_val()
+    assert val_tri.is_val_tri
+
+    result = val_tri.dev_to_val(inplace=True)
```
henrydingliu, line 475: inplace=False?

chainladder/core/tests/test_triangle.py
```diff
@@ -429,6 +665,67 @@ def test_auto_sparse_disabled_returns_self(prism: Triangle) -> None:
         cl.options.reset_option("AUTO_SPARSE")
 
 
+def test_init_defaults_array_backend_to_option() -> None:
+    """
+    When array_backend is not passed to the constructor (i.e. it is None),
```
henrydingliu, line 985: why does this test feel like a deja vu?

chainladder/core/tests/test_triangle.py
```diff
@@ -429,6 +665,67 @@ def test_auto_sparse_disabled_returns_self(prism: Triangle) -> None:
...
+            columns="value",
+            cumulative=True,
+        )
+        assert tri.array_backend == "sparse"
+    finally:
+        cl.options.reset_option("AUTO_SPARSE")
+        cl.options.reset_option("ARRAY_BACKEND")
+
+
+def test_init_calls_set_backend_when_auto_sparse_disabled() -> None:
```
henrydingliu, line 1014: the [implementation](https://github.com/casact/chainladder-python/blame/c45d10b3ae04a066e1825c7624bcf36e11e0f468/chainladder/core/triangle.py#L557-L566) is kinda messy (maybe a small refactor is in order). i think we can benefit from a more consolidated test, i.e. 
```python
#the parameter supplied to the constructor will always override everything else (no idea if this is true, just making up something
assert cl.Triangle(backend = 'numpy').array_backend == 'numpy'

#if parameter is left out, cl.options.AUTO_SPARSE takes precedence
assert ...
```

chainladder/core/tests/test_triangle.py
```diff
@@ -747,6 +1044,104 @@ def test_sort_axis(clrd):
...
+    Parameters
+    ----------
+    raa : Triangle
+        The RAA sample dataset Triangle.
+
+    Returns
+    -------
+    None
+    """
+    reversed_tr = raa.iloc[..., ::-1, :]
```
henrydingliu, line 1487: this is remarkable. and the raa.origin gets reversed from the iloc? did we have a test like this in the test_slicing?

chainladder/core/tests/test_triangle.py
genedan: In Pandas, you can do both:

```python
import pandas as pd
test = pd.DataFrame({'a': [1, 2, 3], 'b': [4, 5, 6]})
test.set_index(pd.Index([5, 6, 7], name="Ballstate"))

           a  b
Ballstate      
5          1  4
6          2  5
7          3  6

test.index = pd.Index([5, 6, 7], name="Ballstate")
test

           a  b
Ballstate      
5          1  4
6          2  5
7          3  6
```


This test is legit behavior, but there are inconsistencies in how a Triangle index works and how a DataFrame index works. This would be stuff like the Triangle index being a DataFrame instead of a MultiIndex, and the way you can use `set_index()` after initialization:

```python

test = pd.DataFrame({'a': [1, 2, 3], 'b': [4, 5, 6]})
test.set_index('a')
```

Is something you can do in Pandas but not in Chainladder. The index, along with ways to set it, could use some refinement. I'll open an issue for further discussion.

genedan: Actually, most of that refinement would depend on replacing the Triangle index with an index/multiindex, which we already have a discussion for. Another would be whether we'd store any metadata outside the 4 axes for re-indexing later on (unlikely, but could open an issue later if anyone ever requests it).

henrydingliu: thanks. so presently there's no recommendation to use `set_index()` over direct setting

genedan: If I had to recommend one over the other, I'd use `set_index()` with `inplace=False`. Using the setter directly would mutate the object, something that's discouraged in functional programming.

chainladder/core/tests/test_triangle.py
genedan: This test revealed newly discovered behavior to me. I didn't even know this was possible (or when it might be used in real life).

henrydingliu:
@genedan checking in to see if you've had a chance to look through my comments?

genedan:
Still working on it - just wanted to prioritize the contributing guidelines and 0.10.0 release work, I'll get back to this once those are done.

chainladder/core/tests/test_triangle.py
genedan: Added, but I'm not sure if this is desirable behavior. The two triangles have different index values, but testing their equality results in True:

```shell
clrd.iloc[:1].index
Out[3]: 
            GRNAME      LOB
0  Adriatic Ins Co  othliab
tri.loc["A"].index
Out[4]: 
  Company
0       A
tri.loc["A"] == clrd.iloc[:1]
Out[5]: np.True_
```

chainladder/core/tests/test_triangle.py
genedan: It should be inplace=True, since the test is deliberately testing the inplace feature.

chainladder/core/tests/test_triangle.py
genedan: Hmmm nevermind. You can actually do this in Pandas. I guess I'm not a huge fan of this ability but we should be ok.

```
df = pd.DataFrame({"Sales": [100, 200, 300]}, index=["Apple", "Banana", "Cherry"])
# Relabel specific index values
df_new = df.rename(index={"Apple": "Green Apple", "Cherry": "Berry"})
df.loc['Apple'] == df_new.loc['Green Apple']
Out[2]: 
Sales    True
dtype: bool
```

chainladder/core/tests/test_triangle.py
genedan: Not sure, but it's meant to test lines 640-641 of `triangle.py`, which are still uncovered:

https://app.codecov.io/github/casact/chainladder-python/blob/main/chainladder%2Fcore%2Ftriangle.py#L640

chainladder/core/tests/test_triangle.py
henrydingliu: > I guess I'm not a huge fan of this ability but we should be ok.

would it change your mind to consider this ability in this context?

```python
df.loc["A"] = df.loc["B"]
assert df.loc["A"] == df.loc["B"]
```

chainladder/core/tests/test_triangle.py
henrydingliu: but the assert in the next line would still pass if inplace=False?

chainladder/core/tests/test_triangle.py
genedan: Yeah, we're  good.

chainladder/core/tests/test_triangle.py
genedan: Ah, you just uncovered an inconsistency between chainladder and pandas. When pandas does `inplace` stuff, it mutates the DataFrame and returns `None`. But chainladder's practice is to return `self`:

```python
        if self.is_val_tri:
            if inplace:
                return self
```

So that's why we're assigning to the variable `result`, it's testing the `return self` line.  To make things consistent with pandas, we'll need to change this behavior to return `None` for a large number of methods. I think we should keep the chainladder behavior for now and address the inconsistencies in a new issue.

chainladder/core/tests/test_triangle.py
genedan: > #the parameter supplied to the constructor will always override everything else (no idea if this is true, just making up something
assert cl.Triangle(backend = 'numpy').array_backend == 'numpy'

When `AUTO_SPARSE` is True, it'll override `array_backend`.

>#if parameter is left out, cl.options.AUTO_SPARSE takes precedence
assert ...

When the `array_backend` is left out, the package option ARRAY_BACKEND is the fallback, subject to AUTO_SPARSE, which also overrides `ARRAY_BACKEND`.

I've added two new tests to illustrate the behavior:

- `test_init_array_backend_overridden_by_auto_sparse_when_enabled()`
- `test_init_array_backend_option_overridden_by_auto_sparse_when_param_omitted()`

 I also updated the constructor docstring in `Triangle.py` to clarify.

genedan:
@henrydingliu , I finally got around to this, thanks for your patience.

chainladder/core/tests/test_triangle.py
henrydingliu: > But chainladder's practice is to return self

it is not mutating then? if it's not mutating, then it's a bug we need to fix. if it's mutating but returning self anyway, then this snippet isn't testing the mutation.

henrydingliu: ah, nvm, read the dostring. we good

henrydingliu (approved):

PR #1149, opened 2026-07-24
Apply lognormal fix for apriori draws and add tests

### LINKED ISSUE #1143: BornhuetterFerguson samples from normal not lognormal as per docstring

### Describe the bug in words

The apriori_sigma parameter on BornhuetterFerguson, Benktander, CapeCod, and ExpectedLoss is documented as drawing the stochastic apriori from a lognormal distribution:

```apriori_sigma: … the model samples aprioris from a lognormal distribution using this argument as a standard deviation.```

But the implementation samples from a Normal distribution. In chainladder/methods/benktander.py, `_get_benktander_aprioris`:

```python
if self.apriori_sigma != 0:
    random_state = xp.random.RandomState(self.random_state)
    apriori = random_state.normal(self.apriori, self.apriori_sigma, X.shape[0])   # <-- Normal, not lognormal
    ...
    apriori = sample_weight * apriori
```

The drawn value is used as the apriori multiplier directly — there is no exp()/log transform anywhere and the sole random draw is this .normal(...)).

```python
import numpy as np
import chainladder as cl

tri  = cl.load_sample("genins")
boot = cl.BootstrapODPSample(n_sims=50_000, random_state=42).fit_transform(tri)

# sample_weight = 1 everywhere, so expectation_ equals the raw apriori multiplier
w = boot.latest_diagonal.copy()
w.values = np.ones_like(w.values)

sigma = 0.8
bf = cl.BornhuetterFerguson(apriori=1.0, apriori_sigma=sigma, random_state=7).fit(boot, sample_weight=w)

mult = np.nanmean(bf.expectation_.values[:, 0, :, 0], axis=1)
mult = mult[np.isfinite(mult)]

print("mean          :", mult.mean())         # ~0.997  (centered on apriori=1.0)
print("std           :", mult.std())          # ~0.795  (== apriori_sigma)
print("min           :", mult.min())          # ~ -2.3  -> impossible for a lognormal
print("frac negative :", (mult < 0).mean())   # ~0.105  == Normal(1, 0.8) P(X<0) = Φ(-1.25)
```
Observed output:

mean          : 0.9966
std           : 0.7950
min           : -2.2916
frac negative : 0.1048

The negative minimum and ~10.5% negative fraction (matching Φ(-1/0.8)=0.1056), together with near-zero skew, confirm a Normal draw. A lognormal would give a strictly positive, right-skewed sample.

Expected behaviour: the docstring promises a lognormal apriori, which is strictly positive. This assertion holds once the bug is fixed; it fail on the current code above, which samples a Normal and produces negative aprioris.
```python
assert (mult > 0).all(), (
    f"apriori_sigma should draw a strictly-positive lognormal apriori, but "
    f"{(mult < 0).mean():.1%} of sampled aprioris are negative (min={mult.min():.3f})"
)
```

- [ ] Yes, absolutely!
- [x] Yes, but I would like some help.
- [ ] No.

friman-howard:
## Summary of Changes  
Switches the stochastic apriori sampler in Benktander._get_benktander_aprioris from a normal to a lognormal distribution.

- Bug: aprioris were drawn with random_state.normal(apriori, apriori_sigma, …). This is producing negative expected ultimates — which are meaningless for a loss reserve.
- Fix: draw from a lognormal instead, parametrized by method-of-moments so the mean stays apriori and the SD stays apriori_sigma
- Scope: BornhuetterFerguson and CapeCod both subclass Benktander and route their apriori scatter through this same method, so the one-line fix covers all three stochastic-apriori methods.
- Tests: adds two regression tests in test_benktander.py:
  - test_bf_apriori_sigma_is_lognormal — asserts BF draws are strictly positive and the approximate equality of empirical mean = apriori and SD = apriori_sigma.
  - test_capecod_apriori_sigma_is_positive — asserts CapeCod aprioris stay positive (around 10% would be negative under the old normal draw).
  
Currently the fix assumes apriori > 0 but it may be worth adding a guard.

## Related GitHub Issue(s)  
Fixes #1143.

---

> [!NOTE]
> **Medium Risk**
> Changes stochastic reserve math used by BF, Benktander, and CapeCod when apriori_sigma is set; deterministic paths (sigma=0) are unchanged.
> 
> **Overview**
> **Stochastic apriori draws** in `_get_benktander_aprioris` no longer use a normal distribution (which could yield negative loss expectations). They now use a **lognormal** draw with method-of-moments parameters so the mean stays `apriori` and the standard deviation stays `apriori_sigma`.
> 
> Because **BornhuetterFerguson** and **CapeCod** inherit Benktander and share this sampler, the behavior change applies to all three methods when `apriori_sigma != 0` (e.g. bootstrap workflows).
> 
> **Tests** add regression coverage: BF multipliers are strictly positive with approximate mean/SD matching `apriori` / `apriori_sigma`, and CapeCod `expectation_` values stay positive.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 69a84b669c9a5ee773130df8f6ffeca12ee02997. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/chainladder/methods/benktander.py
+++ b/chainladder/methods/benktander.py
@@ -254,7 +254,11 @@ class Benktander(MethodBase):
         xp = X.get_array_module()
         if self.apriori_sigma != 0:
             random_state = xp.random.RandomState(self.random_state)
-            apriori = random_state.normal(self.apriori, self.apriori_sigma, X.shape[0])
+            # Draw from lognormal with E[apriori] = self.apriori and SD = self.apriori_sigma.
+            cov = self.apriori_sigma / self.apriori
+            sigma_log = np.sqrt(np.log1p(cov ** 2))
+            mu_log = np.log(self.apriori) - 0.5 * sigma_log ** 2
+            apriori = random_state.lognormal(mu_log, sigma_log, X.shape[0])
             apriori = apriori.reshape(X.shape[0], -1)[..., None, None]
             apriori = sample_weight * apriori
             apriori.kdims = X.kdims
--- a/chainladder/methods/tests/test_benktander.py
+++ b/chainladder/methods/tests/test_benktander.py
@@ -76,3 +76,36 @@ def test_odd_shaped_triangle():
     ult1 = cl.Benktander(apriori = 1,n_iters=10000).fit(cl.Development(average="volume").fit_transform(atr),sample_weight = atr.latest_diagonal).ultimate_.sum()
     ult2 = cl.Benktander(apriori = 1,n_iters=10000).fit(cl.Development(average="volume").fit_transform(tr),sample_weight = tr.latest_diagonal).ultimate_.grain("OYDQ").sum()
     assert abs(ult1 - ult2) < 1e-5
+
+
+def test_bf_apriori_sigma_is_lognormal():
+    """apriori_sigma must draw a strictly-positive lognormal apriori (issue #1143)."""
+    tri = cl.load_sample("genins")
+    boot = cl.BootstrapODPSample(n_sims=50000, random_state=42).fit_transform(tri)
+    w = boot.latest_diagonal.copy()
+    w.values = np.ones_like(w.values)   # sample_weight=1 -> expectation_ == raw multiplier
+
+    sigma = 0.8
+    bf = cl.BornhuetterFerguson(
+        apriori=1.0, apriori_sigma=sigma, random_state=7
+    ).fit(boot, sample_weight=w)
+
+    mult = np.nanmean(bf.expectation_.values[:, 0, :, 0], axis=1)
+    mult = mult[np.isfinite(mult)]
+
+    assert (mult > 0).all()                 # strictly positive -> lognormal, not normal
+    assert abs(mult.mean() - 1.0) < 0.02    # mean preserved (E = apriori)
+    assert abs(mult.std() - sigma) < 0.03   # SD preserved (== apriori_sigma)
+
+
+def test_capecod_apriori_sigma_is_positive():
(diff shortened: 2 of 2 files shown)


kennethshsu:
Hi @friman-howard, welcome to the project, and thanks so much for this PR!

Does it make sense to preserve the normal version too and let the user select the distribution instead? How did you come across this bug? I wonder if this was an error in the documentation instead of the implementation. Do you have research papers that show this with examples?

friman-howard:
Hi @kennethshsu ,

I came across this bug while building a stochastic Bornhuetter-Ferguson reserve-risk model on top of chainladder. When I turned on apriori_sigma I was getting negative a-priori ultimates, traced it back to _get_benktander_aprioris, and found it calls random_state.normal(...) rather than sampling a lognormal as the documentation suggests. 
There are two reasons why I assumed this wasn't an error in the documentation:
1) The documentation references sampling using lognormal in 4 places but the implementation defect is a single shared line of code
2) An a-priori ultimate is a strictly positive quantity, and a normal draws values below zero. I don't think any papers explicitly reference a distribution but for strictly positive support, parameterised by CoV, a lognormal is a natural choice, as would be a gamma distribution but a normal would not be in my opinion.

I'm happy to add an option to use the normal version instead of the lognormal if you would like but I wouldn't recommend  it. Also very happy to be corrected if I have misunderstood anything.

kennethshsu (approved):

kennethshsu:
Ok great, thanks so much for following up!

Approved, thanks so much for your contribution!

PR #1152, opened 2026-07-25
Fix Triangle display crash when stored in a DataFrame cell (#142)

### LINKED ISSUE #142: Error storing <class 'chainladder.core.triangle.Triangle'> object in pd.DataFrame
Issue as per title., other classes in the package do not cause this error.

Tidiest workaround I've round is to pickle & unpickle the object.

Example:

```
import chainladder as cl

tri_name = 'clrd'
tri_obj = cl.load_sample(tri_name)
df_fails = pd.DataFrame(data=[[tri_name, tri_obj]], columns=['name', 'cl_triangle'])

#work around
import pickle
pickle_tri = pickle.dumps(tri_obj)
df = pd.DataFrame(data=[[tri_name, pickle_tri]], columns=['name', 'cl_triangle'])
tri_obj = pickle.loads(df.loc[0,'cl_triangle'])

```

priyam0k:
## Summary of Changes

storing a triangle in a dataframe cell used to crash the display (both text and html repr).

the triangle is a 4d container, but pandas' `is_sequence()` treated it as a 1d sequence. the legacy `__getitem__` iteration protocol plus `__len__` made `iter()` succeed, so pandas tried to iterate the triangle while formatting the cell. that hit `__getitem__(0)`, which reads the int as a column label and raises.

fix is to declare the triangle non-iterable with `__iter__ = None` in `core/dunders.py`. pandas then falls back to `str(triangle)` and renders the summary. nothing in the codebase iterates a triangle directly so it's safe, and `np.asarray()` still works.

added two regression tests in `core/tests/test_display.py`:
- `test_triangle_not_iterable`
- `test_triangle_in_dataframe_cell_display` (the exact case from the issue)

## Related GitHub Issue(s)

fixes #142

no code iterated triangles directly, so making it non-iterable has no downstream impact. verified the exact repro from the issue works for `repr(df)`, `df._repr_html_()` and the individual cell, `np.asarray(tri)` still returns the full `(775, 6, 10, 10)` array, and all 39 display tests plus the full 515 core tests pass locally.

---

> [!NOTE]
> **Low Risk**
> Small protocol change on Triangle display/iteration with no in-repo consumers that iterate Triangle directly; behavior change is limited to failing `iter(triangle)` where iteration was already incorrect.
> 
> **Overview**
> Fixes **GH #142** by setting **`__iter__ = None`** on `Triangle` in `core/dunders.py`, so Python no longer uses the legacy `__getitem__(0), …` iteration path (which misread indices as column labels and raised). Pandas then treats a cell `Triangle` as non-sequence and formats it via **`str`/`repr`** instead of iterating.
> 
> Adds **`test_triangle_not_iterable`** and **`test_triangle_in_dataframe_cell_display`** in `core/tests/test_display.py` to lock in `iter()` failure and safe `repr` / `_repr_html_` for DataFrames holding a `Triangle`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 418c1679c5174515ff85e4734d1a52c097a22edc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/chainladder/core/dunders.py
+++ b/chainladder/core/dunders.py
@@ -290,6 +290,16 @@ class TriangleDunders:
     def __len__(self):
         return self.shape[0]
 
+    # A Triangle is a 4-D container, not a 1-D sequence. Without this, Python
+    # falls back to the legacy iteration protocol (repeatedly calling
+    # __getitem__(0), __getitem__(1), ...), which treats the integer as a
+    # column label and raises. That also makes libraries such as pandas
+    # misclassify a Triangle as a sequence (see is_sequence) and attempt to
+    # iterate it when formatting a Triangle stored in a DataFrame cell,
+    # crashing the display. Declaring the type non-iterable makes pandas fall
+    # back to str(triangle) and render the summary instead (GH #142).
+    __iter__ = None
+
     def __neg__(self):
         obj = self.copy()
         obj.values = -obj.values
--- a/chainladder/core/tests/test_display.py
+++ b/chainladder/core/tests/test_display.py
@@ -8,6 +8,7 @@ import pytest
 import sys
 
 
+from collections.abc import Iterable
 from chainladder.core.display import TriangleDisplay
 from lxml import etree, html as lxml_html
 from unittest import mock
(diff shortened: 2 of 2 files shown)


henrydingliu:
@priyam0k thanks for picking this one up. 

the fix you are proposing is elegant in and of itself. but the result is somewhat less pleasant. 

Image

the pandas equivalent, i.e. stuffing pandas dataframes inside pandas, is equally void of utility, but maybe slightly more pleasant

Image

i can live with the `\n`. but would prefer a cleaner repr. ultimately this issue isn't worth much more of our time. so let me know if you'd like to spend a bit more time on it. i'm happy to approve as is just so get rid of the bug.

priyam0k:
thanks henry. agree it's not worth much more time here, the main goal was just killing the crash.

the cell repr is awkward because pandas formats object cells with str(), which for a triangle falls through to __repr__ (the full summary). the only clean lever would be a __str__ one-liner, but that also changes print(triangle) for everyone, which feels like a worse tradeoff for a nested-in-dataframe case that isn't really useful anyway.

so i'd lean towards merging as is if you're good with it. happy to revisit later if it ever bugs someone.

henrydingliu (approved):

PR #1131, opened 2026-07-15
Add index, columns, origin, development alternatives to Triangle.drop()

priyam0k:
## Summary of Changes

Adds `index`, `columns`, `origin`, and `development` keyword alternatives to `Triangle.drop()`, mirroring `DataFrame.drop()`.

- each alternative routes to its axis, so `columns=...` equals `labels=..., axis=1`, `index=...` equals `axis=0`, and so on
- specifying `labels` together with any alternative raises `ValueError`, matching pandas
- `columns=` works today since the column axis is implemented; `index`, `origin`, and `development` route to their axis and still raise `NotImplementedError` until those axes land in #1051, #1055, #1057
- added tests for the alternatives, the list form, the conflict error, and the not-yet-implemented axes

## Related GitHub Issue(s)

closes #1060, part of #1052

## Additional Context for Reviewers

- kept the `labels=None` with nothing specified path as is, that is handled separately in #1062
- no behavior change for the existing `labels`/`axis` form

- [x] I passed tests locally for code (`pytest`)

---

> [!NOTE]
> **Low Risk**
> API extension with preserved behavior for labels/axis; column drops use existing selection logic and are covered by new tests.
> 
> **Overview**
> **`Triangle.drop()`** now accepts **`index`**, **`columns`**, **`origin`**, and **`development`** as alternatives to **`labels`** + **`axis`**, aligned with **`DataFrame.drop()`**. For example, **`drop(columns='CumPaidLoss')`** matches **`drop(labels='CumPaidLoss', axis=1)`**. Mixing **`labels`** with any of those kwargs raises **`ValueError`**.
> 
> Implementation builds a per-axis drop map, normalizes scalars with **`np.isscalar`** (so integer development labels do not hit the old **`type(labels) is str`** bug), and can apply multiple axis kwargs in one call. Only the column axis still performs drops; **`index`**, **`origin`**, and **`development`** route to their axes and raise **`NotImplementedError`** until follow-up work lands. The **`labels`/`axis`** API is unchanged.
> 
> **`TriangleProtocol`** and docs/examples were updated; tests cover equivalence, lists, conflicts, unimplemented axes, and list-like **`columns`** inputs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 65f70d05f21742774de8df3b40edb3a165380a2b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

Issue #1057, opened 2026-06-27
Allow Triangle.drop() to drop development periods


genedan:
### Description

`Triangle.drop()` should be able to drop a development period when `axis=3`.

### Is your feature request aligned with the scope of the package?

- [x] Yes, absolutely!
- [ ] No, but it's still worth discussing.
- [ ] N/A (this request is not a codebase enhancement).

### Describe the solution you'd like, or your current workaround.

```python
tri = cl.Triangle(
    data={
        'origin': [1985, 1985, 1985, 1986, 1986, 1987],
        'development': [1985, 1986, 1987, 1986, 1987, 1987],
        'paid': [300, 400, 500, 500, 600, 500],
    },
    origin='origin',
    development='development',
    columns=['paid'],
    cumulative=True
)
print(tri)

         12     24     36
1985  300.0  400.0  500.0
1986  500.0  600.0    NaN
1987  500.0    NaN    NaN
```

```python
print(tri.drop(labels='12`, axis=3)

          24     36
1985   400.0  500.0
1986   600.0    NaN
1987     NaN    NaN
```

### Would you be willing to contribute this ticket?

- [ ] Yes, absolutely!
- [ ] Yes, but I would like some help.
- [ ] No.

Issue #1052, opened 2026-06-27
Add parameters to  Triangle.drop()


genedan:
### Description

`Triangle.drop()` can only drop a column at this point. It can be further enhanced by incorporating parameters from the Pandas analogue [DataFrame.drop()](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.drop.html):

<img width="1221" height="873" alt="Image" src="https://github.com/user-attachments/assets/a319633c-8275-4d51-98e7-b45b83734ddf" />

There's a lot to be added, we could split responsibility by dividing these into sub-issues.

### Is your feature request aligned with the scope of the package?

- [x] Yes, absolutely!
- [ ] No, but it's still worth discussing.
- [ ] N/A (this request is not a codebase enhancement).

### Describe the solution you'd like, or your current workaround.

See sub-issues for examples, but the following parameters should be implemented/expanded.

- axis
  -  allow string assignment
  -  enable index dropping
  -  enable origin dropping (outer levels)
  -  enable development dropping (outer levels)
- equivalents
  - index
  - columns
  - origin
  - development
- level
- errors (low priority)

### Do you have any additional supporting notes?

`inplace` does not need to be implemented, since it produces side-effects, a programming style which has fallen out of favor.

### Would you be willing to contribute this ticket?

- [ ] Yes, absolutely!
- [x] Yes, but I would like some help.
- [ ] No.

priyam0k:
I’d like to take this one. I’ll start with a focused PR for axis support and dropping index/columns, then follow up separately for origin, development, level, and errors if that sounds good.

priyam0k:
quick status and a process question. origin dropping is up in #1163 (closes #1055), just waiting on a review, bugbot's finding is resolved and tests are in.

for the rest of this umbrella (#1057 development dropping, #1062 labels=None, #1061 level), they all touch the same drop() method, so separate parallel PRs would mostly just conflict with each other. i'd lean towards finishing them in one branch off main once #1163 lands, so `drop()` gets completed end to end in a single reviewable pass rather than three stacked PRs. i'll keep the commits split per sub-issue so it's still easy to follow.

happy to go one-PR-per-issue instead if you'd prefer to keep them small, just let me know which reads better. @genedan let me know your take

genedan:
Sounds good, thanks for the help!

priyam0k:
Hi @genedan - i finished #1057, #1062, and #1061 in `feature/drop-development`.

[...]

PR #1163, opened 2026-07-29
Allow Triangle.drop() to drop origin levels

### LINKED ISSUE #1055: Allow Triangle.drop() to drop origin levels
### Description

Triangle.drop() should allow for the dropping of origin levels when `axis=2`.

- [x] Yes, absolutely!

```python
tri = cl.Triangle(
    data={
        'origin': [1985, 1985, 1985, 1986, 1986, 1987],
        'development': [1985, 1986, 1987, 1986, 1987, 1987],
        'paid': [300, 400, 500, 500, 600, 500],
    },
    origin='origin',
    development='development',
    columns=['paid'],
    cumulative=True
)
print(tri)

         12     24     36
1985  300.0  400.0  500.0
1986  500.0  600.0    NaN
1987  500.0    NaN    NaN
```

```python
print(tri.drop(labels='1987', axis='2')

         12     24     36
1985  300.0  400.0  500.0
1986  500.0  600.0    NaN
```

Dropping of labels that are not the first or last origin period should not be allowed.

- [ ] Yes, absolutely!
- [ ] No.

priyam0k:
## summary of changes

adds origin-level dropping to `Triangle.drop()` via `origin=` or `axis=2` / `axis='origin'`, building on #1131.

- `tri.drop(origin='1990')` and `tri.drop(labels='1990', axis=2)` are equivalent
- only the first or last origin period(s) can be dropped, interior drops raise `ValueError`
- dropping an origin also trims any dev periods left entirely NaN, so it trims the triangle instead of leaving empty dev columns
- unknown origin labels raise `KeyError`
- added tests and doctest examples covering the above

## related github issue(s)

closes #1055, part of #1052

## additional context for reviewers

- the origin / `axis=2` path from #1131 used to raise `NotImplementedError`, now implemented. updated that test to only assert it for the still-unimplemented index and development axes
- dev-trimming reuses the same all-NaN edge detection as `dropna()`. if every dev period still has data (e.g. dropping the latest origin), the dev axis is left alone
- development-axis dropping (#1057) is a planned follow-up

- [x] i passed tests locally for code (`pytest`)

### DIFF
--- a/chainladder/core/pandas.py
+++ b/chainladder/core/pandas.py
@@ -682,6 +682,42 @@ class TrianglePandas(_TrianglePandasBase):
             ['1988', '1989', '1990']
             ['1987', '1988', '1989']
 
+        Dropping an origin period also trims any development periods that are
+        left entirely NaN, so the triangle is trimmed rather than left with
+        empty development columns.
+
+        .. testcode::
+
+            tri = cl.Triangle(
+                data={
+                    'origin': [1985, 1985, 1985, 1986, 1986, 1987],
+                    'development': [1985, 1986, 1987, 1986, 1987, 1987],
+                    'paid': [300, 400, 500, 500, 600, 500],
+                },
+                origin='origin',
+                development='development',
+                columns=['paid'],
+                cumulative=True
+            )
+            print(tri)
+
+        .. testoutput::
+
+                     12     24     36
+            1985  300.0  400.0  500.0
+            1986  500.0  600.0    NaN
+            1987  500.0    NaN    NaN
+
+        .. testcode::
+
+            print(tri.drop(origin='1985'))
+
+        .. testoutput::
+
+                     12     24
+            1986  500.0  600.0
+            1987  500.0    NaN
+
         """
         alternatives = {0: index, 1: columns, 2: origin, 3: development}
         if any(value is not None for value in alternatives.values()):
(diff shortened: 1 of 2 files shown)


chainladder/core/pandas.py
```diff
@@ -690,9 +704,29 @@ def drop(
...
+                    np.arange(kept_positions[0], kept_positions[-1] + 1),
+                ):
+                    raise ValueError(
+                        "Only the first or last origin periods may be dropped; "
+                        "dropping an interior origin period would leave a gap."
+                    )
+                result = result[keep]
             else:
                 raise NotImplementedError(
-                    "Triangle.drop() only implemented for column axis."
```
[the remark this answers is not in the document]
priyam0k: good catch. switched the scalar check from `np.isscalar` to `pd.api.types.is_list_like`, so a `pd.Period` (the origin axis's native label type, e.g. `tri.origin[-1]`) is treated as a scalar instead of hitting `list(period)`. `str`/`int`/`Period` are wrapped, real list-likes (`list`/`ndarray`/`Index`) still expand, and `None` stays on the `list()` path so the bare `drop()` behavior is unchanged. added a test dropping via a `pd.Period` label.

priyam0k:
pushed an update from the discussion with @henrydingliu in #1055. dropping an origin now also trims development periods left all NaN, so it trims the triangle instead of acting like plain 4d slicing. added a docstring example plus tests for the trim and no-trim cases, all passing locally.

also, first time i tried cursor for this one. worked out well for navigating the codebase, though I can't rely on it entirely!

chainladder/core/pandas.py
```diff
@@ -684,15 +734,55 @@ def drop(
...
+                if ax_labels is None or pd.api.types.is_list_like(ax_labels)
+                else [ax_labels]
             )
             if ax == 1:
                 result = result[
                     [item for item in result.columns if item not in ax_labels]
                 ]
+            elif ax == 2:
+                origin_labels = np.array(result.origin.astype(str))
+                drop_labels = [str(label) for label in ax_labels]
```
henrydingliu, line 747: so this would be a whole new thing in this package. for example, `raa.loc[:,:,1982,:]` would fail. i'm a little torn. make the user's life easier? or make the package more consistent?

priyam0k: definitely a tough call between consistency and ergonomics!

my vote is to prioritize user experience first. .drop() feels like a higher level helper where trimming empty dev columns saves manual cleanup, whereas .loc is meant to be lower level 4d slicing.

kind of like how rust balances strict borrow checker rules with developer experience, leaning into ergonomics first makes the daily api much smoother. we can always standardize consistency across indexing later based on real user feedback!

PR #1180, opened 2026-08-01
Friedland Chapter 10 - Cape Cod Technique

priyam0k:
## Summary of Changes
Recreates Friedland Chapter 10 (Cape Cod) as a notebook with reconciliation asserts, Exhibits I–IV. Same work as #1142 (which went into experimental), rebased straight onto main now that every PR gets an RTD preview.

- Exhibits I (US Industry Auto), II (XYZ Auto BI, with on-level/trend/tort and the cross-method Sheets 4-5), III (US PP Auto, 4 scenarios), IV (US Auto product mix, 2 scenarios)
- Merges the three `friedland_us_auto_*` samples into one Scenario-indexed `friedland_us_auto.csv` with Earned Premium, mirroring `friedland_uspp`; old files removed
- Exhibit tables now pull Latest/CDF/Ultimate/IBNR from `cl.model_diagnostics`

https://github.com/casact/chainladder-python/pull/1180 (Chapter 10): its Exhibit II tort step can drop the [-0.1067, -0.25] conversion for a clean [0.67, 0.75, 1.0] with cumulative=True.

## Related GitHub Issue(s)
refers #588
supersedes #1142

## Additional Context for Reviewers
RTD preview builds from this draft. Reconciliation runs as asserts inside the notebook (`execute_notebooks: force`); all pass currently.

- [x] I passed tests locally for both code (`uv run pytest`) and documentation changes (`uv run jb build docs --builder=custom --custom-builder=doctest`)

---

> [!NOTE]
> **Low Risk**
> Documentation and bundled sample consolidation only; no changes to core reserving logic. Users of the three removed sample names need to migrate to the merged `friedland_us_auto` dataset.
> 
> **Overview**
> Adds **Friedland Chapter 10** (`chapter_10.ipynb`) to the docs TOC, walking through the Cape Cod method with `CapeCod`, `model_diagnostics`, and reconciliation asserts for Exhibits I–IV (U.S. Industry Auto, XYZ Auto BI with on-level/trend/tort, U.S. PP Auto scenarios, U.S. Auto product mix).
> 
> **Sample data:** Replaces three separate `friedland_us_auto_*` CSVs with a single **`friedland_us_auto`** dataset indexed by `Scenario`, including **Earned Premium** and both Steady State and Changing Product Mix rows—aligned with how `friedland_uspp` is structured.
> 
[...]

PR #1181, opened 2026-08-01
[ENH] Support cumulative on-level factors in ParallelogramOLF

### LINKED ISSUE #922: cl.ParallelogramOLF to support cumulative factors
Consider this example in column 11:

Image

We can use the cl.ParallelogramOLF to trend, with a bit of gymnastic.

> ```python
> friedland_gl_self_insurer = cl.load_sample("friedland_gl_self_insurer")
> 
> tort = pd.DataFrame({
>     "EffDate": ["2003-01-01", "2004-01-01", "2005-01-01"],
>     "RateChange": [(0.67/0.75) - 1, 0.75-1, 0.00]
> })
> 
> olf = cl.ParallelogramOLF(
>     rate_history=tort,
>     change_col="RateChange",
>     date_col="EffDate",
>     policy_length=12,
>     approximation_grain="M",
>     vertical_line=True
> ).fit_transform(friedland_gl_self_insurer["Reported Claims"]).olf_
> olf
> ```
> ```python
> 	2008
> 1998	0.6700
> 1999	0.6700
> 2000	0.6700
> 2001	0.6700
> 2002	0.6700
> 2003	0.7500
> 2004	1.0000
> 2005	1.0000
> 2006	1.0000
> 2007	1.0000
> 2008	1.0000
> ``` 

Can we make the rate change a bit easier?

 _Originally posted by @kennethshsu in [#582](https://github.com/casact/chainladder-python/issues/582#issuecomment-4626880017)_

priyam0k:
## Summary of Changes
Adds a `cumulative` flag to `ParallelogramOLF` and `parallelogram_olf` so a rate history can be given as cumulative on-level factors instead of incremental rate changes.

- `cumulative=False` (default) is the current behavior, fully backwards compatible.
- `cumulative=True` reads `change_col` as cumulative factors: each is in force from its effective date to the next, the earliest extends back over the lookback window, and the rate level is `1/factor`. Rolling average, `vertical_line`, and leap-year handling are unchanged. Non-positive factors raise a `ValueError`.

So the tort example drops the gymnastics:
```python
# before
"RateChange": [(0.67 / 0.75) - 1, 0.75 - 1, 0.00]
# after
"Factor": [0.67, 0.75, 1.00]  # cumulative=True
```

## Related GitHub Issue(s)
closes #922

Also completes a follow-up from #1180 (Chapter 10): its Exhibit II tort step can drop the `[-0.1067, -0.25]` conversion for a clean `[0.67, 0.75, 1.0]` with `cumulative=True`.

New tests in `test_parallelogram.py`: the issue's tort example, a cumulative-vs-incremental equivalence check across `M`/`D` grain and `vertical_line` True/False, and a non-positive-factor guard.

---

> [!NOTE]
> **Medium Risk**
> Touches actuarial on-leveling math in a shared utility, but the new path is gated behind `cumulative=True` with broad regression tests; incorrect factor interpretation would still affect premium leveling workflows.
> 
> **Overview**
> Adds **`cumulative`** (default `False`) to **`ParallelogramOLF`** and **`parallelogram_olf`**, so rate history can use **cumulative on-level factors** instead of incremental decimal rate changes—e.g. tort reform factors `[0.67, 0.75, 1.00]` without converting to increments.
> 
> With **`cumulative=True`**, each factor applies from its effective date until the next; the earliest factor is extended backward over the lookback window. **`fit`** routes duplicate effective dates through **`_combine_duplicate_dates`**: last factor wins (not a product). The utility layer maps factors to implied rate levels via **`1/factor`**, assigns levels with backward **`searchsorted`**, and rejects non-positive factors with **`ValueError`**. Incremental behavior and defaults are unchanged.
> 
> New tests cover the GH #922 tort case, equivalence to incremental inputs across **`M`/`D`** and **`vertical_line`**, pre-window factors, duplicate dates, and the positive-factor guard.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ebf009d4a56dad6d8b6c23beb2dad06c9a589f2a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/chainladder/utils/utility_functions.py
+++ b/chainladder/utils/utility_functions.py
@@ -465,12 +465,21 @@ def parallelogram_olf(
     )
 
     if cumulative:
-        factors = pd.Series(np.array(values, dtype="float64"), np.array(dates))
+        factors = pd.Series(
+            np.array(values, dtype="float64"), pd.to_datetime(np.array(dates))
+        ).sort_index()
         if (factors <= 0).any():
             raise ValueError("cumulative on-level factors must be positive")
-        # An on-level factor is the current rate level divided by the rate
-        # level in force, so the implied rate level is its reciprocal.
-        cum_rate_changes = (1 / factors).reindex(date_idx).ffill().bfill()
+        # An on-level factor is the current rate level divided by the rate level
+        # in force, so the implied rate level is its reciprocal. Each factor is
+        # in force from its effective date until the next (a backward/asof match,
+        # so off-grid or pre-window dates are honored); dates before the first
+        # factor take the earliest, extending it back over the lookback window.
+        level = 1 / factors
+        pos = np.searchsorted(level.index.values, date_idx.values, side="right") - 1
+        cum_rate_changes = pd.Series(
+            level.values[np.clip(pos, 0, None)], index=date_idx
+        )
     else:
         rate_changes = pd.Series(np.array(values), np.array(dates)).reindex(
             date_idx, fill_value=0
--- a/chainladder/utils/utility_functions.py
+++ b/chainladder/utils/utility_functions.py
@@ -433,8 +433,16 @@ def parallelogram_olf(
     policy_length=12,
     approximation_grain="M",
     vertical_line=False,
+    cumulative=False,
 ):
-    """Parallelogram approach to on-leveling."""
+    """Parallelogram approach to on-leveling.
+
+    When ``cumulative`` is False (default), ``values`` are incremental rate
+    changes expressed as decimals.  When True, ``values`` are cumulative
+    on-level factors already stated relative to one another, and each value is
+    in force from its effective date until the next one.  The earliest value is
+    extended backwards to cover the lookback window.
+    """
     if approximation_grain not in ["M", "D"]:
         raise ValueError("approximation_grain must be M or D")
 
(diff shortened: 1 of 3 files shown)


chainladder/utils/utility_functions.py
```diff
@@ -456,12 +464,29 @@ def parallelogram_olf(
...
-        date_idx, fill_value=0
-    )
-    cum_rate_changes = pd.Series(
-        np.cumprod(1 + rate_changes.values), rate_changes.index
-    )
+    if cumulative:
+        factors = pd.Series(
+            np.array(values, dtype="float64"), pd.to_datetime(np.array(dates))
+        ).sort_index()
+        if (factors <= 0).any():
```
henrydingliu, line 472: this feels like a convention specific to this one tort reform setup that doesn't necessarily generalize. should we keep the input consistent between cumulative and incremental? i.e. it might be okay if we want to make the user call parallelogram(data = factor - 1)

priyam0k: image

incremental mode reads each value as a compounding change, so `[0.67, 0.75, 1.0]` becomes a rate level of `0.67 → 0.67*0.75 → …` instead of the flat levels the factors mean. to reproduce the factors with incremental input you actually need `[0.4925, -0.1067, -0.25]` (take `1/f`, then consecutive ratios), which is the reverse-cumprod gymnastics the issue was about.

on it being tort-specific: cumulative on-level factors aren't unique to this case, it's just how friedland and most on-level exhibits state them, a factor to current level rather than a series of changes. the flag just lets the user input what's already in the table.

on consistency: the columns stay identical (`date_col`/`change_col`), it's only an interpretation switch behind an explicit `cumulative=` opt-in defaulting to False, so existing calls are untouched.

that said, your call. if you'd rather not add the flag, the alternative is documenting the `1/f`-then-ratios conversion, but that's the step #922 wanted to remove. happy either way.

henrydingliu: I explained myself poorly. 

The current API calls for 0-centric change factors to be passed in. Take this excerpt from Chapter 10 

```
tort_history = pd.DataFrame({
    "date": ["1/1/2006", "1/1/2007"],
    "rate_change": [-0.1067, -0.25],
})
tort = cl.ParallelogramOLF(
    tort_history, change_col="rate_change", date_col="date", vertical_line=True)
```

the issue here, ofc, is that -0.1067 is a derived figure. For consistency's sake, I was expecting the cumulative API to look like this

```
tort_history = pd.DataFrame({
    "date": ["1/1/2006", "1/1/2007"],
    "rate_change": [-0.33, -0.25],
})
tort = cl.ParallelogramOLF(
    tort_history, change_col="rate_change", date_col="date", vertical_line=True, cumulative=True)
```

I'm not necessarily opposed to requiring 1-centric level factors as the input for cumulative. but we should add a bit more explicit explanation in the docstring, along with a dedicated example, if you choose to stick with 1-centric. will leave it up to you.

priyam0k: Makes sense! Sticking with 1-centric factors ([0.67, 0.75, 1.00]) feels best since it lets users pass values directly from Friedland/CAS exhibits without manual shifts.

I'll update the ParallelogramOLF and parallelogram_olf docstrings to explicitly state the 1-centric convention for cumulative=True and add a dedicated example showing it in action. Thanks!

PR #1183, opened 2026-08-03
Change 1d to 1D/update unittest trigger

henrydingliu:
## Summary of Changes  
updating 1d to 1D ahead of pandas deprecation
updating '*' to '**' to catch slash in branch name

## Related GitHub Issue(s)  
fixes #849 
fixes #1182 

## Additional Context for Reviewers  
no more deprecation warning for `d` [here](https://github.com/casact/chainladder-python/actions/runs/30785835466/job/91599117055#step:5:98)

## Checklist
- [X] I passed tests locally for both code (`uv run pytest`) and documentation changes (`uv run --directory docs jb build . --builder=custom --custom-builder=doctest`)

---

> [!NOTE]
> **Low Risk**
> Small compatibility and CI trigger tweaks with no auth, data, or behavioral logic changes beyond avoiding pandas deprecations.
> 
> **Overview**
> Updates the unit-test workflow so **push** runs on all branch names, including those with slashes, by changing the branch filter from `'*'` to `'**'`.
> 
> In **`TriangleBase`**, development-lag calculation now uses pandas’ **`1D`** day rounding instead of deprecated **`1d`**, and datetime mapping calls **`Series.map`** without the removed **`arg=`** keyword.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 337982fb445f4bc885e0e1187b0c4377e2a0b343. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #1191, opened 2026-08-03
Streamlined the unit tests and addressed unit tests bug

### LINKED ISSUE #1182: [BUG] unittest not triggered on branch names with slash

### Describe the bug in words

if a branch name has a slash in it, commits to that branch will not trigger unittest

no test at this branch

https://github.com/casact/chainladder-python/tree/bug/test

test is triggered

- [ ] Yes, absolutely!
- [ ] No.

kennethshsu:
## Summary of Changes  
Fixes 1182, previously, we used "*" only to catch the branch name, which can be escaped by "/", using "**" will fix that.
This fix goes one step further, as we are all familiar, there are WAAY too many tests that run on each PR. This PR will update that to run on every PR and push to main only, it will not rerun main multiple times.

## Related GitHub Issue(s)  
Closes #1182

## Checklist
- [ ] I passed tests locally for both code (`uv run pytest`) and documentation changes (`uv run --directory docs jb build . --builder=custom --custom-builder=doctest`)

---

> [!NOTE]
> **Low Risk**
> Only GitHub Actions trigger configuration changed; no application or test logic was modified.
> 
> **Overview**
> **CI trigger change** in `pytest.yml`: `push` no longer uses the `'*'` branch filter (which could miss branches with `/` in the name). It now runs unit tests only on pushes to **`main`**, while **`pull_request`** still runs the full matrix on every PR.
> 
> This cuts redundant workflow runs on feature branches and avoids duplicate runs when `main` is updated, without changing what runs on PRs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cc2051f61b57dc87f3312ded645bab5664d7c259. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -4,7 +4,7 @@ name: Unit Tests
 on:
   push:
     branches:
-    - '*'
+    - '**'
 
   pull_request:
 
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -4,7 +4,7 @@ name: Unit Tests
 on:
   push:
     branches:
-    - '**'
+    - main
 
   pull_request:
 


henrydingliu:
i'm a little worried about this approach. this workflow file is shared by all forks. so changing it to run on main only also make it harder for everyone to run tests on their forks. and since most contributors are using issue specific branches from their fork to raise PRs, this means basically no tests anywhere unless there's a PR into main. 

gene very recently addressed some of the 'test flood' issue by introducing concurrency guards. so we should already be in better shape. i implore you to reconsider this drastic reduction in testing frequency.

priyam0k (commented):
looks fine

priyam0k:
> since most contributors are using issue specific branches from their fork to raise PRs, this means basically no tests anywhere unless there's a PR into main.

missed this part, can we switch main to ** instead so ci still runs on feature branch pushes?

kennethshsu:
I guess I was thinking about conserving the CI resources here especially with that huge duplicator matrix when both `push` and `PR` run on the same commit.

> means basically no tests anywhere unless there's a PR into main

This is not true? On line 9, CIs run when there's a PR open, even on PRs into branches that's not `main`, that's why you get those emails. This CI run on PR is unrestricted by branch names.

I think this might help:

|Scenario |`push: '**'` |`push: main`|
|---|---|---|
| Push to `main` | Yes | Yes |
| Push to `feature/foo` (no PR) | Yes | No |
| PR into `main` | Yes (twice from `push` + `pull_request`) | Yes (once only) |
| PR into some other branch | Yes (twice) | Yes (once only) |

By the way, I am not married to `push: main` here, but I think it's good practice to avoid running unnecessary tests. Anyways if you both feel strongly about the ** we can go with that!

henrydingliu:
yes, what we are saying is that committing to `feature/foo` is the predominant path that collaborators take. PR into some other branch rarely happens. i think pretty much only I've been using it to move from branches on my person fork to a branch on this repo.

priyam0k:
agreed with henry. let's go with **!

kennethshsu:
K let me close this we can go with #1183

PR #1121, opened 2026-07-14
Friedland Recreation Chapter 8

kennethshsu:
## Summary of Changes  
- Chapter 8 in Friedland in ipynb format
- Added `Population` to `friedland_gl_self_insurer`

## Related GitHub Issue(s)  
#588 

## Additional Context for Reviewers  

- [x] I passed tests locally for both code (`uv run pytest`) and documentation changes (`uv run jb build docs --builder=custom --custom-builder=doctest`)

---

> [!NOTE]
> **Low Risk**
> Documentation and sample-data additions only; no changes to core estimation or API behavior beyond loading an extra column on an existing sample.
> 
> **Overview**
> Adds **Friedland Chapter 8** (Expected Claims Technique) as `docs/friedland/chapter_8.ipynb`, wired into the docs TOC between chapters 7 and 9. The notebook reproduces book exhibits (I–V) with `chainladder` (`ExpectedLoss`, `DevelopmentConstant`, `Chainladder`, `Trend`, parallelogram OLF, etc.) and **assertions** that reconcile outputs to the Friedland PDF pages.
> 
> Extends the **`friedland_gl_self_insurer`** sample with a **`Population`** column in `_manifest.py` and the CSV so Exhibit I Sheet 2 (pure premium / population-based expected claims) can run against packaged data.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1bb99c343ba111dfb12597b3e7e46866cf2d9652. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #1194, opened 2026-08-04
Fixes valuation grain bug

### LINKED ISSUE #1192: Valuation should default to annual when origin grain is annual

### Describe the bug in words

When there's only one valuation (i.e. it's a "1D" triangle - on origin only), the package does not parse the date correctly, and defaults it to month end. See example below.

```python
year_data = [
    [1998,2008,900000,890000],
    [1999,2008,1200000,1170000],
    [2000,2008,1300000,1265000],
    [2001,2008,1800000,1600000],
    [2002,2008,1450000,1200000],
]
year_df = pd.DataFrame(data=year_data,columns=['origin','dev','revenue','expense'])
tri = cl.Triangle(
    data = year_df,
    origin = 'origin',
    development = 'dev',
    columns = 'expense',
    development_format='%Y',
    cumulative = True
)
tri.valuation_date
```

```python
assert pd.to_datetime(tri.valuation_date).date() == pd.Timestamp('2008-12-31').date()
```

- [x] Yes, absolutely!
- [ ] No.

kennethshsu:
## Summary of Changes  
If valuation_date is of length 1 (there's only 1 diagonal), and it contains no month info, it will be defaulted to match origin grain instead

## Related GitHub Issue(s)  
Fixes #1192

Should be pretty easy to review, added 3 new tests.

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes core triangle construction and val/dev conversion logic used across the library; behavior shifts for single-diagonal inputs but is covered by new regression tests.
> 
> **Overview**
> Fixes **valuation / development grain** when every row shares one development date (fixes #1192). Triangles with a single diagonal no longer default to monthly grain when development values are year-only (e.g. `2008` or `%Y`).
> 
> **`Triangle.__init__`** now treats development as year-only when all development column values match `\d{4}`, or when there is only one data row—then **`development_grain` matches `origin_grain`**. If month-level information exists, it still compares the parsed date to the origin period end before choosing monthly vs origin grain.
> 
> **`val_to_dev` / `_val_dev`** always recomputes development-column width from coordinate extents when the array has density, not only when negative development indices need sliding—so single-diagonal triangles keep the correct lag dimension.
> 
> The **`friedland_gl_self_insurer`** sample manifest drops **`development_format: "%Y-12-31`** so loading uses the same inference rules. New tests cover annual vs monthly single-diagonal cases and the friedland sample’s `2008-12-31` valuation date.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bac78c58d32dc33ad8c926d9e670f26204289d07. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/chainladder/core/triangle.py
+++ b/chainladder/core/triangle.py
@@ -482,10 +482,10 @@ class Triangle(TriangleBase):
         )
 
         if len(development_date.unique()) == 1:
-            # checks if development has any non-yearly values
-            dev_has_no_month = not bool(
-                development
-                and not data[development[0]].astype(str).str.fullmatch(r"\d{4}").all()
+            # checks if development is not empty, and if ithas any non-yearly values
+            dev_has_no_month = not development or all(
+                data[col].astype(str).str.fullmatch(r"\d{4}").all()
+                for col in development
             )
 
             if len(data) == 1 or dev_has_no_month:
--- a/chainladder/core/triangle.py
+++ b/chainladder/core/triangle.py
@@ -484,7 +484,11 @@ class Triangle(TriangleBase):
         if len(development_date.unique()) == 1:
             # checks if development is not empty, and if ithas any non-yearly values
             dev_has_no_month = not development or all(
-                data[col].astype("Int64").str.fullmatch(r"\d{4}").all()
+                pd.to_numeric(data[col], errors="coerce")
+                .astype("Int64")
+                .astype(str)
+                .str.fullmatch(r"\d{4}")
+                .all()
                 for col in development
             )
 
--- a/chainladder/core/triangle.py
+++ b/chainladder/core/triangle.py
@@ -484,7 +484,7 @@ class Triangle(TriangleBase):
         if len(development_date.unique()) == 1:
             # checks if development is not empty, and if ithas any non-yearly values
             dev_has_no_month = not development or all(
-                data[col].astype(str).str.fullmatch(r"\d{4}").all()
+                data[col].astype("Int64").str.fullmatch(r"\d{4}").all()
                 for col in development
             )
 
--- a/chainladder/core/triangle.py
+++ b/chainladder/core/triangle.py
@@ -487,10 +487,11 @@ class Triangle(TriangleBase):
                 development
                 and not data[development[0]].astype(str).str.fullmatch(r"\d{4}").all()
             )
-            if (
-                (len(data) == 1 or dev_has_no_month)
-                and self.origin_grain.split("-")[0] in ["Y", "A"]
-            ):
+            # if (
+            #     (len(data) == 1 or dev_has_no_month)
+            #     and self.origin_grain.split("-")[0] in ["Y", "A"]
+            # ):
+            if len(data) == 1 or dev_has_no_month:
                 # if development has no monthly values, match origin
                 self.development_grain = self.origin_grain
             else:
@@ -1475,10 +1476,14 @@ class Triangle(TriangleBase):
         ddims = len(ddims.drop_duplicates())
         if ddims == 1 and sign == -1:
             ddims = len(obj.odims)
-        if obj.values.density > 0 and obj.values.coords[-1].min() < 0:
-            obj.values.coords[-1] = obj.values.coords[-1] - min(
-                obj.values.coords[-1].min(), min_slide
-            )
+        if obj.values.density > 0:
+            if obj.values.coords[-1].min() < 0:
+                obj.values.coords[-1] = obj.values.coords[-1] - min(
+                    obj.values.coords[-1].min(), min_slide
+                )
+            # The lag axis is stepped in development grain while ddims above is
+            # counted in origin periods, so it can be too short to hold the
+            # offset coordinates when the two grains differ.
             ddims = np.max([np.max(obj.values.coords[-1]) + 1, ddims])
         obj.values.shape = tuple(list(obj.shape[:-1]) + [ddims])
         if options.AUTO_SPARSE == False or backend == "cupy":
--- a/chainladder/core/triangle.py
+++ b/chainladder/core/triangle.py
@@ -487,10 +487,7 @@ class Triangle(TriangleBase):
                 development
                 and not data[development[0]].astype(str).str.fullmatch(r"\d{4}").all()
             )
-            # if (
-            #     (len(data) == 1 or dev_has_no_month)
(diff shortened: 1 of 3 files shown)


chainladder/core/triangle.py
```diff
@@ -482,7 +482,14 @@ def __init__(
         )
 
         if len(development_date.unique()) == 1:
-            if len(data) == 1 and self.origin_grain.split("-")[0] in ["Y", "A"]:
+            # checks if development has any non-yearly values
+            dev_has_no_month = not bool(
+                development
+                and not data[development[0]].astype(str).str.fullmatch(r"\d{4}").all()
```
priyam0k: if pandas parses the development column as floats or with missing values (e.g. 2008.0), .astype(str) yields "2008.0".

that will fail fullmatch(r"\d{4}") and default back to 'M'. stripping trailing .0 or converting to int first would make it a bit safer.

chainladder/core/triangle.py
```diff
@@ -482,7 +482,14 @@ def __init__(
         )
 
         if len(development_date.unique()) == 1:
-            if len(data) == 1 and self.origin_grain.split("-")[0] in ["Y", "A"]:
+            # checks if development has any non-yearly values
+            dev_has_no_month = not bool(
```
priyam0k: optional nitpick: the double negative not bool(development and not ...) is a bit hard to follow at first glance. might be clearer as an if/else block, but your call!

kennethshsu: Thanks, will rewrite this, good call here. I'll also improve it to loop over it instead of checking only the first index. Though I don't think this is really necessary

chainladder/core/triangle.py
kennethshsu: Sorry, I don't follow you here, you are saying if the data in the column actually says 2008.0? This is not a valid date string format?

priyam0k: yeah agree it's not valid date formatting!

it's just a pandas edge case where integer year columns with NaNs get cast to float64 (`2008.0`). so `.astype(str)` gives "2008.0", failing `r"\d{4}"` and defaulting grain to `'M'`. super minor edge case though!

kennethshsu: Can you give me an actual example? Either with a constructed pandas DF or CSV? I'm having trouble seeing this

priyam0k: Here's a concrete one. Same annual data as my test, but the dev column comes in as float64 (which is what you often get from a CSV/Excel read, or any time that column lands as float64):

```
import pandas as pd, chainladder as cl
df = pd.DataFrame({
    "origin": [1998, 1999, 2000, 2001, 2002],
    "dev":    [2008.0, 2008.0, 2008.0, 2008.0, 2008.0],  # float64
    "expense":[890000, 1170000, 1265000, 1600000, 1200000],
})
tri = cl.Triangle(df, origin="origin", development="dev",
                  columns="expense", development_format="%Y", cumulative=True)
print(tri.development_grain)   # -> 'M', but with int years it's 'Y'
```
The regex runs on the raw column, and `astype(str)` turns` 2008.0` into `"2008.0"`, which fails `\d{4}`. It then drops into the else branch, and since `%Y` parses to Jan 1 (year start), the month-end vs year-end alignment check fails and it picks M. Casting the column first, e.g. `data[col].astype("Int64").astype(str)`, or stripping a trailing `.0`, fixes it.

chainladder/core/triangle.py
priyam0k: The rewrite reads a lot better, and good call looping over all the development columns. One question: Shall I keep flagging small readability nitpicks, or focus only on functional and correctness issues?

chainladder/core/triangle.py
kennethshsu: That's a very intentional way wrongly coding dates lol, but easy fix, I'll do that, thank you!

chainladder/core/triangle.py
kennethshsu: That's up to you. 

For me, I often rely on AI as little as possible, or even use weaker models on purpose to force me to learn (and because I'm cheap/poor lol). I sense that both these comments have some degree of AI involvement there, and I have no problem with that, but I personally wouldn't use AI to bring these up or catch these myself. I guess my point is if you think it's useful to have these discussions (I think yes to some degree) and you are learning, then yes bring them up. I did learn something through these discussions though. Now should we thrive for perfection and resolve all of bots comments 100%? I think that could be silly and pointless. Just my humble opinion :)

If you read the governing doc, I try to put this in there. We don't need to be perfectionists, if someone runs into a bug on production (the other comment) then yes we should def fix it.

chainladder/core/triangle.py
kennethshsu: Eh something broke, let me put this in draft for now. Will fix later

chainladder/core/triangle.py
priyam0k: thanks kenneth! i checked out the governing doc and completely agree with prioritizing real production bugs over perfectionism on small details.

i do use AI to cross verify ideas while learning, but your tips are super helpful. i'll definitely take your advice to stick to smaller models so i can learn more on my own!

thanks again for sharing your perspective, i'll stick strictly to functional correctness for future reviews!

kennethshsu:
@priyam0k this is ready now, the way that I'm handling it is to cast to int and then string, kinda ugly and not sure if there's a better way. But I added your example as a test too.

By the way on the other commend re AI use. Please don't feel like I don't appreciate the feedback, I actually do, and I learned something, but I leave it up to you to decide on how much you use it and what feedback you want to share from your future work.

priyam0k:
> cast to int and then string, kinda ugly and not sure if there's a better way

nice way when cleaning up float-represented year columns, so nothing ugly. Glad the example was helpful.

> Please don't feel like I don't appreciate the feedback, I actually do, and I learned something,

Never, I took all your note completely as constructive feedback. Since the experience gap is real, I always value hearing your perspective

I'm always looking to try (and even fail) mutiple times, so one day i can finally master it. Here  is one meme I often relate, :)
image

on small nitpicks it is so discretionary to pick up relevant ones, I think only experience can reduce it, ...

kennethshsu:
Sounds good and thanks for the review here!

@henrydingliu @genedan can one of you do another secondary review? Will need this merged to unblock #1121, thanks guys!

henrydingliu (approved):

Issue #1050, opened 2026-06-27
Centralize supported grain values


genedan:
### Description

There are a bunch of places where the grain is looked up using a hardcoded dictionary. For example:

https://github.com/casact/chainladder-python/blob/6f802006f68828fc0126f90c908afc3b1304266d/chainladder/tails/curve.py#L190

There are some clues in the docs that indicate that the supported grain types have changed over time. For example, `A` used to be allowed for yearly grain. For semiannual grain, `S` seems to be the supported value, although there are places where I see `2Q` used, so it's not clear to me what the status is (or what the purpose of `2Q` might be).

Should the supported grain ever be changed again, we'd have to find all the scattered locations of these dictionaries and update them. Storing a unified dict in a central location and then importing it will improve maintainability.

### Is your feature request aligned with the scope of the package?

- [x] Yes, absolutely!
- [ ] No, but it's still worth discussing.
- [ ] N/A (this request is not a codebase enhancement).

### Describe the solution you'd like, or your current workaround.

`{"Y": 12, "S": 6, "Q": 3, "M": 1}` Should be stored somewhere central, maybe in `__init__.py`? Something like:

`__grain_to_months__ = {"Y": 12, "S": 6, "Q": 3, "M": 1}`

But I'm open to better ideas/names.

### Would you be willing to contribute this ticket?

- [ ] Yes, absolutely!
- [ ] Yes, but I would like some help.
- [ ] No.

henrydingliu:
I recall 2Q is a frequency parameter for Pandas.

Issue #524, opened 2024-05-24
[BUG] Parallelogram_olf is buggy for monthly grain


jbogaardt:
**Describe the bug**
There is some special leap year handling that isn't generalized well. I think it is this line:
https://github.com/casact/chainladder-python/blob/master/chainladder/utils/utility_functions.py#L244-L245

This assumes that the format of the of the time dimension is annual, but it can also be monthly.

**To Reproduce**
Steps to reproduce the behavior. Code should be self-contained and runnable against publicly available data. For example:
```python
Python 3.12.2 | packaged by Anaconda, Inc. | (main, Feb 27 2024, 17:28:07) [MSC v.1916 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import chainladder as cl
>>> import pandas as pd
>>> print(cl.__version__)
0.8.22
>>> print(pd.__version__)
2.1.4
>>> rate_history = pd.Series({
...     '2020-01-01': 0.05,
...     '2021-01-15': 0.03,
...     '2022-07-20': 0.01,
...     '2023-01-01': -0.02,})
>>> rate_history.index = pd.DatetimeIndex(rate_history.index)
>>>
>>> cl.parallelogram_olf(
...     values=rate_history.values, date=rate_history.index,
...     start_date = pd.Timestamp('2019-07-01 00:00:00'),
...     end_date=pd.Timestamp('2024-05-31 23:59:59.999999999'),
...     grain='M',
...     vertical_line=False,
...     approximation_grain='M'
...     )
C:\Users\jbogaard\Documents\bitbucket\chainladder-python\chainladder\utils\utility_functions.py:188: FutureWarning: Series.__getitem__ treating keys as positions is deprecated. In a future version, integer keys will always be treated as labels (consistent with DataFrame behavior). To access a value by position, use `ser.iloc[pos]`
  crl = cum_rate_changes[-1]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\jbogaard\Documents\bitbucket\chainladder-python\chainladder\utils\utility_functions.py", line 243, in parallelogram_olf
    combined["is_leap"] = pd.to_datetime(
                          ^^^^^^^^^^^^^^^
  File "C:\Users\jbogaard\AppData\Local\miniconda3\envs\cl_dev\Lib\site-packages\pandas\core\tools\datetimes.py", line 1112, in to_datetime
    values = convert_listlike(arg._values, format)
[...]

PR #1205, opened 2026-08-12
fix(parallelogram): support quarterly and semiannual origin grains

ppcvote:
## Summary of Changes

`ParallelogramOLF` raises `ValueError` on two of the four grains `Triangle.origin_grain` can return, so on-leveling a quarterly or semiannual premium triangle is not possible today:

```python
cl.ParallelogramOLF(rate_history=rates, change_col="RateChange",
                    date_col="EffDate", vertical_line=False).fit(premium_triangle)
```

| origin grain | result on `main` |
| --- | --- |
| `Y` | ok |
| `S` | `ValueError: Invalid frequency: S` |
| `Q` | `ValueError: unconverted data remains when parsing with format "%Y": "Q1"` |
| `M` | ok |

There are two independent causes, both in `parallelogram_olf`.

**The leap-year flag was recovered from the label after it had been stringified.** `utility_functions.py:536` parsed the origin string with a format built as `"%Y" + ("-%m" if grain == "M" else "")`, which only ever describes the `Y` and `M` label shapes, so `"2016Q1"` fails to parse. The flag is now taken off the `PeriodIndex` while it is still a `PeriodIndex`, which drops the dependency on label shape entirely.

**`to_period("S")` does not exist, and `"2Q"` is not a substitute.** pandas reads `"S"` as seconds. `"2Q"` looks like the obvious stand-in and is not one, though not for the reason I first gave here: the multiple *is* honoured, so a 2Q period is six months long. Each one is anchored to the quarter of the observation rather than to a fixed half-year boundary, so consecutive windows overlap (correction courtesy of @henrydingliu below):

```python
>>> idx = pd.date_range("2016-01-01", "2017-12-31", freq="MS")
>>> list(zip(idx[:4].date, idx[:4].to_period("2Q")))
[(2016-01-01, 2016Q1), (2016-02-01, 2016Q1), (2016-03-01, 2016Q1), (2016-04-01, 2016Q2)]
>>> idx.to_period("2Q")[0].start_time, idx.to_period("2Q")[0].end_time
(2016-01-01, 2016-06-30)
>>> idx.to_period("2Q")[3].start_time, idx.to_period("2Q")[3].end_time
(2016-04-01, 2016-09-30)
>>> len(set(idx.to_period("Q"))), len(set(idx.to_period("2Q")))
(8, 8)
```

Two years therefore give eight overlapping windows rather than four half years.

That matters more than it looks. `parallelogram_olf` is broadcast against the triangle positionally (`parallelogram.py:296` takes `.values[None, None]`), so a semiannual result of length 12 would misalign against a 6-origin triangle **silently**, rather than raising. The new `_origin_periods` helper therefore builds half years explicitly, labelled by the quarter they start in, which matches Triangle's own `"%YQ%q"` semiannual format (`triangle.py:855`).

## Related GitHub Issue(s)

Continues #524, which fixed the monthly grain. The leap-year handling was generalised as far as `M` there and stopped, leaving `Q` and `S` uncovered.

Related to #1050 (Centralize supported grain values). This PR is compatible with that refactor rather than a substitute for it: if the `S` / `2Q` mapping is centralised later, `_origin_periods` is the single place this function would consume it from.

**Values, not just absence of a crash.** Results are checked against the aggregation identity used to validate #524, `OLF(period) == 1 / mean(1 / OLF(month))`. It holds to 2.2e-16 for both `S` and `Q`. The `Y` case passes on unfixed code as well and is kept in the suite as a control, so the identity is demonstrated rather than assumed.

**Row counts are asserted, not just values.** Y 3, S 6, Q 12, M 36 for a three-year triangle. The `S == 6` assertion is the one that catches a wrong semiannual bucketing, for the positional-broadcast reason above.

**Nothing changes for `Y` and `M`.** 72 scenarios covering both grains x `approximation_grain` M and D x `vertical_line` both ways x `policy_length` 6/12/24 x windows spanning the 2016 and 2020 leap years, including a 2020-02-29 effective date so the daily leap branch is exercised, hash identically before and after.

**The tests discriminate.** On unmodified code the new tests are 4 failed / 3 passed, the three passes being the `Y` and `M` controls; with the fix, 7 passed. Full suite 1049 passed, 2 skipped.

**About the second commit.** The ruff workflow lints whole changed files, and `utility_functions.py` already carried `F401` (unused `typing.Iterable`) and `E721` on `main` before this branch, so any PR touching this file reports them. I cleared them in a separate commit so it is easy to drop if you would rather keep that out of scope. `is` was used rather than `isinstance` to keep the comparison's exact meaning.

## Checklist
- [x] I passed tests locally for both code (`uv run pytest`) and documentation changes

### DIFF
--- a/chainladder/adjustments/tests/test_parallelogram.py
+++ b/chainladder/adjustments/tests/test_parallelogram.py
@@ -507,3 +507,55 @@ def test_cumulative_duplicate_date_last_wins():
         np.round(olf.to_frame().values.flatten(), 6)
         == [0.67, 0.67, 0.67, 0.67, 0.67, 0.75, 1.0, 1.0, 1.0, 1.0, 1.0]
     )
+
+
+def _olf_for_freq(freq, rates):
+    """Fit ParallelogramOLF against a premium triangle of the given origin grain."""
+    origin = pd.date_range("2016-01-01", "2018-12-31", freq=freq)
+    triangle = cl.Triangle(
+        pd.DataFrame({"origin": origin, "values": 1.0}),
+        origin="origin",
+        columns="values",
+    )
+    olf = cl.ParallelogramOLF(
+        rate_history=rates,
+        change_col="RateChange",
+        date_col="EffDate",
+        vertical_line=False,
+    ).fit(triangle)
+    return triangle.origin_grain, np.asarray(olf.olf_.values).flatten()
+
+
+RATE_HISTORY = pd.DataFrame(
+    {
+        "EffDate": [pd.Timestamp("2016-07-01"), pd.Timestamp("2017-04-01")],
+        "RateChange": [0.08, 0.05],
+    }
+)
+
+
+@pytest.mark.parametrize(
+    "freq, grain, origins", [("YS", "Y", 3), ("2QS", "S", 6), ("QS", "Q", 12), ("MS", "M", 36)]
+)
+def test_olf_supports_every_origin_grain(freq, grain, origins):
+    # Quarterly and semiannual premium triangles used to raise ValueError, so
+    # only two of the four grains Triangle.origin_grain can return were usable.
+    # The row count is asserted because parallelogram_olf is broadcast against
+    # the triangle positionally: a semiannual result of length 12 would align
+    # against the wrong origins rather than fail.
+    observed_grain, olf = _olf_for_freq(freq, RATE_HISTORY)
+    assert observed_grain == grain
+    assert len(olf) == origins
+
+
+@pytest.mark.parametrize("grain, freq, months", [("Y", "YS", 12), ("S", "2QS", 6), ("Q", "QS", 3)])
+def test_olf_aggregates_consistently_with_monthly(grain, freq, months):
+    # An on-level factor is a ratio of rate levels, so a coarser grain is the
+    # reciprocal of the mean reciprocal of the months it spans. The "Y" case
+    # holds on unfixed code too and is kept as a control on the identity.
+    _, monthly = _olf_for_freq("MS", RATE_HISTORY)
+    _, coarse = _olf_for_freq(freq, RATE_HISTORY)
+    expected = np.array(
+        [1 / np.mean(1 / monthly[i : i + months]) for i in range(0, len(monthly), months)]
+    )
+    np.testing.assert_allclose(coarse, expected, rtol=1e-9)
--- a/chainladder/utils/utility_functions.py
+++ b/chainladder/utils/utility_functions.py
@@ -428,11 +428,18 @@ def _origin_periods(index, grain):
     """Bucket a DatetimeIndex into origin periods of the given Triangle grain.
 
     ``DatetimeIndex.to_period`` covers "Y", "Q" and "M" directly. It has no
-    semiannual frequency, and "2Q" does not stand in for one: pandas ignores the
-    multiple when converting to periods, so ``to_period("2Q")`` returns the same
-    four buckets a year that ``to_period("Q")`` does. "S" is therefore built by
-    collapsing each half year onto the quarter it starts in, which also matches
-    how Triangle labels a semiannual origin (``"%YQ%q"``).
+    semiannual frequency, and "2Q" does not stand in for one. The multiple is
+    honoured, so a "2Q" period is six months long, but each one is anchored to
+    the quarter of the observation rather than to a fixed half-year boundary,
+    so the windows overlap::
+
+        2016-01-01 -> 2016Q1 [2016-01-01 .. 2016-06-30]
+        2016-04-01 -> 2016Q2 [2016-04-01 .. 2016-09-30]
+
+    Grouping on that yields eight buckets over two years instead of four, which
+    would then be broadcast against a six-origin triangle. "S" is therefore
+    built by collapsing each half year onto the quarter it starts in, which also
+    matches how Triangle labels a semiannual origin (``"%YQ%q"``).
     """
     if grain == "S":
         quarters = index.to_period("Q")
--- a/chainladder/utils/utility_functions.py
+++ b/chainladder/utils/utility_functions.py
@@ -424,6 +424,27 @@ def read_json(json_str, array_backend=None):
         return cl.__dict__[json_dict["__class__"]]().set_params(**json_dict["params"])
 
 
+def _origin_periods(index, grain):
+    """Bucket a DatetimeIndex into origin periods of the given Triangle grain.
+
+    ``DatetimeIndex.to_period`` covers "Y", "Q" and "M" directly. It has no
+    semiannual frequency, and "2Q" does not stand in for one: pandas ignores the
+    multiple when converting to periods, so ``to_period("2Q")`` returns the same
+    four buckets a year that ``to_period("Q")`` does. "S" is therefore built by
+    collapsing each half year onto the quarter it starts in, which also matches
+    how Triangle labels a semiannual origin (``"%YQ%q"``).
+    """
+    if grain == "S":
+        quarters = index.to_period("Q")
+        return pd.PeriodIndex(
+            [
+                pd.Period(year=q.year, quarter=1 if q.quarter <= 2 else 3, freq="Q")
+                for q in quarters
+            ]
+        )
+    return index.to_period(grain)
+
+
 def parallelogram_olf(
     values,
     dates,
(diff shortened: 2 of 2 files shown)


henrydingliu:
@ppcvote Welcome to the repo and thanks for raising this PR!

ppcvote:
Thanks @henrydingliu, glad to be here.

Worth saying explicitly since you are the one who is likely to check the math: the numbers in this PR are validated with **your own** reciprocal-mean check from #524, the one you used to confirm kennethshsu' Werner and Modlin reconciliation:

```python
monthly_grain["Recip"] = 1 / monthly_grain["OLF"]
1 / monthly_grain[["EY", "Recip"]].groupby(by="EY").mean()
```

I applied it to the semiannual and quarterly grains rather than re-deriving anything, and it holds to 2.2e-16 for both. The annual case is kept in the test suite as a control because it passes on unfixed code too, so the identity is demonstrated rather than assumed.

The one part I would most want a second pair of eyes on is the semiannual bucketing. `to_period("2Q")` looks like the obvious way to express half years and it silently is not one, since pandas drops the multiple:

```python
>>> idx = pd.date_range("2016-01-01", "2017-12-31", freq="MS")
>>> len(set(idx.to_period("Q"))), len(set(idx.to_period("2Q")))
(8, 8)
```

Because `parallelogram_olf` is broadcast positionally at `parallelogram.py:296`, that would hand a 6-origin triangle a 12-row result and misalign it without raising, which is why the tests assert row counts and not just values.

Happy to scope this down to the quarterly crash alone and leave semiannual to #1050 if you would rather keep the `S` labelling convention in one place. The `Q` half is self-contained.

henrydingliu:
> The one part I would most want a second pair of eyes on is the semiannual bucketing. `to_period("2Q")` looks like the obvious way to express half years and it silently is not one, since pandas drops the multiple:
> 
> ```python
> >>> idx = pd.date_range("2016-01-01", "2017-12-31", freq="MS")
> >>> len(set(idx.to_period("Q"))), len(set(idx.to_period("2Q")))
> (8, 8)
> ```

pandas actually picked up the multiple. 

```python
idx.to_period("Q").values[0].end_time == idx.to_period("2Q").values[0].end_time
```
```python
False
```

the issue is that the anchoring doesn't quite behave how we want it to in this context.

ppcvote:
You are right and my explanation was wrong. Thank you for checking it rather than taking my word.

I ran your test and then looked at what the buckets actually are:

```
Q  first period : 2016-01-01 .. 2016-03-31
2Q first period : 2016-01-01 .. 2016-06-30
```

So the multiple **is** honoured, a 2Q period is six months long, and my "pandas drops the multiple" was simply false. The distinct-count evidence I posted was consistent with the wrong explanation because it did not distinguish between "not six months long" and "six months long but overlapping". It is the second:

```
2016-01-01 -> 2016Q1 [2016-01-01 .. 2016-06-30]
2016-02-01 -> 2016Q1 [2016-01-01 .. 2016-06-30]
2016-03-01 -> 2016Q1 [2016-01-01 .. 2016-06-30]
2016-04-01 -> 2016Q2 [2016-04-01 .. 2016-09-30]
2016-05-01 -> 2016Q2 [2016-04-01 .. 2016-09-30]
2016-06-01 -> 2016Q2 [2016-04-01 .. 2016-09-30]
```

Each window is anchored to the quarter of the observation rather than to a fixed half-year boundary, so consecutive windows overlap by a quarter and two years produce eight buckets rather than four. Anchoring, exactly as you said.

The conclusion and the code are unchanged, since eight buckets broadcast against a six-origin triangle either way, but the docstring stated the wrong reason and would have misled the next person. Pushed 879f3df with the corrected wording and your name on it, and I will fix the same sentence in the PR description.

This also strengthens the case for #1050 rather than weakening it: `"S"` -> `"2Q"` is not a wrong-length mapping, it is a correct-length mapping with the wrong anchor, which is a good deal harder to notice at a glance.

henrydingliu:
I'm ready to approve this PR. but open to feedback from the other code owners, as there were no prior issue opened to provide a forum for discussion. 

@ppcvote thanks again for your work! curious how you came across this bug. there wasn't an issue raised but it was a fantastic find. i use chainladder regularly in my day job and i didn't come across this. would love to learn more about your tech stack.

kennethshsu:
No comments from me on this PR.

I do think an improvement is that if we can get various style of dataset ready it could be good. For example, simple triangle, multi-D triangle, with combinations of (M,Q,Y)x(M,Q,Y) combinations all readily available.

kennethshsu:
Looks good, I approved. 

@ppcvote welcome to the project. 

I'm also super curious about your background. It doesn't look like you are a practicing actuary? How did you come across this project and this bug?

kennethshsu (approved):

ppcvote:
@henrydingliu @kennethshsu thank you both, and Kenneth guessed right: not an actuary.

I am a financial adviser in Taiwan, ten years in insurance and financial planning, MDRT. So the domain is familiar from the distribution side rather than the pricing side, which is exactly why I would not have found this by using the library. I have never on-levelled a premium triangle in anger.

What I do run is a small AI products company, and the honest account of how this surfaced is that I was reading the code rather than using it, with Claude Code as the working environment. The specific move that found it was not clever: take `Triangle.origin_grain`, enumerate every value it can return, and check that each one actually survives a round trip through the public estimator. `Y` and `M` did. `S` and `Q` raised. That question is boring to ask and nobody asks it, because the grains that appear in your own data are the ones you exercise, and a library this mature has no obvious reason to fail on two of four.

Which is also the answer to why you had not hit it, @henrydingliu. Annual and monthly cover most real work. The failure needed someone with no dataset at all, generating one grain at a time on purpose.

Two things I would flag rather than leave implied.

The `is_leap_year` route I took reads the flag off the `PeriodIndex` before it is stringified. That is more robust than parsing the label back, but it does mean the leap handling now depends on `PeriodIndex.is_leap_year` staying available. If pandas ever moves that, the failure is silent rather than loud, and a test asserting the 2016/2020 values would catch it. Happy to add one.

@kennethshsu on the fixture idea, yes, and I think it is the more valuable piece of work. What this PR needed was a grain matrix, and I ended up building a throwaway one inside the test. A shared `(M,Q,S,Y) x (M,Q,S,Y)` fixture set, simple and multi-dimensional, would have turned this from a discovery into a routine check, and would give future grain work somewhere to hang. If you would like that as a follow-up PR, say the word and I will take it, with the caveat that you two should sight the expected values since I am reasoning from the library's own aggregation identity rather than from practice.

Thanks for the review. Being corrected on the pandas mechanism mid-thread was the useful part.

henrydingliu:
@ppcvote that's very cool. anything else you noticed while reading the code? or any other suggestions on making the capabilities of the repo more AI-friendly?

ppcvote:
Happy to share, with the caveat that everything below comes from one week of reading, so weigh it accordingly. Five things, each with the evidence that put it on the list.

**1. The grain logic wants one home, which is #1050.** The two sharpest edges I hit were both grain-adjacent: `to_period("2Q")` honouring the multiple but anchoring per-observation (the correction you handed me above), and `_fcrl_for_leap` joining its two frames by row position, which #1219 showed misaligns the moment the leap variant drops one extra boundary-crossing day. say-5's #1221 fixes the instance; genedan's #1050 is the structural fix, and I would happily contribute grain semantics to it once it has a shape.

**2. Type completeness is the highest-leverage "AI-friendly" change, and your own CI already measures it.** The pyright bot on this PR reported 15.2% of exported symbols fully typed (202 / 1331). Coding agents navigate by signatures even more than humans do: an unannotated return means the tool reads the whole function body to learn what comes back, every time, and guesses wrong on `Series | DataFrame` unions. Annotating just the public Triangle methods and `utils/utility_functions.py` returns would move that number and the agent experience more than anything else on this list.

**3. CI lint is stricter than a contributor's local run, silently.** `ruff.yml` runs `--config lint.per-file-ignores={}`, so a tree that passes plain `ruff check` locally can fail CI — that is how the two pre-existing errors in the second commit here surfaced. Relatedly, `[tool.black]` is configured in pyproject while the CI black step is commented out and files have drifted from it. One line in CONTRIBUTING stating the real commands, or aligning the configs, saves every future contributor one confused round trip.

**4. Invariants beat examples as test oracles, and #1218 is the vehicle.** The most useful thing in this PR's review was your reciprocal-mean identity, because it is a property, not a hand-computed expectation: it validated grains nobody had values for. With the fixture factory from #1218, that identity (and cumulative-vs-incremental, and grain-aggregation consistency) could run across the whole (M,Q,S,Y)x(M,Q,S,Y) matrix mechanically. Machine-checkable invariants are precisely what lets an AI-assisted contributor verify their own change without asking a human what the right answer is.

**5. For AI *users* of chainladder, the error messages are the interface.** The two failures that started all this surfaced as `ValueError: unconverted data remains when parsing with format "%Y": "Q1"` and a broadcast-shape mismatch — both point at pandas internals, neither names the concept. An error that says which origin grain was seen and what the estimator supports turns a debugging session into a one-line fix, for humans and agents alike.

None of these are complaints; the review experience here was the best I have had in this campaign, and the identity-as-oracle idea is one I have already reused in another project's PR. If any of the five is worth an issue, say which and I will write it up properly. #1218's scoping questions are still open on my side whenever kennethshsu has a moment.

ppcvote:
@henrydingliu one bit of context since you asked, and then I will stop taking over your PR thread.

One of [Ultra Lab](https://ultralab.tw/en)'s product lines sits in your world: an AI regulation copilot for financial advisers in Taiwan, built on retrieval over insurance policy clauses and tax rulings, and benchmarked against the national insurance-broker licensing exam so we know when it is guessing.

Two lessons from that project came straight into this PR. Domain invariants beat hand-computed expectations as test oracles — your #524 identity is exactly the kind of thing we lean on to let AI-written code verify itself. And error messages are the real interface: an agent, like a junior analyst, can only self-serve on what the error actually names.

Always happy to trade notes — here or async.

Issue #1216, opened 2026-08-14
Initial Linting Cleanup


genedan:
Ruff has now been added as a workflow with the following rules: E2, E4, E7, E9, and F. According to these rules, the project has a large number of existing errors that need to be cleaned up. Because of this, the workflow has been configured to only run on **changed** files in a pull request. Otherwise, PRs will be bogged down with linting errors on unrelated files.

The exclusion of existing errors has been documented in `pyproject.toml` under the header `[tool.ruff.lint.per-file-ignores]`:

https://github.com/casact/chainladder-python/blob/52be42ad1ce7a00df889010e28a90f163c74c3a7/pyproject.toml#L111-L156

These exclusions only apply if a file remains untouched in a pull request. Once someone edits the file, all rules are checked against the entire file, so the author of the PR will be responsible for the whole file's cleanup at that time.

Once the file has been cleaned up, the pull request author should remove its entry from `[tool.ruff.lint.per-file-ignores]`. We will proceed in this manner until `[tool.ruff.lint.per-file-ignores]` no longer has any entries.

Once the cleanup is complete, we will update the workflow to check the entire project on each pull request.

**Adding/Removing Rules** 

[Ruff has a lot of rules](https://docs.astral.sh/ruff/rules/). Because of this, I believe it is impossible to know all of the rules we need to apply ahead of time, but I do expect us to add rules over time as we discover errors that weren't caught by the initial ruleset, which was recommended to me by Claude. For example, E2 was not part of the first attempt. Once we found out that whitespace around operators was not enforced, we added it, which led to the discovery of even more errors in the project base, causing the list of entries in `[tool.ruff.lint.per-file-ignores]` to expand.

Likewise, we may find certain rules to be too strict and remove them. It will be a judgment call as to how broad we apply new rules - too broad, and we may find the linter to be overly strict, but too narrow, and we might not catch enough errors.

When I added E2, that covers all the rules in the E200s, for example, [E251](https://docs.astral.sh/ruff/rules/unexpected-spaces-around-keyword-parameter-equals/), [E252](https://docs.astral.sh/ruff/rules/missing-whitespace-around-parameter-equals/), are both covered by simply specifying "E2". I believe that we need almost all the rules in E2, so I added it broadly. But if I am wrong on this, we need to be more granular on the rules.

**Voting**

[...]

PR #1229, opened 2026-08-26
Allow development to accept an age in months relative to origin

### LINKED ISSUE #1094: Hardcoded 1970 value in core/base.py
There's a hardcoded `1970` value in `core/base.py`:

https://github.com/casact/chainladder-python/blob/6c9ce99650f627adf5fde18690fad3e8a912a8dd/chainladder/core/base.py#L142-L145

I'm not sure what this is for, the meaning should become clear after this file goes through the annotation process. In any case, we've been moving values like this to the package `__init__.py` (from which I'd propose a distinct options module be spun-off eventually) to centralize the constants.

Abhayindia:
Closes #1094 (the enhancement route agreed with @henrydingliu in that thread).

`_to_datetime` tries `%Y%m`, `%Y`, `%Y-%m-%d` and then falls through to a bare `pd.to_datetime`, which for a numeric column silently reads the values as nanoseconds since epoch. That collapsed age-lag columns like `[12, 24, 36]` to 1970-01-01, and the old guard just checked whether the result landed in 1970 - so a triangle whose development really is the year 1970 got the same error.

This replaces that with a check at the point of parsing: if none of the explicit formats matched and the column is numeric, it's an age in months. `_init_development` then resolves each value to the valuation date that many months after the start of that row's origin period. Ages are measured from the origin *period* start (using the constructor's own `origin_grain`, so mid-period origin dates and fiscal-year anchors line up), not the literal recorded origin date. Bare years still match `%Y` and are untouched. Non-numeric unparseable input still raises, with the same message as before. Semiannual grains only support the calendar Jan/Jul anchor; a non-calendar one raises a clear error rather than guessing.

Verified:

```
$ python -m pytest chainladder/core/tests/test_triangle.py -q
312 passed, 1 skipped in 100.59s
```

Full suite was 1091 passed / 8 skipped when I posted the branch on the issue.

### DIFF
--- a/chainladder/core/base.py
+++ b/chainladder/core/base.py
@@ -175,16 +175,38 @@ class TriangleBase(
             data: DataFrame,
             development: list,
             development_format: None | str,
-            origin_date: Series
+            origin_date: Series,
+            origin_grain: str
     ) -> Series:
         """Initialize development and its grain"""
         if development:
-            development_date: Series = TriangleBase._to_datetime(
+            development_date: Series | None = TriangleBase._to_datetime(
                 data=data,
                 fields=development,
                 period_end=True,
-                date_format=development_format
+                date_format=development_format,
+                allow_age=True,
             )
+            if development_date is None:
+                # age in months relative to origin's period start, using the
+                # constructor's own origin_grain so fiscal-year anchors match
+                age: Series = pd.to_numeric(data[development[0]]).round().astype(int)
+                grain_base = origin_grain.split("-")[0]
+                if grain_base == "2Q":
+                    # no native pandas semiannual period; only calendar Jan/Jul anchors supported
+                    if origin_grain not in ("2Q", "2Q-DEC"):
+                        raise ValueError(
+                            "Development expressed as an age is not yet supported for a "
+                            f"non-calendar semiannual origin grain ({origin_grain})."
+                        )
+                    origin_period_start: Series = origin_date.apply(
+                        lambda d: d.replace(month=((d.month - 1) // 6) * 6 + 1, day=1)
+                    )
+                else:
+                    origin_period_start = origin_date.dt.to_period(origin_grain).dt.to_timestamp(how="s")
+                development_date = (
+                    origin_period_start.dt.to_period("M") + (age - 1)
+                ).dt.to_timestamp(how="e")
         else:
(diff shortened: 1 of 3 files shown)


henrydingliu:
just unblocked tests. expecting ruff and codecov fails.

henrydingliu:
moving to draft due to conflicts

Abhayindia:
@henrydingliu  the conflict is genedan's #1088, which added `test_set_development_age_instead_of_date_raises` - it asserts development-as-age raises ValueError, the opposite of what this PR does. @genedan do you still want the age-in-months feature from #1094, or has the direction shifted to "reject ages with a clear error"? happy to rework either way.

henrydingliu:
good to deprecate test_set_development_age_instead_of_date_raises

Abhayindia:
rebased on main - kept the age-in-months tests and dropped test_set_development_age_instead_of_date_raises per your call. `ruff check` clean; the `ruff format` fail is the same pre-existing whole-file base.py/test_triangle.py nonconformance.

henrydingliu:
> the ruff format fail is the same pre-existing whole-file base.py/test_triangle.py nonconformance

is it a bug that the per-file ignore didn't work?

codecov still shows uncovered changes

Abhayindia:
per-file-ignores only apply to `ruff check`, not `ruff format` - the formatter has no per-file exclusion, so any PR that touches base.py trips the whole-file format check. base.py needs ~200 lines of reformatting, which really belongs in its own cleanup PR (same situation #1216 describes). I can run `ruff format` on test_triangle.py here (~60 lines) if that helps, but base.py I'd rather leave.

codecov: the uncovered lines are the semiannual-origin (`2Q`) branch of the age path - it only triggers for a non-default semiannual anchor and is awkward to exercise in a unit test. Happy to add one if you want it covered.

henrydingliu:
understood on ruff format.

>it only triggers for a non-default semiannual anchor and is awkward to exercise in a unit test. 

does the new prism triangles help?

Abhayindia:
the prism triangles are all date-based so they don't hit the age path. added a direct test instead - semiannual origins (2017H1/H2, 2018H1) with development given as ages, covers the 2Q anchor branch. passes locally.

henrydingliu (approved):
only a single error handling line is uncovered.

PR #1230, opened 2026-08-27
Fix ParallelogramOLF daily-grain ValueError on triangles starting after a leap year

### LINKED ISSUE #1219: ParallelogramOLF daily grain ValueError after leap year

### Describe the bug in words

while going through the pr #1205 findings , i was trying a daily approximation test case and ran into a broadcast error. when approximation_grain="D" is used on a triangle starting right after a leap year (like 2017-01-01 following 2016), .fit() crashes.

the non-leap and leap dataframes in _fcrl_for_leap drop different numbers of days (366 vs 367), so they end up with different row counts. combining them with positional .join() shifts the origin periods by one row, causing:` ValueError: operands could not be broadcast together with shapes (1,1,3,1) (1,1,4,1)`

```python
import chainladder as cl
import pandas as pd
rates = pd.DataFrame({'EffDate': [pd.Timestamp('2016-07-01')], 'RateChange': [0.05]})
origin = pd.date_range('2017-01-01', '2019-12-31', freq='YS')
triangle = cl.Triangle(pd.DataFrame({'origin': origin, 'values': 1.0}), origin='origin', columns='values')
# raises ValueError: operands could not be broadcast together
olf = cl.ParallelogramOLF(
    rate_history=rates, change_col='RateChange', date_col='EffDate', approximation_grain='D'
).fit(triangle)

```

```python
olf = cl.ParallelogramOLF(
    rate_history=rates, change_col='RateChange', date_col='EffDate', approximation_grain='D'
).fit(triangle)

assert len(olf.olf_.origin) == len(triangle.origin)
assert not olf.olf_.to_frame().isna().any().any()
```

- [ ] Yes, absolutely!
- [x] Yes, but I would like some help.
- [ ] No.

Abhayindia:
Closes #1219.

With `approximation_grain="D"`, `parallelogram_olf` trimmed the lookback window from the front of the daily grid by a fixed row count (`366 * lookback_years`). The leap pass trims one extra row than the non-leap pass, and when that row is the only day of the first origin period the two passes group into a different number of origins - 3 vs 4 on the repro. The result is then broadcast against the triangle positionally, so the mismatch surfaces as `ValueError: operands could not be broadcast together with shapes (1,1,3,1) (1,1,4,1)` inside `.fit()`.

It also meant the trim landed a day or so away from the real window start, so a partial origin period (a single 2016-12-31 day here) could leak in even when the counts happened to match.

The fix trims by date instead: both passes keep everything from the actual window start (`origin[0]` for the ParallelogramOLF path, Jan 1 of the earliest rate year otherwise), so they always span the same origins. The two passes are then joined on the origin period rather than on row position, and the leap flag is filled from whichever side is present, so a future divergence in either pass can't silently mis-pair origins again. Monthly and non-leap daily results are unchanged apart from at most a one-day shift at the leading origin from trimming at the true boundary.

```
$ python -m pytest chainladder/adjustments/tests/test_parallelogram.py -q
20 passed

$ python -m pytest chainladder/adjustments/ chainladder/utils/tests/ -q
140 passed, 1 skipped
```

The daily-after-leap case from the issue is added as a regression test.

### DIFF
--- a/chainladder/adjustments/tests/test_parallelogram.py
+++ b/chainladder/adjustments/tests/test_parallelogram.py
@@ -559,3 +559,30 @@ def test_olf_aggregates_consistently_with_monthly(grain, freq, months):
         [1 / np.mean(1 / monthly[i : i + months]) for i in range(0, len(monthly), months)]
     )
     np.testing.assert_allclose(coarse, expected, rtol=1e-9)
+
+
+def test_daily_grain_after_leap_year():
+    # A daily approximation on a triangle whose origins start right after a leap
+    # year used to trim the lookback window by row count, which dropped one more
+    # row in the leap pass than the non-leap pass. When that row was the only one
+    # of its origin period the two passes produced a different number of origins
+    # and .fit() failed with a broadcast error.
+    rates = pd.DataFrame({
+        "EffDate": [pd.Timestamp("2016-07-01")],
+        "RateChange": [0.05],
+    })
+    origin = pd.date_range("2017-01-01", "2019-12-31", freq="YS")
+    triangle = cl.Triangle(
+        pd.DataFrame({"origin": origin, "values": 1.0}),
+        origin="origin",
+        columns="values",
+        cumulative=True,
+    )
+    olf = cl.ParallelogramOLF(
+        rate_history=rates,
+        change_col="RateChange",
+        date_col="EffDate",
+        approximation_grain="D",
+    ).fit(triangle)
+    assert len(olf.olf_.origin) == len(triangle.origin)
+    assert not olf.olf_.to_frame(origin_as_datetime=False).isna().any().any()
--- a/chainladder/adjustments/tests/test_parallelogram.py
+++ b/chainladder/adjustments/tests/test_parallelogram.py
@@ -47,7 +47,8 @@ def test_non_vertical_line():
     )
 
     result = (
-        cl.parallelogram_olf([0.20], ["7/1/2017"], approximation_grain="D")
+        cl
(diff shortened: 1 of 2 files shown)


henrydingliu:
unblocked tests

priyam0k:
Thanks @Abhayindia for picking this up and submitting this PR.

Just curious, how did you come across this issue? Is your background in actuarial?

priyam0k (commented):
Your solution is correct and all checks passed. Just a small note: we have a PR [template](https://github.com/casact/chainladder-python/blob/main/.github/pull_request_template.md) which we encourage to follow for PRs.

Abhayindia:
> Thanks @Abhayindia for picking this up and submitting this PR.
> 
> 
> 
> Just curious, how did you come across this issue? Is your background in actuarial?

@priyam0k by profession I am a robotics engineer but I take interest in finance which has led to me becoming a student of Institute of Actuaries of India. I am en route to become an Associate Member soon.

Abhayindia:
> Your solution is correct and all checks passed. Just a small note: we have a PR [template](https://github.com/casact/chainladder-python/blob/main/.github/pull_request_template.md) which we encourage to follow for PRs.

Will follow from next time.

priyam0k:
> robotics engineer but I take interest in finance which has led to me becoming a student of Institute of Actuaries of India. I am en route to become an Associate Member soon.

Great combinations. Looking forward to more of ur contribution. Approving this now , leaving final signoff to core reviewers.

priyam0k (approved):

henrydingliu (approved):
no additional issues

PR #1243, opened 2026-08-29
Add Ruff rules B018 and UP034

### LINKED ISSUE #1235: Add Ruff rule: useless-expression (B018)
What this [rule](https://docs.astral.sh/ruff/rules/useless-expression/) does is flag statements that have no effect, for example:

```python
def my_function(x: int) -> int:
    1 + 1
    return x + 5
```

`1 + 1` would have no effect, and Ruff should flag it so that it can be removed.

### LINKED ISSUE #1236: Add Ruff rule: extraneous-parentheses (UP034)
This Ruff [rule](https://docs.astral.sh/ruff/rules/extraneous-parentheses/) will flag extraneous parentheses, i.e.,:

```python
print((((("Hello world!)))))
```

Abhayindia:
## Summary of Changes

Bundles B018 and UP034 since the cleanup overlaps and you mentioned several rule issues can go in one PR.

B018 (useless-expression) turned up a real bug: `MunichAdjustment.transform` builds the ldf triangle but had `obj._set_slicers` without the call, so the returned triangle never rebuilt its `.iloc`/`.loc`/`.at` indexers after the shape change. The rest was test-side - a couple of dead lines, some smoke-check expressions bound to `_`, and the empty-triangle check in `TriangleDisplay._dimensionality` switched to `hasattr` instead of catching `AttributeError` on a bare `self.values`.

UP034 (extraneous-parentheses) was a few redundant paren pairs in `clark.py`, `benktander.py` and a couple of tests/notebooks.

Since touching a file drops its per-file-ignore protection, I cleaned each file I touched to the full ruleset and removed its entry: `clark.py` (also `type(x) ==` -> `isinstance` and a stray assigned lambda -> `def`), `display.py`, `test_bootstrap.py`, `test_munich.py`. `test_development.py`, `incremental.py`, `test_voting.py` and two notebooks still trip one of the new rules but are big enough that a full cleanup belongs in its own PR, so they're grandfathered.

## Related GitHub Issue(s)

Closes #1235
Closes #1236

ruff check + format with both rules is clean over the project apart from the grandfathered files. `pytest` passes for every touched module locally (234 tests). `ruff format` didn't touch any `.. testcode::` block so the doctest build is unaffected.


henrydingliu:
looks good. please resolve conflict and i can approve

Abhayindia:
rebased on main - the only conflict was in pyproject.toml (main dropped the base.py per-file-ignore, this PR drops clark.py's), resolved to drop both. ruff check + format clean on the changed files.

henrydingliu (approved):

PR #1245, opened 2026-08-30
[BUG] Respect dropped link ratios when applying n_periods

### LINKED ISSUE #604: 'n_periods' does not respect dropped link ratios

## Description
Currently, when using the `Development` estimator, the `n_periods` parameter is applied *before* any exclusions from `drop`, `drop_valuation`, etc., are considered. This leads to unexpected results when attempting to exclude specific origin-development combinations from LDF calculations.

## Current Behavior
In the example below when `n_periods=1` is specified along with `drop=[("2009", 12)]`, the system:
1. Identifies the most recent period (2009)
2. Applies the drop, resulting in no valid periods for the calculation
3. Returns NaN or uses an incorrect period

## Expected Behavior
When `n_periods=1` is specified along with `drop=[("2009", 12)]`, the system should:
1. Identify which periods are valid (non-dropped)
2. Select the most recent valid period (skipping 2009)
3. Use that period (e.g., 2008) for the LDF calculation

```python
import chainladder as cl
import pandas as pd

# create example triangle
data = {
    'origin': ["2007-01-01", "2007-01-01", "2007-01-01", "2007-01-01", "2008-01-01", "2008-01-01", "2008-01-01", "2009-01-01", "2009-01-01", "2010-01-01"],
    'development': ["2007-01-01", "2008-01-01", "2009-01-01", "2010-01-01", "2008-01-01", "2009-01-01", "2010-01-01", "2009-01-01", "2010-01-01", "2010-01-01"],
    'loss': [100, 200, 300, 400, 150, 300, 450, 200, 250, 50]
}

df = pd.DataFrame(data)
tri = cl.Triangle(
    df, 
    origin='origin',
    development='development',
    columns='loss',
    cumulative=True
)

# calculate ldf with n_periods=1 and the most recent period (2009) dropped
dev = cl.Development(n_periods=1, drop=[('2009', 12)]).fit(tri)
dev.ldf_

# current behaviour : uses only the excluded ratio from 2009 in dev month 12 resulting in NaN
# desired behaviour: use valid ratios (that haven't been dropped) resulting in the ratio from 2008 in dev month 12 and so expected result is 300 / 150 = 2

```

The `n_periods` parameter should be applied *after* all drop logic. It should select the `n` most recent link ratios from the set of all valid, non-dropped, non-NaN data points. This ensures that the `n_periods` argument is always honored when enough valid data is available.

son1403:
Fixes #604

## Summary

n_periods now selects the most recent valid link ratios after applying drop and drop_valuation.

Previously, dropping the latest ratio could leave the selected period empty and result in a NaN LDF.

## Changes

- Apply explicit exclusions before selecting periods.
- Skip non-finite link ratios when choosing the fallback period.
- Add regression coverage for both cases.

### DIFF
--- a/chainladder/development/base.py
+++ b/chainladder/development/base.py
@@ -82,11 +82,31 @@ class DevelopmentBase(
             .to_frame()
         )
 
-    def _assign_n_periods_weight(self, X, n_periods):
-        """Used to apply the n_periods weight"""
+    def _assign_n_periods_weight(self, X, n_periods, candidate_weight=None):
+        """Used to apply the n_periods weight.
 
-        def _assign_n_periods_weight_int(X, n_periods):
+        When ``candidate_weight`` is supplied, it represents factors that have
+        already been excluded (for example through ``drop`` or
+        ``drop_valuation``).  ``n_periods`` is then applied to the most recent
+        remaining factors rather than to a fixed calendar window.
+        """
+
+        def _assign_n_periods_weight_int(X, n_periods, candidate_weight):
             xp = X.get_array_module()
+            if candidate_weight is not None:
+                candidate = (
+                    candidate_weight[None, None]
+                    if candidate_weight.ndim == 2
+                    else candidate_weight
+                )
+                valid = candidate > 0
+                if n_periods < 1 or n_periods >= X.shape[-2] - 1:
+                    return candidate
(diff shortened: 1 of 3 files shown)


henrydingliu:
@son1403 welcome to the repo. did you get a chance to read the entire discussion on the issue?

son1403:
ah sorry for that, I've just jumped in here

PR #1262, opened 2026-09-02
fix(correlation): keep empty diagonals when building z_critical (#320)

### LINKED ISSUE #320: Exception raised with .valuation_correlation().z_critical
Hello,

I received the following traceback when trying to run the .valuation_correlation method:

```
Traceback (most recent call last):
  File "/usr/lib/python3.10/code.py", line 90, in runcode
    exec(code, self.locals)
  File "<input>", line 1, in <module>
  File "/home/ubuntu/chainladder-python/chainladder/core/display.py", line 18, in __repr__
    data = self._repr_format()
  File "/home/ubuntu/chainladder-python/chainladder/core/display.py", line 87, in _repr_format
    return pd.DataFrame(out, index=origin, columns=development)
  File "/home/ubuntu/chainladder-python/venv/lib/python3.10/site-packages/pandas/core/frame.py", line 694, in __init__
    mgr = ndarray_to_mgr(
  File "/home/ubuntu/chainladder-python/venv/lib/python3.10/site-packages/pandas/core/internals/construction.py", line 351, in ndarray_to_mgr
    _check_values_indices_shape_match(values, index, columns)
  File "/home/ubuntu/chainladder-python/venv/lib/python3.10/site-packages/pandas/core/internals/construction.py", line 422, in _check_values_indices_shape_match
    raise ValueError(f"Shape of passed values is {passed}, indices imply {implied}")
ValueError: Shape of passed values is (1, 10), indices imply (1, 9)

```

The code used to trigger the exception is below, although I did have to execute the last line twice to see the output in the console:

```
import chainladder as cl
import pandas as pd

df_xyz = pd.read_csv('friedland_xyz_auto_bi.csv')

tri_xyz = cl.Triangle(
    data=df_xyz,
    origin='Accident Year',
    development='Calendar Year',
    columns=['Paid Claims'],
    cumulative=True
)

res = tri_xyz.valuation_correlation(p_critical=0.1, total=False).z_critical
```

The data are from the XYZ example in the Friedland paper, a csv of which can be found here:

https://github.com/casact/FASLR/blob/main/faslr/samples/friedland_xyz_auto_bi.csv

Notably, this is an incomplete triangle as the first two diagonals are missing, which I think may be the reason for the exception:

```
         12       24       36       48       60       72       84       96       108      120      132
1998     NaN      NaN   6309.0   8521.0  10082.0  11620.0  13242.0  14419.0  15311.0  15764.0  15822.0
1999     NaN   4666.0   9861.0  13971.0  18127.0  22032.0  23511.0  24146.0  24592.0  24817.0      NaN
2000  1302.0   6513.0  12139.0  17828.0  24030.0  28853.0  33222.0  35902.0  36782.0      NaN      NaN
2001  1539.0   5952.0  12319.0  18609.0  24387.0  31090.0  37070.0  38519.0      NaN      NaN      NaN
2002  2318.0   7932.0  13822.0  22095.0  31945.0  40629.0  44437.0      NaN      NaN      NaN      NaN
2003  1743.0   6240.0  12683.0  22892.0  34505.0  39320.0      NaN      NaN      NaN      NaN      NaN
2004  2221.0   9898.0  25950.0  43439.0  52811.0      NaN      NaN      NaN      NaN      NaN      NaN
2005  3043.0  12219.0  27073.0  40026.0      NaN      NaN      NaN      NaN      NaN      NaN      NaN
2006  3531.0  11778.0  22819.0      NaN      NaN      NaN      NaN      NaN      NaN      NaN      NaN
2007  3529.0  11865.0      NaN      NaN      NaN      NaN      NaN      NaN      NaN      NaN      NaN
2008  3409.0      NaN      NaN      NaN      NaN      NaN      NaN      NaN      NaN      NaN      NaN
```

I'm wondering, what should the output be in this case? I'm not too familiar with the details of the method, so I'm not sure if we should expect the whole thing to fail, or just have NaNs for the years or years containing missing data points. If it's the former, maybe we could have the error message state that the test cannot be done with missing diagonals or if the latter, output for the years which can be computed and a warning saying that not all years can be computed due to missing data.

Thanks,

Gene

abdulsamie10:
## Summary of Changes

Fixes the `ValueError` from #320: `valuation_correlation(total=False).z_critical` crashed
on any triangle whose earliest diagonals are entirely missing (e.g. `load_sample('xyz')`,
the Friedland XYZ insurer data from the issue).

`self.probs` always has one column per link-ratio diagonal, but the `z_critical` frame it
is written into was built with `.dropna()`, which also drops interior all-NaN diagonals.
On triangles with missing leading diagonals the frame ended up with fewer ddims than value
columns, and repr (or anything else that walks ddims) raised
`ValueError: Shape of passed values is (1, 10), indices imply (1, 9)`.

The fix builds the frame by valuation slicing instead, so empty diagonals keep their
column and stay aligned with `probs`:

- Complete triangles produce identical output to before (checked ddims/odims/values
  against the old construction on raa, clrd, quarterly and genins).
- Incomplete triangles now return a well-formed boolean triangle; diagonals with no data
  test non-significant (False). If you'd rather surface those as NaN it would need a dtype
  change away from bool — happy to discuss, this PR keeps the minimal behavior.
- Removed the stale commented-out construction above these lines.

Because the Ruff workflow lints changed files with per-file-ignores cleared, this also
fixes the pre-existing E741 in `DevelopmentCorrelation` (`I` → `num_dev_periods`, pure
rename), removes the corresponding per-file-ignore from `pyproject.toml`, and applies
`ruff format` (0.16.1) to the two touched files.

Added `test_val_corr_incomplete_triangle` using the existing `xyz` fixture, which runs on
both the numpy and sparse backends. It fails on main and passes with this change.

## Related GitHub Issue(s)

#320

The one behavioral question worth a reviewer's eye: for diagonals with no data,
`z_critical` now reports False (not significant) rather than being silently dropped from
the output. That matches the "partial results" option discussed in the issue, and keeps
the column count consistent with `z`, `z_expectation` and `z_variance`.

Local runs: `uv run pytest` → 1129 passed, 8 skipped. `ruff check` (with per-file-ignores
cleared) and `ruff format --check` are clean on the changed files with ruff 0.16.1.

### DIFF
--- a/chainladder/core/correlation.py
+++ b/chainladder/core/correlation.py
@@ -140,20 +140,23 @@ class DevelopmentCorrelation:
         numerator.values = numerator.values[..., :-1]
         numerator.ddims = numerator.ddims[:-1]
 
-        # I is the number of development periods in the triangle
-        I = len(triangle.development)
+        # num_dev_periods is the number of development periods in the
+        # triangle, denoted I in the Mack 97 paper
+        num_dev_periods = len(triangle.development)
 
         # k values are the column indexes for which we are calculating T_k
         k = xp.array(range(2, 2 + numerator.shape[3]))
 
         # denominator is the one in formula G4 of the Mack 97 paper
-        denominator = ((I - k) ** 3 - I + k)[None, None, None]
+        denominator = ((num_dev_periods - k) ** 3 - num_dev_periods + k)[
+            None, None, None
+        ]
 
         # complete formula G4, results in array of each T_k value
         self.t = 1 - 6 * xp.nan_to_num(numerator.values) / denominator
 
         # per Mack, weight is one less than the number of pairs for each T_k
-        weight = (I - k - 1)[None, None, None]
+        weight = (num_dev_periods - k - 1)[None, None, None]
 
         # Calculate big T, the weighted average of the T_k values
         t_expectation = (
@@ -163,7 +166,7 @@ class DevelopmentCorrelation:
         idx = triangle.index.set_index(triangle.key_labels).index
 
         # variance is result of formula G6
-        self.t_variance = 2 / ((I - 2) * (I - 3))
+        self.t_variance = 2 / ((num_dev_periods - 2) * (num_dev_periods - 3))
 
         # array of t values
         self.t = pd.DataFrame(self.t[0, 0, ...], columns=k, index=["T_k"])
@@ -317,19 +320,18 @@ class ValuationCorrelation:
         if not self.total:
             T = []
             for i in range(0, xp.max(m1large.shape[2:]) + 1):
-                T.append(
-                    [
-                        pZlower(i, j, 0.5)
-                        for j in range(0, xp.max(m1large.shape[2:]) + 1)
-                    ]
-                )
+                T.append([
+                    pZlower(i, j, 0.5) for j in range(0, xp.max(m1large.shape[2:]) + 1)
+                ])
             T = np.array(T)
             z_idx, n_idx = z.astype(int), n.astype(int)
             self.probs = T[z_idx, n_idx]
-            z_critical = triangle[triangle.valuation > triangle.valuation.min()]
-            # z_critical = z_critical[z_critical.development > z_critical.development.min()].dev_to_val().sum(
-            #     "origin") * 0
-            z_critical = z_critical.dev_to_val().dropna().sum("origin") * 0
+            # One column per link-ratio diagonal, labeled by ending valuation.
+            # Slicing by valuation (rather than dropna) keeps diagonals that
+            # are entirely NaN, so the columns stay aligned with self.probs
+            # when a triangle is missing its earliest diagonals (#320).
+            z_critical = triangle.dev_to_val().sum("origin")
+            z_critical = z_critical[z_critical.valuation > triangle.valuation.min()] * 0
             z_critical.values = np.array(self.probs) < p_critical
             z_critical.odims = triangle.odims[0:1]
             self.z_critical = z_critical
--- a/chainladder/core/tests/test_correlation.py
+++ b/chainladder/core/tests/test_correlation.py
@@ -3,18 +3,34 @@ import pytest
 
 raa = cl.load_sample("RAA")
 
+
 def test_val_corr_total_true():
     assert raa.valuation_correlation(p_critical=0.5, total=True)
+
 
 def test_val_corr_total_false():
     assert raa.valuation_correlation(p_critical=0.5, total=False)
 
+
 def test_dev_corr():
     assert raa.development_correlation(p_critical=0.5)
 
+
 def test_dev_corr_sparse():
-    assert raa.set_backend('sparse').development_correlation(p_critical=0.5)
+    assert raa.set_backend("sparse").development_correlation(p_critical=0.5)
+
 
 def test_validate_critical():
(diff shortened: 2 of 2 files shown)


henrydingliu:
@abdulsamie10 welcome! thanks for picking up this old bug. looks to be a single line fix?

curious how you came upon our repo and what your interest in actuarial science is

henrydingliu:
moving to draft due to conflict

abdulsamie10:
Rebased onto `main` — conflict resolved, back in review.

The conflict was in `pyproject.toml`: #1278 / #1279 added `N802` to
`"chainladder/core/correlation.py" = ["E741", "N802"]`, and this branch had removed that line.

I've dropped my `pyproject.toml` change entirely and kept yours as-is. The lint edit was scope
creep on a bug fix — with `E741` still ignored for the file it buys nothing, and it was the only
thing that could conflict with your ruff work. The diff is now just the fix and its test:

```
chainladder/core/correlation.py            | 32 ++++++++++++++++--------------
chainladder/core/tests/test_correlation.py | 20 +++++++++++++++++--
2 files changed, 35 insertions(+), 17 deletions(-)
```

`I` → `num_dev_periods` is still in there for readability, but the existing `E741` ignore is
left untouched, so nothing about your lint config has to move for this to land. Happy to drop
the rename too if you'd rather the diff were only the `z_critical` slice.

Re-verified after the rebase: `test_correlation.py` 7 passed, and the reported case is aligned
again — `z_critical.values.shape[-1] == len(z_critical.ddims) == 10`, `repr()` no longer raises.

henrydingliu:
apologies for the run-around. i prioritized #1287 as it's a more comprehensive ruff fix. would you mind rebasing main and re-adding the bug fix once #1287 is merged?

henrydingliu:
turning to draft due to conflict

henrydingliu:
@abdulsamie10 hi there, are you interested in updating the PR? if not, do you mind if we reuse your code in another PR?

abdulsamie10:
Yes — still interested, and updated. Rebased onto `main` (`337a163c`), conflict resolved, back in review.

The conflict was the `I` rename. #1287 renamed it to `n_dev_periods`; this branch had renamed it
to `num_dev_periods`. I've dropped mine entirely and kept yours, so the diff is now only the
`z_critical` fix and its regression test:

```
chainladder/core/correlation.py            | 12 ++++++++----
chainladder/core/tests/test_correlation.py | 10 ++++++++++
2 files changed, 18 insertions(+), 4 deletions(-)
```

Nothing outside `correlation.py` is touched now — no `pyproject.toml`, no lint config — so it
should stop colliding with the ruff work.

Verified after the rebase:

- `test_correlation.py` — 10 passed.
- The new test discriminates the bug: reverting just `correlation.py` to `main` while keeping the
  test gives `2 failed` (`normal_run` and `sparse_only_run`); restoring the fix gives `2 passed`.

And thank you for offering to carry it — but no need, it's current now.

henrydingliu:
I'm going to approve and address the ruff issue in another PR. 

also changed the PR description to not close #320. even though the error is fixed, there are other considerations such as warning emission under discussion in the issue.

henrydingliu (approved):

henrydingliu:
@abdulsamie10 thanks for all your work! welcome to the repo! let me know what interests you and i can recommend some other issues!

Issue #1274, opened 2026-09-03
CapeCod.predict infers the fitted grain from key_labels: should that inference exist?


ppcvote:
Split out of #1265 at @henrydingliu's request. No bug here, this is the design question that issue kept running into.

There are two ways to say "fit by line of business, apply per company":

```python
# A: the parameter
cl.CapeCod(groupby="LOB").fit(tri, sample_weight=prem)

# B: aggregate the training data and let predict() work out what happened
model = cl.CapeCod().fit(
    tri.groupby("LOB").sum(), sample_weight=prem.groupby("LOB").sum())
model.predict(tri, sample_weight=prem)
```

B works by diffing `key_labels` between `sample_weight` and `apriori_` and regrouping if they differ (`capecod.py:325`). That inference is where the #1265 bug lived.

The two routes agree, once development is grouped consistently (`clrd`, comauto):

| route | apriori |
| --- | --- |
| A, `Development()` ungrouped | 0.5602627308 |
| A, `Development(groupby="LOB")` | 0.5689995797 |
| B | 0.5689995797 |

One correction to what I wrote in #1265: I presented A and B as giving different aprioris. The difference is entirely the development grain, not the CapeCod grain.

So: with `groupby` available and giving the identical answer, does `predict()` need to infer the grain at all? As I see it:

1. Leave it. It is the behaviour #400 asked for and it works once #1265 is fixed.
2. Keep it, document that A and B are equivalent, and warn when the inference fires.
3. Deprecate the inference in favour of `groupby`.

I lean 2. The inference is convenient, but it currently happens invisibly: a user who calls `predict()` on a 775-company triangle gets a 6-row `apriori_` back and nothing says why. A warning turns a silent regrouping into a visible one.

I can implement whichever you prefer.

henrydingliu:
thanks for raising! i'm leaning towards 2 as well. before we proceed, can you please investigate whether 2 is happening invisibly on other estimators? just the most common ones (`Development`, `Chainladder`, `BornhuetterFerguson`)

ppcvote:
Checked all three. The inference is unique to `CapeCod`, and the reason looks structural rather than accidental.

Fitting at LOB grain on `clrd` and applying at company grain, on the #1275 branch:

| estimator | fitted total | applied total | what comes back |
| --- | --- | --- | --- |
| `Development.transform` | n/a | n/a | `ldf_` stays 6 rows, values identical to the fitted ones |
[...]

PR #1275, opened 2026-09-03
[FIX] CapeCod.predict discards the fitted apriori when the prediction data has exactly one extra index level

### LINKED ISSUE #1265: CapeCod.predict discards the fitted apriori when the model was fit on exactly one fewer index level

- [x] Yes, this bug occurs on the latest version (v0.10.0, verified against `main` at 5f34bcd).

> Revised after @henrydingliu's note: this issue originally said "grain" where it meant index levels. Reworded throughout, since grain properly refers to origin and development period granularity.

### Describe the bug in words

When a `CapeCod` model is fit on training data that has already been aggregated to fewer index levels, `predict()` silently discards the fitted apriori and recomputes it at the full index. Pooling sparse segments is the main reason to aggregate before fitting, so the returned `apriori_` is not the quantity the caller fitted.

The guard at `chainladder/methods/capecod.py:325` reads:

```python
# If model was fit at a higher grain, then need to aggregate predicted aprioris too
if len(set(sample_weight.key_labels) - set(self.apriori_.key_labels)) > 1:
```

That set difference is the number of index levels the prediction carries which the fitted apriori does not, so it is `>= 1` exactly when the model was fit on fewer levels. At `> 1` the commonest case, one fewer level, falls through to the `else` and recomputes at the full index.

Using the repository's own `clrd` sample, whose `key_labels` are `['GRNAME', 'LOB']`:

```python
import chainladder as cl
import numpy as np

clrd = cl.load_sample("clrd")
tri  = clrd[clrd["LOB"] == "comauto"]["CumPaidLoss"]
prem = clrd[clrd["LOB"] == "comauto"]["EarnedPremDIR"].latest_diagonal

# aggregate to one fewer index level before fitting: LOB only, dropping GRNAME
cc = cl.CapeCod().fit(tri.groupby("LOB").sum(),
                      sample_weight=prem.groupby("LOB").sum())

print(len(set(prem.key_labels) - set(cc.apriori_.key_labels)))   # 1  -> guard is False
print(float(np.asarray(cc.apriori_).ravel()[0]))                 # 0.5689995797  (fitted)

pred = cc.predict(tri, sample_weight=prem)
print(np.asarray(pred.apriori_).ravel()[:1])                     # 1.2516635774  (recomputed)
```

The fitted apriori is `0.5689995797`; `predict()` returns `1.2516635774`. I checked the fitted value by hand in numpy, `sum(latest) / sum(premium / cdf)`, and it agrees to 1e-9.

The existing `predict` tests cannot catch this because both fit at `['Line']` on `prism`, whose `key_labels` are six long, so the difference is always 5 and they always take the intended branch. `clrd` has a difference of 1 and is not covered.

```python
assert np.allclose(np.asarray(pred.apriori_), np.asarray(cc.apriori_))
```

Changing the threshold to `> 0` produces that. I checked the blast radius: with `> 0` the `diff == 0` case is bit-identical (still `1.2352941176` on the same data) and `diff >= 2` is unchanged, so only the broken case moves.

One thing worth deciding rather than assuming: with `> 0`, a genuinely disjoint index turns today's `ValueError: Index broadcasting is ambiguous` into a less helpful `KeyError`. An explicit subset check would read better than a count, and I did not want to pick that shape for you.

- [x] Yes, absolutely!

ppcvote:
## Summary of Changes

`CapeCod.predict()` regroups the prediction data up to the grain the model was fit at, and decides whether to do so by counting the index levels `sample_weight` carries that `apriori_` does not. The count required more than one, so the ordinary case of a single extra level fell through to the ungrouped branch and recomputed the apriori from the prediction data instead of using the fitted one.

```python
import chainladder as cl

clrd = cl.load_sample("clrd")
tri, prem = clrd["CumPaidLoss"], clrd["EarnedPremDIR"].latest_diagonal

model = cl.CapeCod().fit(tri.groupby("LOB").sum(),
                         sample_weight=prem.groupby("LOB").sum())
model.predict(tri, sample_weight=prem)
```

The fitted comauto apriori is 0.5689995797. Before this change `predict()` returned 1.2516635774, and all 775 rows differed from the value fitted for their LOB. After it, every row carries its fitted apriori and `expectation_` is exactly premium times that apriori across all 775 rows.

`test_capecod_predict2` already covers this code path, but it uses `prism`, whose triangle carries five index levels more than the fitted model, so it stays on the grouped branch. The added test uses `clrd`, which carries exactly one.

## Related GitHub Issue(s)

Fixes #1265. The API question that thread raised is split out into #1274.

- `pytest chainladder` passes: 1130 passed, 7 skipped.
- `ruff check` and `ruff format --check` under the CI flags report the same eight findings on these two files before and after the change, so the added lines introduce none of their own.
- The new test fails on the unpatched line and passes on the patched one.
- The `predict()` docstring example (ukmotor) is unaffected: it fits and predicts at the same grain, so the level difference is empty and neither branch changes. Its documented output still reproduces exactly.

## Declarations

I am adhering to the standards in the Governing Doc. I am a human contributor, not a bot, and I opened this because @henrydingliu asked for the `> 0` fix in #1265.

AI disclosure, per the AI Usage Policy: I used Claude Code on this. It reproduced the bug against `clrd`, ran the before and after comparison, and drafted the regression test. The one-character fix itself is the one @henrydingliu and I settled on in #1265 before any code was written. I reviewed the diff and the test myself, and the suite was run locally on my machine.

## Checklist

  `pytest` locally: 1130 passed, 7 skipped. There are no documentation changes in this PR, so I did not run the docs doctest build. The only docstring the change could reach is the `CapeCod.predict` ukmotor example, and I ran that directly: it still reproduces its documented output exactly. CI's Doctest job passes on this branch.


henrydingliu:
looks good! would you mind picking up the couple ruff linter fixes?

ppcvote:
Done. `ruff check --force-exclude --config lint.per-file-ignores={}` now passes on both files: the `numpy` import in `capecod.py` was unused, the rest was whitespace in existing tests. I also dropped the two `per-file-ignores` entries for these files, since the table's comment says to remove them as files are cleaned up and both are dead now. Rebased onto current `main`.

I left `ruff format` alone. It touches 40 of the 107 lines in `test_capecod.py` and grows it to 132, and six of the eight files under `chainladder/methods` are unformatted on `main`, so it looked like a separate decision rather than something to slip into this PR. Say the word if you would rather have it here.

henrydingliu (approved):

henrydingliu:
thanks again!

Issue #1283, opened 2026-09-04
[CHORE] Move options and config related tests to _config folder.


genedan:
Options are tested in `test_utilities.py`, but with the creation of the new `_config` folder we should move options/config tests to that folder as well.

Issue #1281, opened 2026-09-04
[FEAT] Make Jupyter auto-formatting configurable


genedan:
In Jupyter, triangles are formatted based on the magnitude of values within them:

https://github.com/casact/chainladder-python/blob/14becfc30ac2b06d04667ab9b614b65fff44be1d/chainladder/core/display.py#L97-L113

We should move the hardcoding out of the function, and make the thresholds configurable:

- Allow configuration of format strings for small/medium/large triangles.
- Allow configuration of thresholds at which a triangle is determined to be small/medium/large.

PR #1276, opened 2026-09-03
[CHORE] Move options, deprecations into chainladder._config

genedan:
## Summary of Changes  

Moves options and deprecation utilities into a new folder, `_config`. This is to adhere to the single-responsibility principle. i.e., each module should be responsible for one thing.

The overall package `__init__.py` is now only responsible for defining the public API - with the exception of the `__version__` constant. I didn't have strong feelings about where to put that, and pandas just leaves it in `__init__.py`. I could move it too, if you want.

Refactored `_dask_parallel_warned: bool = False` to be a class, this allows it to better maintain state across multiple modules that use it when it's located in a deeper location than `__init__.py`. This pattern also matches how the `Options` class maintains the state of its options, to keep things consistent.

## Related GitHub Issue(s)  

#1145 

## Additional Context for Reviewers  

Made some ruff fixes on existing files.

## Checklist
- [x] I passed tests locally for both code (`uv run pytest`) and documentation changes (`uv run --directory docs jb build . --builder=custom --custom-builder=doctest`)

---

> [!NOTE]
> **Low Risk**
> Mostly a module move with re-exports preserving the public API; small internal state and type-guard changes are covered by existing option/deprecation tests.
> 
> **Overview**
> **Refactors package configuration** out of the top-level `chainladder/__init__.py` into a new `chainladder/_config` subpackage (`options.py` for `Options`, datetime constants, and `options`; `deprecation.py` for backend deprecation helpers and `_resolve_pat`). The public surface stays the same: `cl.options`, `cl.Options`, `__dt64_*`, and the existing private re-exports on `chainladder` are unchanged.
> 
> `ARCHITECTURE.md` documents the new `_config` layout. **Minor implementation tweaks:** one-time dask parallel deprecation state is now `_dask_parallel_state.warned` (replacing module-level `_dask_parallel_warned`), and `set_option` for `ARRAY_BACKEND` only warns on deprecated backends when the value is a `str`.
> 
> Tests in `test_utilities.py` were updated for the new state object and formatted to satisfy ruff (the file’s per-file ignore entry was removed from `pyproject.toml`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3f1986cdec7adcad88f29a65678dccc145c79735. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

Issue #1154, opened 2026-07-25
Create options for triangle and pattern formatting


genedan:
### Description

Chainladder is used around the world, so we should have a way to adjust the numeric formatting to accommodate various locales. 

Chainladder currently does not have a centralized configuration for `Triangle` formatting. Printing to console vs. printing to a Jupyter notebook will lead to different levels of precision since the former is handled by Pandas whereas the latter is inferred via this function:

https://github.com/casact/chainladder-python/blob/07a92af807d14c7f4fb4d64b88a943832dd3ddeb/chainladder/core/display.py#L100-L116

For console and terminal output, there is no chainladder-native way to configure the formatting, you need to go through Pandas. See https://github.com/casact/chainladder-python/discussions/875

### Is your feature request aligned with the scope of the package?

- [x] Yes, absolutely!
- [ ] No, but it's still worth discussing.
- [ ] N/A (this request is not a codebase enhancement).

### Describe the solution you'd like, or your current workaround.

1. Move the auto-formatting based on magnitude to the `Options` class as the default for all output.
2. Add options to fine-tune numeric formatting.

Examples:

```python
import pandas as pd
import chainladder as cl

data = pd.DataFrame({
    "origin":     ["2020", "2020", "2020", "2021", "2021", "2022"],
    "valuation":  ["2020-12-31", "2021-12-31", "2022-12-31", "2021-12-31", "2022-12-31", "2022-12-31"],
    "values":     [1234567, 2345678, 2456789, 1345678, 2456789, 1456789],
})

tri = cl.Triangle(data, origin="origin", development="valuation", columns="values", cumulative=True)

print(tri)
             12         24         36
2020  1234567.0  2345678.0  2456789.0
2021  1345678.0  2456789.0        NaN
2022  1456789.0        NaN        NaN
```

**Thousands separator, accepts a format string**:
```python
cl.options.set_option("display.value_format", "{:,.0f}")

print(tri)
             12         24         36
2020  1,234,567  2,345,678  2,456,789
2021  1,345,678  2,456,789
2022  1,456,789
```

**Swiss-format, accepts a callable**:

```python
def swiss_format(x):
      return f"{x:,.0f}".replace(",", "'")

cl.options.set_option("display.value_format", swiss_format)
print(tri)

               12         24         36
  2020  1'234'567  2'345'678  2'456'789
  2021  1'345'678  2'456'789
  2022  1'456'789
```

**Indian format:**
```python
def indian_format(x):
[...]

PR #1284, opened 2026-09-04
[FEAT] Add numeric formatting options

genedan:
# Must approve PR #1276 before this one.

## Summary of Changes  

- Adds options to pass format string/callable to configure numeric formatting (values and link ratios).
- Makes auto-formatting based on triangle size in Jupyter/HTML output configurable.
- Moves options-related tests to _config folder.

## Related GitHub Issue(s)  

#1154 
#1281 
#1283

## Checklist

---

> [!NOTE]
> **Low Risk**
> Display-only behavior with defaults matching prior HTML auto-formatting; global options are mutable but tested and reset in tests.
> 
> **Overview**
> Adds **package-wide display formatting** via dotted `cl.options` keys so users can control how triangles print in the console, Jupyter/HTML, and `heatmap()`.
> 
> **`display.value_format`** and **`display.pattern_format`** accept a format string or callable; when set they override defaults for value triangles vs pattern triangles (e.g. link ratios) respectively. When unset, console `repr` keeps Pandas’ default floats while HTML/heatmap still auto-pick precision from magnitude—now driven by configurable **`display.html.auto_format_*`** strings and **`display.html.auto_format_*_threshold`** cutoffs (replacing hardcoded 10/1000 logic in `TriangleDisplay`).
> 
> `describe_option` parsing is relaxed so multi-word type hints (e.g. `str, callable, or None`) document correctly. Options tests move from `utils/tests/test_utilities.py` into **`chainladder/_config/tests/test_options.py`**; display tests cover overrides, thresholds, and pattern/value independence.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 23625d97c329827dfc4cd27c4e8101b70d5aa494. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chainladder/core/display.py
```diff
@@ -105,12 +121,45 @@ def _get_format_str(data: DataFrame) -> str:
         """
         if np.all(np.isnan(data)):
             return ""
-        elif np.nanmean(abs(data)) < 10:
-            return "{0:,.4f}"
-        elif np.nanmean(abs(data)) < 1000:
-            return "{0:,.2f}"
+        elif np.nanmean(abs(data)) < options.get_option(
```
henrydingliu, line 124: small nitpick. not uncommon to have massive factor(s) in the first dev period for either a selected pattern or a ldf triangle. are you open to setting the default pattern format to 4 decimals?

chainladder/_config/options.py
```diff
@@ -65,6 +96,15 @@ def __init__(self):
         self.ULT_VAL = str(
             pd.Timestamp("2262-01-01") - pd.Timedelta(1, unit=__dt64_unit__)  # noqa
         )
+        # Dotted names are supported transparently: setattr/getattr/vars()
+        # all work with arbitrary string keys, not just valid identifiers.
+        setattr(self, "display.value_format", None)
+        setattr(self, "display.pattern_format", None)
+        setattr(self, "display.html.auto_format_small", "{0:,.4f}")
+        setattr(self, "display.html.auto_format_medium", "{0:,.2f}")
+        setattr(self, "display.html.auto_format_large", "{:,.0f}")
```
henrydingliu, line 105: does python formatting allows you to display numbers in thousands or millions (e.g. 1,234,567 as 1,235) and do you think there's value to adding that to the package?

chainladder/core/display.py
```diff
@@ -176,7 +225,10 @@ def heatmap(
         """
         if self._dimensionality == "single":
             data = self._repr_format()
-            fmt_str = self._get_format_str(data)
+            value_format = self._display_option(is_pattern=self.is_pattern)
```
henrydingliu, line 228: why not autoformat for console as well?

chainladder/_config/options.py
genedan: Yeah it does. I think this PR should already accommodate that, since it can accept an arbitrary format string.

chainladder/core/display.py
genedan: I think the Jupyter autoformat was something that was personally useful to John and Kenneth, presumably in their own work or in building the documentation. 

I think when making executive-level/public presentations, this makes sense, you need the thousands separator, and it looks better than having the lower diagonal filled with NaNs. For console output I think it makes less sense since programmers often work with raw data.

I think it's possible to allow both Jupyter and console output to be customized by the user (changing the number of magnitude tiers, NaN representation, etc.) by accepting a function. 

However, I don't think the auto-format should be the default setting. If the user didn't want any auto-formatting in Jupyter, there's no way to change that at the moment. It would make more sense to default to base Python/Pandas and then allow for customization. Then for the docs, we can set auto-format at the beginning and have the setting cascade through the rest of the docs.

chainladder/_config/options.py
henrydingliu: we can leave auto-format-ex-large as a future enhancement

chainladder/core/display.py
genedan: Hmm I've got a better idea. I'm going to revert the small/medium/large settings (they were out of the original scope anyway) - they are highly specific to a formatting scheme that I'm not sure everyone would want.

Instead we should simply have the current auto-format as a built in option - akin to a pretty print, while allowing the option for anything else via a callable or format string. I'll articulate this idea for general feedback in a new issue.

chainladder/core/display.py
genedan: If the average of all the factors is greater than 1000 due to the first factor...yeah the rest will be integers in the current version, even if it's something like 1.234.

This auto-format stuff shouldn't be the default. I'll articulate more in the new issue I said I'd create in my other comment. In short, I think the current Jupyter formatting is very specific to the "Executive Summary" style presentation of the Chainladder docs, but might not be applicable to general use. We should change this to a non-default option that gets toggled on just for the docs.

henrydingliu: sounds good

PR #1302, opened 2026-09-05
Add Triangle.ffill() to forward-fill along the origin/development axes

### LINKED ISSUE #1030: Add Triangle.ffill()
### Description

Pandas [DataFrame.ffill()](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.ffill.html#pandas.DataFrame.ffill) can be mirrored to fill NaNs in a forward direction along the specified axis.

Leaving this as a proposed "Great First Contribution" as this would be easier than the proposed interpolation/extrapolation feature.

- [x] Yes, absolutely!

```python
tri = cl.Triangle(
    data={
        'origin': [1985, 1985, 1985, 1985, 1986, 1986, 1986, 1987, 1987, 1988],
        'development': [1985, 1986, 1987, 1988, 1986, 1987, 1988, 1987, 1988, 1988],
        'paid': [500, np.nan, 700, np.nan, np.nan, 1000, 1100, 1200, 1300, np.nan]
    },
    origin='origin',
    development='development',
    columns=['paid'],
    cumulative=True
)
tri

          12      24      36  48
1985   500.0     NaN   700.0 NaN
1986     NaN  1000.0  1100.0 NaN
1987  1200.0  1300.0     NaN NaN
1988     NaN     NaN     NaN NaN
```

```python
tri.ffill(axis=3)

          12      24      36     48
1985   500.0   500.0   700.0  700.0
1986     NaN  1000.0  1100.0 1100.0
1987  1200.0  1300.0     NaN    NaN
1988     NaN     NaN     NaN    NaN
```

```python
tri.ffill(axis=4)
          12      24      36  48
1985   500.0     NaN   700.0 NaN
1986   500.0  1000.0  1100.0 NaN
1987  1200.0  1300.0     NaN NaN
1988  1200.0     NaN     NaN NaN

_No response_

- [ ] Yes, absolutely!
- [ ] No.

### PR DESCRIPTION
**Pull request by Abhayindia**
Closes #1030.

Adds `Triangle.ffill(axis)`, mirroring pandas' `ffill` but along the origin or development axis, and never writing a value into a cell that hasn't been valued yet.

On the semantics question I asked in #1030 and never got an answer to: genedan's own worked example in the issue answers it, once you check it against the triangle's actual `nan_triangle`. In his `axis=3` sketch, 1987 at ages 36/48 stays missing even though 1300 precedes it - because those cells are beyond the triangle's evaluation date, not because they're a "later" NaN. His `axis=4` example (there's no axis 4, so I read it as origin) is consistent with the same rule the whole way through: 1988 at ages 24-48 stays missing for the same reason. There's exactly one cell in the whole issue that breaks the pattern - 1986 at age 48 in the `axis=3` example, which the sketch fills to 1100 even though that cell is also beyond the evaluation date. Given eleven other cells across both examples agree and only this one doesn't, I'm reading it as a typo in the hand-written sketch rather than the intended behavior, and implemented the self-consistent rule: fill forward within the observed triangle, leave everything outside it untouched.

Docstring example reuses that same triangle from the issue (with the corrected 1986/48 cell) for both axes. Tests cover the development and origin directions, that it returns a new object rather than mutating in place, and the axis validation.

```
$ pytest chainladder/core/tests/test_triangle.py -q
353 passed, 1 skipped in 813s
```

### CODE REVIEW

**Comment by genedan** [comment]
I'll review this one. @Abhayindia, thanks for the contribution. From first glance, could you move the ffill method to be a method of the `TrianglePandas` class in `pandas.py`? This is where we put the methods that we mirror from Pandas.

I'll get around to answering your remaining questions tomorrow.

**Comment by genedan** [comment]
Whoops, just saw that henry requested @priyam0k on this one - it's yours to review if you still want it.

**Comment by henrydingliu** [comment]
you can review priyam's prs :)

**Comment by genedan** [comment]
You got it. If you guys don't mind, would you be able to get that Ruff stack approved beforehand? Priyam's PR has a merge conflict with mine. We'll have to resolve either way but since my stack is bigger it'll probably be easier to have Priyam resolve on his end.

**Comment by Abhayindia** [comment]
moved it onto TrianglePandas next to fillna/fillzero. no rush on the semantics questions - whenever you get to them.

File: chainladder/core/pandas.py [hunk 1]
```diff
@@ -589,6 +589,114 @@ def fillzero(self, inplace: bool = False) -> Triangle:
...
+        n = self.shape[axis]
+        columns = (
+            [self.iloc[..., i : i + 1] for i in range(n)]
+            if axis == 3
+            else [self.iloc[..., i : i + 1, :] for i in range(n)]
+        )
+        filled = [columns[0]]
+        for current in columns[1:]:
+            previous = filled[-1]
+            is_missing = xp.nan_to_num(current.values) == 0
```
priyam0k: If a triangle has a valid `0.0` (e.g. 0 paid claims), `nan_to_num(0.0) == 0` evaluates to `True`, so genuine zeros get treated as missing and overwritten here.
Can we switch this to..
```python
is_missing = xp.isnan(current.values)

priyam0k (commented):
Thanks for putting this together @Abhayindia 
Do you think we should just use isnan directly and mask with nan_triangle using where so legitimate zeros stay intact? Also, could we add a quick test case with zeros to be safe?

Abhayindia:
switched to `xp.isnan` for the gap check and `where` for the future-cell mask (`cd24e1d2`). turns out the constructor already maps a 0 in the input to NaN, so a legit interior zero isn't representable and both forms give the same result - but isnan states the intent better. added `test_ffill_zero_input_is_missing_and_fills` pinning that.

priyam0k (approved):

henrydingliu (approved):
no additional concerns

Issue #1053, opened 2026-06-27
[DISCUSSION] Create centralized axis mapping


genedan:
### Description

The recurring pattern `{0: 'index', 1: 'columns', 2: 'origin', 3: 'development'`}` is used to look up and map axes values, and validate arguments.

It's not applied consistently though, there's a spot in pandas.py where we have the keys and values reversed in order to extract the integer value: 

`{"index": 0, "columns": 1, "origin": 2, "development": 3}`

Anyhow, keeping the axes in a central location such as `__init__.py` will improve maintainability and make it clear which axes are supported, as well as solidify the text to integer mapping.

### Is your feature request aligned with the scope of the package?

- [x] Yes, absolutely!
- [ ] No, but it's still worth discussing.
- [ ] N/A (this request is not a codebase enhancement).

### Describe the solution you'd like, or your current workaround.

I propose a tuple: `__axes__ = ('index', 'columns', 'origin', 'development')`. This is immutable and can be used to lookup values in both directions:

```
print(__axes__[2])

'origin'

print(__axes__.index('origin'))

'2'
```

### Do you have any additional supporting notes?

Labeling this as a discussion, to see if there's a better way to do this. I don't really like how a tuple has a method called `index` which can be confused with the Triangle's index.

Will drop the discussion label and set status to ready to be picked up once this is decided.

### Would you be willing to contribute this ticket?

- [ ] Yes, absolutely!
- [x] Yes, but I would like some help.
- [ ] No.

henrydingliu:
does this have any bearing on what we discussed a few weeks ago in #601? potentially, if we have end up creating some type of axis object, we could just map both `'origin'` and `2` to the `OriginAxis()`? 

how does pandas handle it?

genedan:
In pandas, `AxisProperty` just stores the integer, 0 or 1. For methods that accept a string argument to the `axis` parameter, i.e., `DataFrame.rename(..,axis='columns')`, that argument is routed to a dict that looks up the integer.

https://github.com/pandas-dev/pandas/blob/8188eb1d65d6250c9916e54a0fa417d46af3296a/pandas/core/generic.py#L503

So the `OriginAxis()` would only need to store `2`, for methods that accept `axis="origin"`, it would look up the integer via:

`{"index": 0, "columns": 1, "origin": 2, "development": 3}`

genedan:
I just realized the link from the last post doesn't have a columns key.

[...]

PR #1303, opened 2026-09-05
feat(core): deprecate vdims (#1011) and kdims (#1010)

### LINKED ISSUE #1010: Deprecate kdims
Subissue of #601. kdims is an alias for the Triangle index. kdims is to be removed without changing the public API.

### LINKED ISSUE #1011: Deprecate vdims
### Description

Subissue of #601. vdims is an alias for the Triangle columns. vdims is to be removed without changing the public API.

- [ ] Yes, absolutely!

_No response_

_No response_

- [ ] Yes, absolutely!
- [ ] No.

priyam0k:
## Summary of Changes
- Deprecate vdims (#1011) and kdims (#1010) on Triangle with FutureWarnings pointing to columns/columns_label and index/key_labels
- Update internal usage across core modules, estimators, adjustments, and tails to prevent internal warnings
- Add test coverage verifying deprecation warnings when reading or setting both attributes
- Address ruff lint and formatting after rebasing on main (added noqa: N802 on Clark and Munich formulas to preserve actuarial notation)

## Related GitHub Issue(s)
Closes #1010, closes #1011, refs #601, refs #1216

Existing code reading or setting vdims and kdims continues to work as expected for backward compatibility. Date dimensions (odims and ddims) will follow in a separate PR.

## Checklist

---

> [!NOTE]
> **High Risk**
> This is a broad refactor of Triangle’s dimension metadata and alignment paths; regressions could affect slicing, arithmetic, concat, and serialized objects even with migration logic.
> 
> **Overview**
> Replaces internal **`kdims` / `vdims` numpy labels** with a pandas-aligned model: row keys live in **`_index`** (`Triangle.index` as a `DataFrame`), value columns in **`_axes["columns"]`** via a new **`TriangleAxis`** descriptor, plus **`key_labels`** as index column names and an **`axes`** property.
> 
> **Indexing, arithmetic, slicing, groupby, concat, and I/O** are updated to use `index` / `columns` (e.g. broadcast/align on `DataFrame.index`, column assignment through the descriptor, JSON/pickle skipping or migrating `_index`). **`concat`** gains clearer **`ignore_index`** handling on columns, origin, and development; **`sort_index(inplace=True)`** and **boolean `Series`** row selection are supported.
> 
> Downstream code (bootstrap resampling, correlation tables, Benktander, tails, etc.) assigns **`index`** / **`columns`** instead of mutating `kdims`/`vdims`. **`__setstate__`** migrates old pickles; tests cover columns/index setters, copy isolation, JSON roundtrip, and legacy unpickling.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b17787be48fdb5bd9cb39a3b741593a043eb5a8a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/chainladder/adjustments/bootstrap.py
+++ b/chainladder/adjustments/bootstrap.py
@@ -267,16 +267,17 @@ class BootstrapODPSample(DevelopmentBase):
         resampled_triangles = resampled_triangles[None, ...].swapaxes(0, 1)
         obj = X.copy()
         if X.key_labels == ["Total"]:
-            obj._kdims = np.arange(self.n_sims)
+            obj._index = pd.DataFrame({"Simulation_#": np.arange(self.n_sims)})
             obj.key_labels = ["Simulation_#"]
         else:
-            obj._kdims = np.concat(
+            kdims = np.concat(
                 [
-                    np.tile(X._kdims, (self.n_sims, 1)),
+                    np.tile(X.index.to_numpy(dtype=str), (self.n_sims, 1)),
                     np.arange(self.n_sims).reshape(-1, 1),
                 ],
                 axis=1,
             )
+            obj._index = pd.DataFrame(kdims, columns=X.key_labels + ["Simulation_#"])
             obj.key_labels = X.key_labels + ["Simulation_#"]
         obj.values = resampled_triangles
         obj._set_slicers()
--- a/chainladder/adjustments/bootstrap.py
+++ b/chainladder/adjustments/bootstrap.py
@@ -219,11 +219,6 @@ class BootstrapODPSample(DevelopmentBase):
             self.resampled_triangles_, self.scale_ = self._get_simulation(
                 X, exp_incr_triangle
             )
-            # n_obs = xp.nansum(self.w_)
-            # n_origin_params = X.shape[2]
-            # n_dev_params = X.shape[3] - 1
-            # deg_free = n_obs - n_origin_params - n_dev_params
-            # deg_free_adj_fctr = xp.sqrt(n_obs / deg_free)
         return self
 
     def _get_simulation(self, X, exp_incr_triangle):
@@ -272,12 +267,12 @@ class BootstrapODPSample(DevelopmentBase):
         resampled_triangles = resampled_triangles[None, ...].swapaxes(0, 1)
         obj = X.copy()
         if X.key_labels == ["Total"]:
-            obj.kdims = np.arange(self.n_sims)
+            obj._kdims = np.arange(self.n_sims)
             obj.key_labels = ["Simulation_#"]
         else:
-            obj.kdims = np.concat(
+            obj._kdims = np.concat(
                 [
-                    np.tile(X.kdims, (self.n_sims, 1)),
+                    np.tile(X._kdims, (self.n_sims, 1)),
                     np.arange(self.n_sims).reshape(-1, 1),
                 ],
                 axis=1,
--- a/chainladder/adjustments/bootstrap.py
+++ b/chainladder/adjustments/bootstrap.py
@@ -270,15 +270,10 @@ class BootstrapODPSample(DevelopmentBase):
             obj._index = pd.DataFrame({"Simulation_#": np.arange(self.n_sims)})
             obj.key_labels = ["Simulation_#"]
         else:
-            kdims = np.concat(
-                [
-                    np.tile(X.index.to_numpy(dtype=str), (self.n_sims, 1)),
-                    np.arange(self.n_sims).reshape(-1, 1),
-                ],
-                axis=1,
-            )
-            obj._index = pd.DataFrame(kdims, columns=X.key_labels + ["Simulation_#"])
-            obj.key_labels = X.key_labels + ["Simulation_#"]
+            obj_idx = pd.concat([X.index] * self.n_sims, ignore_index=True)
+            obj_idx["Simulation_#"] = np.arange(self.n_sims)
+            obj._index = obj_idx
+            obj.key_labels = list(obj_idx.columns)
         obj.values = resampled_triangles
         obj._set_slicers()
         return obj, scale_phi
--- a/chainladder/core/dunders.py
+++ b/chainladder/core/dunders.py
@@ -6,10 +6,7 @@ from __future__ import annotations
 import numpy as np
 import pandas as pd
 
-from chainladder.utils.utility_functions import (
-    num_to_nan,
-    concat
-)
+from chainladder.utils.utility_functions import num_to_nan, concat
 
 from chainladder.core.pandas import TriangleGroupBy
 from chainladder.utils.sparse import sp
@@ -25,13 +22,14 @@ from typing import TYPE_CHECKING
 if TYPE_CHECKING:
     from typing import Any
 
+
 class TriangleDunders:
-    """ Class that implements the dunder (double underscore) methods for the
-        Triangle class
+    """Class that implements the dunder (double underscore) methods for the
+    Triangle class
     """
 
     def _validate_arithmetic(self, other: Any) -> tuple:
-        """ Common functionality BEFORE arithmetic operations """
+        """Common functionality BEFORE arithmetic operations"""
         if isinstance(other, TriangleDunders):
             obj, other = self._compatibility_check(self, other)
             if obj.is_pattern != other.is_pattern:
@@ -43,17 +41,17 @@ class TriangleDunders:
             if isinstance(other, TriangleDunders):
                 other = other.values
         else:
-            if isinstance(other, np.ndarray) and self.array_backend != 'numpy':
+            if isinstance(other, np.ndarray) and self.array_backend != "numpy":
                 obj = self.copy()
                 other = obj.get_array_module().array(other)
-            elif isinstance(other, sp.COO) and self.array_backend != 'sparse':
-                obj = self.set_backend('sparse')
+            elif isinstance(other, sp.COO) and self.array_backend != "sparse":
+                obj = self.set_backend("sparse")
             else:
                 obj = self.copy()
         return obj, other
 
     def _arithmetic_cleanup(self, obj):
-        """ Common functionality AFTER arithmetic operations """
+        """Common functionality AFTER arithmetic operations"""
         obj.values = obj.values * obj.get_array_module().nan_to_num(obj.nan_triangle)
         obj.values = num_to_nan(obj.values)
         return obj
@@ -62,10 +60,9 @@ class TriangleDunders:
         from chainladder.utils.utility_functions import set_common_backend
 
         x, y = set_common_backend([x, y])
-        if (
-            x.origin_grain != y.origin_grain
-            or (x.development_grain != y.development_grain and 
-                min(x.shape[-1], y.shape[-1]) > 1)
+        if x.origin_grain != y.origin_grain or (
+            x.development_grain != y.development_grain
+            and min(x.shape[-1], y.shape[-1]) > 1
         ):
             raise ValueError(
                 "Triangle arithmetic requires both triangles to be the same grain."
@@ -83,7 +80,9 @@ class TriangleDunders:
             return x, y
         if x.kdims.shape[0] == y.kdims.shape[0] == 1 and x.key_labels != y.key_labels:
             kdims = x.kdims if len(x.key_labels) > len(y.key_labels) else y.kdims
-            key_labels = x.key_labels if len(x.key_labels) > len(y.key_labels) else y.key_labels
+            key_labels = (
+                x.key_labels if len(x.key_labels) > len(y.key_labels) else y.key_labels
+            )
             x.kdims = y.kdims = kdims
             x.key_labels = y.key_labels = key_labels
             return x, y
@@ -97,7 +96,11 @@ class TriangleDunders:
             if x_labels != y_labels or x.kdims.shape[0] != y.kdims.shape[0]:
                 x = x.groupby(list(common))
                 y = y.groupby(list(common))
-            elif x.kdims.shape[0] > 1 and not np.array_equal(x.kdims, y.kdims) and not x.index.equals(y.index):
+            elif (
+                x.kdims.shape[0] > 1
+                and not np.array_equal(x.kdims, y.kdims)
+                and not x.index.equals(y.index)
+            ):
                 x = x.sort_index()
                 try:
                     y = y.loc[x.index]
@@ -107,7 +110,12 @@ class TriangleDunders:
             return x, y
 
         else:
-            raise ValueError('Index broadcasting is ambiguous between ' + str(x_labels) + ' and ' + str(y_labels))
+            raise ValueError(
+                "Index broadcasting is ambiguous between "
+                + str(x_labels)
+                + " and "
+                + str(y_labels)
+            )
 
     def _prep_columns(self, x, y):
         if len(x.columns) == 1 and len(y.columns) > 1:
@@ -118,9 +126,9 @@ class TriangleDunders:
             y._vdims = x._vdims
         elif x.shape[1] == y.shape[1] and np.array_equal(x.columns, y.columns):
             return x, y
-        else:            
+        else:
             # Find columns to add to each triangle
-            cols_to_add_to_x = [col for col in y.columns if col not in x.columns] 
+            cols_to_add_to_x = [col for col in y.columns if col not in x.columns]
             cols_to_add_to_y = [col for col in x.columns if col not in y.columns]
 
             # Start with case with no new columns, y simply has a different order
@@ -145,7 +153,8 @@ class TriangleDunders:
         xp = obj.get_array_module()
         is_broadcastable = all(
             (m == n) or (m == 1) or (n == 1)
-            for m, n in zip(obj.shape[-2:][::-1], other.shape[-2:][::-1]))
+            for m, n in zip(obj.shape[-2:][::-1], other.shape[-2:][::-1])
+        )
         if (len(obj.odims), len(obj.ddims)) == (len(other.odims), len(other.ddims)):
             if np.all(obj.ddims != other.ddims) and len(obj.ddims) > 1:
                 is_broadcastable = False
@@ -191,10 +200,20 @@ class TriangleDunders:
             ldl = int(np.where(~d_arr0 == 1)[0].min())
             ldh = int(np.where(~d_arr0 == 1)[0].max() + 1)
             if obj.array_backend != "sparse":
-                other_arr = xp.zeros((other.shape[0], other.shape[1], len(odims), len(ddims)))
+                other_arr = xp.zeros((
+                    other.shape[0],
+                    other.shape[1],
+                    len(odims),
+                    len(ddims),
+                ))
                 other_arr[:] = xp.nan
                 other_arr[:, :, rol:roh, rdl:rdh] = other.values
-                obj_arr = xp.zeros((self.shape[0], self.shape[1], len(odims), len(ddims)))
+                obj_arr = xp.zeros((
+                    self.shape[0],
+                    self.shape[1],
+                    len(odims),
+                    len(ddims),
+                ))
                 obj_arr[:] = xp.nan
                 obj_arr[:, :, lol:loh, ldl:ldh] = obj.values
             else:
@@ -203,7 +222,12 @@ class TriangleDunders:
                 other_arr.coords[3] = other_arr.coords[3] + rdl
                 obj_arr.coords[2] = obj_arr.coords[2] + lol
                 obj_arr.coords[3] = obj_arr.coords[3] + ldl
-                other_arr.shape = (other.shape[0], other.shape[1], len(odims), len(ddims))
+                other_arr.shape = (
+                    other.shape[0],
+                    other.shape[1],
+                    len(odims),
+                    len(ddims),
+                )
                 obj_arr.shape = (self.shape[0], self.shape[1], len(odims), len(ddims))
             obj.odims = np.array(odims.index)
             if isinstance(obj.ddims, pd.DatetimeIndex):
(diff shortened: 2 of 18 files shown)


chainladder/core/triangle.py
```diff
@@ -724,30 +724,79 @@ def index(self) -> DataFrame:
...
+            "The 'vdims' attribute is deprecated and will be removed in a future release. "
+            "Use 'Triangle.columns' instead.",
+            FutureWarning,
+            stacklevel=2,
+        )
+        if type(value) is str:
+            value = np.array([value])
+        elif type(value) is list:
+            value = np.array(value)
+        self._vdims = value
```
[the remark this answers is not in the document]
priyam0k: Resolved in commit bf0e00a7

genedan:
Some general thoughts on your approach:

- I see that `kdims` and `vdims` have been renamed `_kdims` and `_vdims`. What we'd like to do is actually get rid of these labels and define the Triangle's `index` and `columns` directly.
- `Triangle.index` and `Triangle.columns` should no longer be properties of the Triangle, they should resolve to `Index/MultiIndex` data types

I have an attempt at a general approach in #601, starting with [this comment](https://github.com/casact/chainladder-python/issues/601#issuecomment-4742913966). What I tried to do was examine how a Pandas DataFrame initialized its columns:

https://github.com/pandas-dev/pandas/blob/721ab260c67dd480e83d1510fa5ac08553536038/pandas/core/frame.py#L4692-L4697

And see if we can port over the logic. One strategy I would suggest trying is changing `kdims` and `vdims` to be `index` and `columns`, and then get rid of the old properties. Hope that helps.

priyam0k:
I went through your comments in #601 and the Pandas AxisProperty references.

Storing `columns` as a direct `pd.Index` on the Triangle (and replacing `_vdims` references with `self.columns.values`) makes total sense and eliminates the need for `_vdims` completely. I can definitely update that.

For _Triangle.index_, what are your thoughts on keeping it stored as a _pd.DataFrame_ vs. migrating to _pd.Index_/_pd.MultiIndex_? 

Currently, _Triangle.index_ returns a DataFrame across the public API (which estimators and _TriangleGroupBy_ depend on for column-based slicing and `key_labels`). We could store _self.index_ directly as a DataFrame (and use _self.index.values_ in place of `_kdims`), or if we want _Triangle.index_ to become a _MultiIndex_, that might be a broader public API change worth tackling in a dedicated PR as part of #601.

In the meantime, I will proceed with updating `columns` to be stored as a `pd.Index` and storing `index` directly.

genedan:
> For Triangle.index, what are your thoughts on keeping it stored as a pd.DataFrame vs. migrating to pd.Index/pd.MultiIndex?

For 0.11.0, we should retain it as a DataFrame, otherwise it'll lead to a breaking change. I think we should switch to Index/MultiIndex for 1.0.0 - you could make a new issue for that if there isn't one already. I'm not sure if there's a good way to notify the public of this change other than to have it be part of a major version bump. It would be too annoying to warn a user every time they call the index.

Anyhow, let me know when you're ready for a review.

genedan:
I also don't think the `dims` need to go through a deprecation style, I think we can go straight to removing them. They were never meant to be part of the public API and aren't mentioned in the docs. @casact/chainladder-maintainers, let me know if you disagree.

priyam0k:
> For 0.11.0, we should retain it as a DataFrame, otherwise it'll lead to a breaking change. I think we should switch to Index/MultiIndex for 1.0.0

Makes total sense. Keeping `Triangle.index` as a DataFrame for 0.11.0 avoids breaking anything downstream, and we can look at the Index/MultiIndex shift for 1.0.0 (maybe track that under #601).

> I also don't think the dims need to go through a deprecation style, I think we can go straight to removing them.

Agreed. removal is much cleaner since they were never part of the public API, and it saves us having to maintain deprecation shims for internal attributes. 

I am just clearing up a couple of CI checks on this branch (Read the Docs and coverage) and will ping you here as soon as it is ready for your review.

priyam0k:
All checks passed. @genedan this is ready for review.

chainladder/adjustments/bootstrap.py
```diff
@@ -272,17 +267,13 @@ def _get_simulation(self, X, exp_incr_triangle):
         resampled_triangles = resampled_triangles[None, ...].swapaxes(0, 1)
         obj = X.copy()
         if X.key_labels == ["Total"]:
-            obj.kdims = np.arange(self.n_sims)
+            obj._index = pd.DataFrame({"Simulation_#": np.arange(self.n_sims)})
```
genedan, line 270: We should avoid accessing _index directly.

chainladder/core/tests/test_triangle.py
```diff
@@ -438,17 +437,154 @@ def test_set_index_not_inplace(clrd: Triangle) -> None:
...
     assert result is not tri
     assert result.key_labels == ["Company"]
-    np.testing.assert_array_equal(result.kdims, new_index.values)
+    assert result.index.equals(new_index)
     assert tri.key_labels == original_key_labels
-    np.testing.assert_array_equal(tri.kdims, original_kdims)
+    assert tri.index.equals(original_index)
+
+
+def test_vdims_deprecation_warning(raa):
```
genedan, line 453: We can remove the tests for deprecation warnings since we'll be removing the dims without going through the deprecation cycle.

chainladder/core/triangle.py
```diff
@@ -724,30 +724,104 @@ def index(self) -> DataFrame:
...
+    def key_labels(self, value) -> None:
+        if isinstance(value, str):
+            value = [value]
+        else:
+            value = list(value)
+        self._index.columns = value
+        self._set_slicers()
+
+    @property
+    def kdims(self):
```
genedan, line 755: We can remove the dims properties.

chainladder/core/triangle.py
```diff
@@ -558,23 +558,23 @@ def __init__(
             self.index_label: list = index
             data_agg[index[0]] = "Total"
 
-        self.kdims: np.ndarray
+        self._index: DataFrame
```
genedan, line 561: This block will be key to refactoring the dimensions - I'll sketch out a plan to achieve our goals.

chainladder/core/triangle.py
```diff
@@ -558,23 +558,23 @@ def __init__(
...
-        self.vdims: np.ndarray
+        self._columns: pd.Index
         self.odims: np.ndarray
         orig_idx: np.ndarray
         self.ddims: ArrayLike
         dev_idx: np.ndarray
 
-        self.kdims, key_idx = self._set_kdims(data_agg, index)
-        self.vdims = np.array(columns)
+        kdims_arr, key_idx = self._set_kdims(data_agg, index)
```
genedan, line 569: looks like the _set_kdims method still exists, we should at least rename it to `_set_index` (assuming we keep the function).

genedan:
I've submitted some comments, but consider it a partial review to help guide you where we need to go. This is one of the larger issues to close out, so it will most likely get it take a lot of back and forth between us until we can merge it into main (more commits credited to you!). We'll have to take it step-by-step, but I think we can get it done. Some general goals:

- The need to define internal variables `_index` and `_columns` makes me think that our current definition of of `Triangle.index` and `Triangle.columns` is lacking in some way, especially when you find the need to reference those protected variables outside the `Triangle` class. Once defined, we should be able to access what we need from `Triangle.index` and `Triangle.columns` without the need to reference those protected variables. This is one of the driving motivations of the refactor.
- This leads me to believe we should get started on defining a generalized `TriangleAxis` class from which `Triangle.index`, `Triangle.columns`, `Triangle.origin`,  and `Triangle.development` You can reference this comment for the general idea that I had: https://github.com/casact/chainladder-python/issues/601#issuecomment-4743187859
- We can remove all vdims/kdims code, and no longer need to issue deprecation warnings.

Refactoring this block will be key to closing out the issue:

https://github.com/casact/chainladder-python/blob/6b51f649cf8056c2d2d225afbd846ac6796c9c11/chainladder/core/triangle.py#L561-L572

Ideally, we'd have something like:

```python
index = TriangleAxis("index", fset=_set_index)
columns = TriangleAxis("columns", fset=_set_columns)
origin = TriangleAxis("origin", fget=_get_origin, fset=_set_origin)
development = TriangleAxis("development", fget=_get_development, fset=_set_development)
```

I have a rough sketch of what `TriangleAxis` should look like

```python
  class TriangleAxis:
      """Generalized class for representing a Triangle dimension,
      analogous to Pandas AxisProperty."""

      def __init__(self, key, *, fget=None, fset=None, doc=None):
          self.key = key
          self.fget = fget   # raw -> public; identity if None
          self.fset = fset   # (obj, public) -> raw; identity if None
          self.__doc__ = doc

      def __get__(self, obj, objtype=None):
          if obj is None:
              return self                                                                                                                                                                                                                                               
          raw = obj._axes[self.key]
          return self.fget(obj, raw) if self.fget else raw

      def __set__(self, obj, value):
          raw = self.fset(obj, value) if self.fset else value
          obj._axes[self.key] = raw
          obj._set_slicers()
```

With these helper functions:

```python

  def _set_index(obj, value):
      if not isinstance(value, pd.DataFrame):
          raise TypeError("index must be a pandas DataFrame")
      obj._len_check(obj.index, value)
      return value.copy().reset_index(drop=True)

  def _set_columns(obj, value):
      if isinstance(value, str):
          value = [value]                                                                                                                                                                                                                                               
      obj._len_check(obj.columns, value)
      return pd.Index(value, name="columns")

  def _get_origin(obj, raw):
      if obj.is_pattern and len(raw) == 1:
          return pd.Series(["(All)"])
      freq = {"S": "2Q", "H": "2Q"}.get(obj.origin_grain, obj.origin_grain)
      freq = freq if freq == "M" else freq + "-" + obj.origin_close
      return pd.DatetimeIndex(raw, name="origin").to_period(freq=freq)

  def _set_origin(obj, value):
      obj._len_check(obj.origin, value)
      freq = {"S": "2Q"}.get(obj.origin_grain, obj.origin_grain)
      freq = freq if freq == "M" else freq + "-" + obj.origin_close
      return pd.PeriodIndex(list(value), freq=freq).to_timestamp().values
```

You may need to tweak these and this might take trial and error from the both of us. How about we give it a go, create a new `axis.py` file to place the class in, and see if we can refactor the code for one dimension, `Triangle.columns`?

priyam0k:
Thanks for the thorough review and guidance,

> We can remove all vdims/kdims code, and no longer need to issue deprecation warnings.

Completely agree. Removing `kdims` and `vdims` directly keeps things much cleaner and avoids carrying deprecation shims for internal attributes that were never documented or public. I'll drop those properties and their deprecation tests, and rename `_set_kdims` to `_set_index`.

> The need to define internal variables `_index` and `_columns` makes me think that our current definition of `Triangle.index` and `Triangle.columns` is lacking in some way, especially when you find the need to reference those protected variables outside the `Triangle` class.

Spot on. Reaching into `_index` in `bootstrap.py` was a workaround for the length check during simulation reshaping; I will refactor that to use the public API instead.

> How about we give it a go, create a new `axis.py` file to place the class in, and see if we can refactor the code for one dimension, `Triangle.columns`?

Nice plan. I really like the `TriangleAxis` descriptor design. I'll set up `chainladder/core/axis.py`, hook up `Triangle.columns` using `TriangleAxis`, and make sure tests pass cleanly.

Will push an update shortly so we can take a look..

genedan:
Whew, looks like the PR is still intact. This was definitely one of the more significant ones that we wanted in.

image

priyam0k:
I'm back, really back. Thanos snap
image

priyam0k:
@genedan I pushed an update with the first step of the descriptor refactor.

The new `TriangleAxis` descriptor is in [axis.py](https://github.com/casact/chainladder-python/blob/feature/deprecate-dims/chainladder/core/axis.py), and I hooked up columns to use it in [triangle.py](https://github.com/casact/chainladder-python/blob/feature/deprecate-dims/chainladder/core/triangle.py#L430). As discussed, I also dropped `kdims` and `vdims` directly, and cleaned up [bootstrap.py](https://github.com/casact/chainladder-python/blob/feature/deprecate-dims/chainladder/adjustments/bootstrap.py#L269-L275) so it sets the index publicly instead of touching private variables. I also safeguarded `sort_index` against mutating the live index.

Take a look when free. If this shape works I'll follow up with index next.

chainladder/core/triangle.py
```diff
@@ -23,7 +24,7 @@
 except ImportError:
     db = None
 
-from typing import cast, Optional, TYPE_CHECKING
+from typing import Any, cast, Optional, TYPE_CHECKING
```
genedan: Let's put these on multiple lines.

chainladder/core/axis.py
```diff
@@ -0,0 +1,57 @@
...
+
+from typing import TYPE_CHECKING, Any, Callable
+import pandas as pd
+
+if TYPE_CHECKING:
+    from chainladder.core.triangle import Triangle
+
+
+class TriangleAxis:
+    """
```
genedan, line 13: Add docstrings for:

- The class (add parameters)
- Methods

chainladder/core/axis.py
```diff
@@ -0,0 +1,57 @@
...
+    analogous to Pandas AxisProperty.
+    """
+
+    def __init__(
+        self,
+        key: str,
+        *,
+        fget: Callable[[Triangle, Any], Any] | None = None,
+        fset: Callable[[Triangle, Any], Any] | None = None,
+        doc: str | None = None,
```
genedan, line 38: I think this doc parameter might be extraneous. Compare these two outputs:

```python
import chainladder as cl
import pandas as pd

pd.DataFrame().columns.__doc__

raa = cl.load_sample('raa')
raa.columns.__doc__
```

The both resolve to the doc for an Index object, which is right. The `raa.columns.__doc__` mentions Pandas though, I wonder if there's an easy way to replace the word "pandas" with "chainladder" without too much engineering?

Seeing "pandas" in the doc probably shouldn't be a big deal for now though, let's put it low on the priority list of changes but nice to have if you can get around to it.

chainladder/core/triangle.py
```diff
@@ -442,6 +449,7 @@ def __init__(
         *args,
         **kwargs,
     ):
+        self._axes: dict[str, Any] = {}
```
genedan, line 457: We should turn this into a public property that returns a list of the access. Doing so would match the Pandas signature:

```python
    @property
    def axes(self) -> list[Index]:
        """
        Return a list representing the axes of the DataFrame.
```

chainladder/core/triangle.py
```diff
@@ -2040,10 +2067,38 @@ def copy(self):
...
         X = object.__new__(self.__class__)
         X.__dict__.update(vars(self))
+        if hasattr(self, "_axes"):
+            X._axes = {k: v.copy() for k, v in self._axes.items()}
         X._set_slicers()
         X.values = X.values.copy()
+        X._index = self._index.copy()
         return X
 
+    def __setstate__(self, state: dict) -> None:
```
genedan, line 2075: Will this block still be necessary once you're done with the PR? A passing test at the end of step 2 of this PR is enough to convince me that we haven't broken anything. We can remove this block and the associated test in a new PR step 3.

chainladder/core/triangle.py
```diff
@@ -721,30 +728,50 @@ def index(self) -> DataFrame:
...
+    @key_labels.setter
+    def key_labels(self, value) -> None:
+        if isinstance(value, str):
+            value = [value]
+        else:
+            value = list(value)
+        self._index.columns = value
         self._set_slicers()
 
+    @property
```
genedan, line 766: Let's work towards removing _columns.

chainladder/core/dunders.py
```diff
@@ -319,10 +318,9 @@ def __add__(self, other):
         if isinstance(obj, TriangleGroupBy):
 
             def f(k, self, obj, other):
-                # fmt: off
```
genedan, line 322: Henry intentionally set #fmt: off (I think for readability), could you revert?

chainladder/core/dunders.py
```diff
@@ -369,10 +367,9 @@ def __sub__(self, other):
...
```
genedan, line 372: Let's revert here, looks like an intentional code style from Henry.

chainladder/core/dunders.py
```diff
@@ -448,10 +445,9 @@ def __mul__(self, other):
...
```
genedan, line 451: revert

chainladder/core/dunders.py
```diff
@@ -466,10 +462,9 @@ def __pow__(self, other):
...
```
genedan, line 469: revert

chainladder/core/dunders.py
```diff
@@ -540,10 +535,9 @@ def __truediv__(self, other: Any):
...
```
genedan, line 543: revert

chainladder/development/clark.py
```diff
@@ -329,10 +329,16 @@ def ldf(age):
         self.ldf_ = obj
         self.ldf_.valuation_date = pd.to_datetime(options.ULT_VAL)
         rows = X.index.set_index(X.key_labels).index
-        self.omega_ = pd.DataFrame(params[..., 0, 0], index=rows, columns=X.vdims)
-        self.theta_ = pd.DataFrame(params[..., 0, 1], index=rows, columns=X.vdims)
+        self.omega_ = pd.DataFrame(
```
genedan, line 332: Do we need to use columns_label here, or can we extract it from the new columns index?

chainladder/tails/bondy.py
```diff
@@ -192,9 +192,9 @@ def fit(self, X, y=None, sample_weight=None):
         )
         fitted = xp.repeat(fitted, self.ldf_.shape[2], axis=2)
         rows = X.index.set_index(X.key_labels).index
-        self.b_ = pd.DataFrame(self.b_[..., 0, 0], index=rows, columns=X.vdims)
+        self.b_ = pd.DataFrame(self.b_[..., 0, 0], index=rows, columns=X.columns_label)
```
genedan: Same as the other comment, can we extract from columns index?

chainladder/tails/curve.py
```diff
@@ -290,13 +289,13 @@ def slope_(self):
         """Does not work with munich"""
         rows = self.ldf_.index.set_index(self.ldf_.key_labels).index
         return pd.DataFrame(
-            self._slope_[..., 0, 0], index=rows, columns=self.ldf_.vdims
+            self._slope_[..., 0, 0], index=rows, columns=self.ldf_.columns_label
```
genedan: extract from columns index?

genedan (commented):
This is starting to look way cleaner, good work. I left some comments requesting some changes, and I asked a few questions too.

This is going to be an awesome PR!

chainladder/core/triangle.py
priyam0k: Done, split them across multiple lines.

chainladder/core/axis.py
priyam0k: Added docstrings for the class parameters and all methods.

chainladder/core/axis.py
priyam0k: Agree. When accessing .columns.__doc__ on an instance it delegates straight to pandas Index, which is why pandas shows up there. I kept doc on the descriptor for now so the attribute has documentation on the class level, but agreed we can keep custom doc tweaking low priority.

chainladder/core/triangle.py
priyam0k: Bugbot flagged that earlier because older pickles saved with kdims/vdims would fail to load without it. If backward compatibility for pre-refactor pickles isn't required for this release, I'm completely onboard with removing it and the test in step 3 like you suggested.

chainladder/core/triangle.py
priyam0k: Removed _columns completely now that `columns` is cleanly handled by TriangleAxis.

chainladder/core/dunders.py
priyam0k: Reverted. Restored all the # fmt: off / # fmt: on blocks across the arithmetic methods to keep Henry's layout intact.

chainladder/core/triangle.py
priyam0k: Added the public axes property returning all 4 dimensions, plus a unit test for it.

chainladder/core/triangle.py
priyam0k: Left this for now since we focused on `columns` first. I'll clean it up when we tackle `index` next (and avoid stepping on the existing _set_index helper in base.py).

chainladder/development/clark.py
priyam0k: Switched to columns=X.columns.to_list() directly.

chainladder/tails/bondy.py
priyam0k: Updated to extract directly via .to_list() here too.

chainladder/tails/curve.py
priyam0k: Updated to extract directly via .to_list() here too.

priyam0k (commented):
Addressed all the comments. Let me know if this looks good to move ahead with index

chainladder/core/axis.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+class TriangleAxis:
+    """
+    Generalized descriptor for representing a Triangle dimension,
+    analogous to Pandas AxisProperty.
+
+    Parameters
+    ----------
+    key : str
```
genedan: Could we change the str to int? We might be able to knock out #1053 as well.

chainladder/core/axis.py
genedan: k, I'll create a new issue so we can focus on the big picture stuff.

chainladder/core/triangle.py
genedan: Got it. Looks like the remaining spots we use `self._axes` require the rest of the axes to be completed, then we can remove it.

chainladder/core/triangle.py
genedan: Now I understand what this is for. I suppose we'll need to leave it. Let's add a warning telling users that a legacy pickle was detected and that they should upgrade to chainladder 0.11.0. Once we resolve our discussion in #1027 we can determine what version we'll remove `__setstate__` in.

genedan:
The main outstanding changes are to:
- Set the key on the TriangleAxis to be int
- Work towards removing `Triangle._axis`

Doing the second part is somewhat involved, I sketched out the steps in #1053. I'll let you decide whether you think it's doable within this PR or whether you'd like to move on to working on the other dimensions first, or create a 3rd PR step, etc.

Issue #1037, opened 2026-06-21
[API] Promote `MethodBase.intersection`


henrydingliu:
### Description

`MethodBase.intersection` is a useful method that should be under `utils`, rather than being tucked under `MethodBase`. 

### Is your feature request aligned with the scope of the package?

- [x] Yes, absolutely!
- [ ] No, but it's still worth discussing.
- [ ] N/A (this request is not a codebase enhancement).

### Would you be willing to contribute this ticket?

- [ ] Yes, absolutely!
- [ ] Yes, but I would like some help.
- [ ] No.

Issue #400, opened 2022-12-03
[BUG] CapeCod predict at different index grain


jbogaardt:
**Describe the bug**
The chainladder estimators are supposed to broadcast LDFs and Aprioris down do a lower grain. For example, one might fit countrywide patterns, but predict at a state level. For BornheutterFerguson and Chainladder this works.
```python
import chainladder as cl
prism = cl.load_sample('prism')[['reportedCount', 'Paid']]

bf_pipe = cl.Pipeline(
    [('dev', cl.Development()),
     ('model', cl.BornhuetterFerguson())]
)
# Fit at Line level
bf_pipe.fit(
    X=prism.groupby('Line')['Paid'].sum(), 
    sample_weight=prism.groupby('Line')['reportedCount'].sum().sum('development'))

# Predict at claim level
bf_pipe.predict(
    X=prism['Paid'], 
    sample_weight=prism['reportedCount'].sum('development')
)
```

**To Reproduce**
The same code errors out when using the `CapeCod` estimator:
```python
import chainladder as cl
prism = cl.load_sample('prism')[['reportedCount', 'Paid']]
# CapeCod Fails
cc_pipe = cl.Pipeline(
    [('dev', cl.Development()),
     ('model', cl.CapeCod())]
)
# Fit at Line level
cc_pipe.fit(
    X=prism.groupby('Line')['Paid'].sum(), 
    sample_weight=prism.groupby('Line')['reportedCount'].sum().sum('development'))
# Predict at claim level
cc_pipe.predict(
    X=prism['Paid'], 
    sample_weight=prism['reportedCount'].sum('development'))
```

**Expected behavior**
I expect the `CapeCod` code not to crash.

jbogaardt:
This is strictly a `predict` problem. Avoiding `predict` and using the `groupby` parameter in each esitmator, I can get a valid answer. 

```python
import chainladder as cl
prism = cl.load_sample('prism')[['reportedCount', 'Paid']]

# CapeCod Fails
cc_pipe = cl.Pipeline(
    [('dev', cl.Development(groupby='Line')),
     ('model', cl.CapeCod(groupby='Line'))]
)
cc_pipe.fit(
    X=prism['Paid'], 
    sample_weight=prism['reportedCount'].sum('development'))

print(cc_pipe.named_steps.model.ldf_.shape) # LDFs at Line level
print(cc_pipe.named_steps.model.apriori_.shape) # Apriori at line level
[...]

PR #1310, opened 2026-09-07
[FIX] raise when the fitted pattern cannot be applied to the triangle passed in

### LINKED ISSUE #1288: Chainladder.predict changes the shape of the result without saying so, where BornhuetterFerguson and CapeCod raise
Split out of #1274, where @henrydingliu asked whether the grain broadcasting is robust outside `CapeCod`. Two cases in `Chainladder.predict` change the shape of the result without saying so. `BornhuetterFerguson` and `CapeCod` raise on both.

### 1. A group the model never saw is silently dropped

```python
import chainladder as cl

clrd = cl.load_sample("clrd")
tri = clrd["CumPaidLoss"]
train = tri[tri.index["LOB"] != "wkcomp"].groupby("LOB").sum()

model = cl.Chainladder().fit(cl.Development().fit_transform(train))
pred = model.predict(tri)

print(tri.shape[0], "->", pred.ultimate_.shape[0])   # 775 -> 643
```

No warning. The 643 rows returned are exactly the companies in the five LOBs the model saw, and the 132 dropped are exactly the held-out LOB. Anything joined to `ultimate_` afterwards has a hole with no signal that it is there.

`BornhuetterFerguson.predict` and `CapeCod.predict` on the same inputs raise `ValueError: operands could not be broadcast together with shapes (775,1,10,1) (643,1,10,1)`.

### 2. Predicting at a coarser grain returns the fitted grain

```python
import chainladder as cl

tri = cl.load_sample("clrd")["CumPaidLoss"]

model = cl.Chainladder().fit(cl.Development().fit_transform(tri))
pred = model.predict(tri.groupby("LOB").sum())

print(pred.ultimate_.shape[0])          # 775, from a 6-row input
print(pred.ultimate_.key_labels)        # ['GRNAME', 'LOB'], caller passed ['LOB']
```

No warning. The returned object carries the model's index rather than the caller's, and the values are not the fitted ultimates either. The same call on `BornhuetterFerguson` and `CapeCod` raises `ValueError`.

### Where both come from

One line, `self.intersection(X_new, X_new.ldf_)` in `MethodBase.predict` (`base.py:111`), with two failure shapes:

```
case 1
  after val_to_dev                       (775, 1, 10, 10)  ['GRNAME', 'LOB']
  after intersection()                   (643, 1, 10, 10)  ['GRNAME', 'LOB']

case 2
  ldf_ after fit                         (775, 1, 1, 11)   ['GRNAME', 'LOB']
  after intersection(): the triangle     (6, 1, 10, 10)    ['LOB']
  after intersection(): its ldf_         (775, 1, 1, 11)   ['GRNAME', 'LOB']
```

In case 1 it narrows both operands to their common index values, so companies whose LOB the model never saw fall out of data the caller passed in. Everything before that line still has all 775 rows.

In case 2 it narrows only one. The caller's six LOB values already cover every LOB present in the 775-row `ldf_`, so nothing is dropped on that side, and `predict()` returns an object whose triangle is at the caller's grain and whose `ldf_` is at the model's. `Chainladder._get_ultimate` consumes that `ldf_`, which is where the 775-row `ultimate_` comes from.

### What I think it should do

The narrower point, and the one I would argue for regardless of the rest: the three methods disagree about how to fail. `BornhuetterFerguson` and `CapeCod` refuse both of these loudly, `Chainladder` does neither. Whatever the intended contract is, one of the three is wrong.

If a grain mismatch is meant to be rejected, `Chainladder` should raise too, ideally with a message naming the offending index values rather than a numpy shape error. If narrowing to the intersection is the intended contract for case 1, a warning saying how many rows were dropped would at least make it visible. For case 2 I am not arguing that predicting at a coarser grain should work, only that returning `ultimate_` at a grain the caller never asked for, silently, is the wrong way to decline.

Environment: a fresh clone of `main` at 8a47051b, Python 3.11, the `clrd` sample. Everything above was run there, not on a branch.

ppcvote:
## Summary of Changes

`intersection()` narrows two triangles to their shared index. That is the right thing for the grain broadcast #400 asked for, and the wrong thing for a mismatch, and it cannot tell the two apart. Three mismatches were resolving quietly:

- **`X` carries a group the model was never fit on.** `intersection` drops those rows, so `Chainladder` returned 643 of the 775 rows it was handed with no warning, while `BornhuetterFerguson` and `CapeCod` raised `operands could not be broadcast together with shapes (775,1,10,1) (643,1,10,1)`, which names no index value.
- **`ldf_` is at a finer grain than `X`.** `intersection` narrows neither side, so `predict` returns an object whose triangle sits at the caller's grain and whose `ldf_` sits at the model's, and `Chainladder` built a 775 row `ultimate_` from a 6 row input.
- **Both sides are a single row.** `intersection` short circuits on that at its first line, so a pattern fit on one group was applied to another with no warning. Fitting on `State Farm Mut Grp` comauto and predicting on `State Farm Mut Grp` othliab returns an ultimate of 1,638,150.65 carrying `[['State Farm Mut Grp', 'othliab']]` as its index, where othliab's own pattern gives 2,640,829.49. Which side's index the result carries depends on whether the two triangles share key labels: where they share at least one it is the group predicted on, and where they share none the addition below takes the fitted triangle's labels instead.

`validate_ldf` checks all three, in the shape `validate_weight` already uses.

Two things about it are worth flagging, because neither is obvious from the diff.

**Where the call sits.** It runs before `X_new = X_new + (self.X_.val_to_dev().iloc[0,0].sum(2) * 0)`, not next to `intersection`. That addition goes through `_prep_index`, which borrows `self.X_`'s index when both sides are a single row, so by the time control reaches `intersection` the caller's identity is already gone and the check sees two triangles that agree. Placing the call after that line looks natural and silently does nothing for the single-row case.

**The `"(All)"` exemption.** A pattern whose index is entirely the `"(All)"` sentinel came from `Triangle.sum` and carries no group identity, so neither its index values nor its key labels constrain what it may be applied to, and it is exempt from both checks. Without that, this rejects `test_different_backends`, which fits on `clrd.sum()` and predicts on 132 companies. Row count cannot separate that from the single-group mistake above: both are one row sharing no index value with the target. The limit of that exemption is worth stating plainly, because it is wider than it first looks. `Triangle.sum` stamps `"(All)"` on whatever subset it was called on, so a pattern summed from a single line of business is exempt exactly as a pattern summed from the whole book is. Fitting on `tri[tri.index["LOB"] == "wkcomp"].sum()` and predicting on comauto is allowed and returns 157 rows, which is the third case above reached with one extra `.sum()`. Closing that means recording aggregation provenance on the Triangle, which is a larger change than this and your call rather than mine. What this PR does is remove the cases where nothing was summed at all.

A pattern coarser than `X` stays allowed, which is the #400 flow.

**Columns, added after review.** The same mismatch on the columns axis: fitting on `CumPaidLoss` and predicting on `IncurLoss` returned an ultimate of 228,088,946 labelled `IncurLoss`, where fitting on the incurred triangle gives 150,105,776. `paid -> [paid, incurred]` was allowed too. The check is one direction, `set(X.columns) - set(ldf.columns)`; the reverse is noted on the thread.

## Related GitHub Issue(s)

Fixes #1288. `intersection` itself is untouched, per your point on that thread about #1037.

Note for merge order: #1306 touches the adjacent lines in `CapeCod.predict`, so whichever lands second will want a rebase.

- Seven tests, placed next to `test_misaligned_index2`. Those two cover predicting on a strict subset of the fitted data, which is the case this check is most likely to reject by mistake, so the boundary sits in one place. Removing `validate_ldf` from both call sites fails exactly the five rejection tests; the two that assert a flow still works pass either way, which is what they are for. `test_predict_checks_before_the_index_is_borrowed` fails on its own if the call is moved back after the addition, so the ordering above is held by a test rather than by a comment.
- `pytest chainladder` passes: 1156 passed, 7 skipped, against 1142 and 7 on `main`. The fourteen are the seven new tests across both backend parametrisations; no existing test changes behaviour.
- `ruff check --force-exclude --config lint.per-file-ignores={}` reports the same 15 findings on these three files before and after, and `ruff format --check` reports the same hunks line for line.
- `CapeCod` calls `validate_ldf` as well. For the finer-`ldf_` case that call is the only thing between the caller and a raw numpy error inside `_get_capecod_aprioris`, which runs before `MethodBase.predict` is reached. For the unseen-group case `MethodBase.predict` would catch it anyway through `super().predict`, just after CapeCod has done work it did not need to.
- One deliberate deviation: `validate_weight` next door uses a multi-line signature and `'''` docstrings, and `ruff format` wants neither. I wrote `validate_ldf` the way the formatter wants so it adds nothing to the format diff, rather than matching its neighbour.

## Declarations

I am adhering to the standards in the Governing Doc. I am a human contributor, not a bot, and I opened this because @henrydingliu asked for it on #1288.

AI disclosure, per the AI Usage Policy: I used Claude Code on this. It ran the reproductions across the estimators and all three directions, and found both the single-row case and the call-site ordering by running the suite against earlier attempts that were wrong in each of those ways. I reviewed the diff and the tests myself, and the suite was run locally on my machine.

### DIFF
--- a/chainladder/methods/base.py
+++ b/chainladder/methods/base.py
@@ -102,6 +102,9 @@ class MethodBase(BaseEstimator, EstimatorIO, Common):
         X_new = X.val_to_dev()
         if sum(X_new.ddims > self.ldf_.ddims.max()) > 0:
             raise ValueError("X has ages that exceed those available in model.")
+        # Before the line below, which borrows self.X_'s index when both sides
+        # are a single row and so would erase what the caller actually passed.
+        self.validate_ldf(X_new, self.ldf_)
         X_new = X_new + (self.X_.val_to_dev().iloc[0, 0].sum(2) * 0)
         self.validate_weight(X_new, sample_weight)
         if sample_weight:
@@ -154,6 +157,45 @@ class MethodBase(BaseEstimator, EstimatorIO, Common):
         else:
             process_var = None
         return process_var
+
+    @staticmethod
+    def validate_ldf(X: Triangle, ldf: Triangle) -> None:
+        """
+        Checks that a fitted pattern can be applied to X as it was passed in.
+        The index and the columns of the two have to line up: values or columns
+        X carries that the pattern does not cannot be predicted, and index
+        levels the pattern carries that X does not cannot be applied.
+        """
+        # A pattern whose index is entirely the "(All)" sentinel that Triangle.sum
+        # sets carries no group identity, so nothing about it constrains what it
+        # may be applied to. Note the limit of that: sum() stamps "(All)" on
+        # whatever subset it was called on, so a pattern summed from one line of
+        # business is exempt here just as a pattern summed from everything is.
+        # Telling those apart needs aggregation provenance on the Triangle.
+        if len(ldf) == 1 and set(ldf.index.values.flatten()) == {"(All)"}:
+            return
+        shared = sorted(set(X.key_labels) & set(ldf.key_labels))
+        if shared:
+            missing = sorted(
+                set(X.index.set_index(shared).index)
+                - set(ldf.index.set_index(shared).index)
+            )
+            if missing:
+                raise ValueError(
+                    "X has index values the model was not fit on: "
+                    + str(missing[:5])
+                    + (", and others" if len(missing) > 5 else "")
+                )
+        columns = sorted(set(X.columns) - set(ldf.columns))
+        if columns:
+            raise ValueError("X has columns the model was not fit on: " + str(columns))
+        finer = sorted(set(ldf.key_labels) - set(X.key_labels))
+        if finer:
+            raise ValueError(
+                "The fitted pattern has index levels that X does not: "
+                + str(finer)
+                + ". It cannot be applied to a triangle that does not carry them."
+            )
 
     @staticmethod
     def validate_weight(
--- a/chainladder/methods/capecod.py
+++ b/chainladder/methods/capecod.py
@@ -319,6 +319,7 @@ class CapeCod(Benktander):
         if sample_weight is None:
             raise ValueError("sample_weight is required.")
         X_new = X.copy()
+        self.validate_ldf(X_new, self.ldf_)
         _, X_new.ldf_ = self.intersection(X_new, self.ldf_)
         # If model was fit at a higher grain, then need to aggregate predicted aprioris too
         if len(set(sample_weight.key_labels) - set(self.apriori_.key_labels)) > 0:
--- a/chainladder/methods/tests/test_predict.py
+++ b/chainladder/methods/tests/test_predict.py
@@ -190,6 +190,168 @@ def test_misaligned_index2(clrd):
     assert abs(a - b) < 1e-5
 
 
+def test_predict_rejects_index_values_the_model_never_saw(clrd):
+    """github issue #1288
+
+    intersection() narrows both operands to their shared index, so a group the
+    model was never fit on used to be dropped from the result instead of
+    predicted, silently on Chainladder and as a numpy shape error on the others.
+    """
+    tri = clrd["CumPaidLoss"]
+    sample_weight = clrd["EarnedPremDIR"].latest_diagonal
+    seen = tri.index["LOB"] != "wkcomp"
+    train = tri[seen].groupby("LOB").sum()
+    train_weight = sample_weight[seen].groupby("LOB").sum()
+
+    models = [
+        cl.Chainladder().fit(cl.Development().fit_transform(train)),
+        cl.BornhuetterFerguson(apriori=0.7).fit(train, sample_weight=train_weight),
+        cl.CapeCod().fit(train, sample_weight=train_weight),
+    ]
+    for model in models:
+        with pytest.raises(ValueError, match="index values the model was not fit on"):
+            if isinstance(model, cl.Chainladder):
+                model.predict(tri)
+            else:
+                model.predict(tri, sample_weight=sample_weight)
+
+
+def test_predict_rejects_a_different_single_group(clrd):
+    """github issue #1288
+
+    Both sides being one row is the case intersection() short circuits on, so
+    the pattern from one group used to be applied to another silently, with the
+    result labelled as the group that was predicted on.
+    """
+    lob = clrd["CumPaidLoss"].groupby("LOB").sum()
+    fitted_on = lob[lob.index["LOB"] == "comauto"]
+    predicted_on = lob[lob.index["LOB"] == "othliab"]
+    assert len(fitted_on) == len(predicted_on) == 1
+
+    weight = clrd["EarnedPremDIR"].latest_diagonal.groupby("LOB").sum()
+    fitted_weight = weight[weight.index["LOB"] == "comauto"]
+    predicted_weight = weight[weight.index["LOB"] == "othliab"]
+
+    model = cl.Chainladder().fit(cl.Development().fit_transform(fitted_on))
+    with pytest.raises(ValueError, match="index values the model was not fit on"):
+        model.predict(predicted_on)
+    # the same single group is still fine
+    assert model.predict(fitted_on).ultimate_.shape[0] == 1
+
+    for cls in (cl.BornhuetterFerguson, cl.CapeCod):
+        exposure_model = cls().fit(fitted_on, sample_weight=fitted_weight)
+        with pytest.raises(ValueError, match="index values the model was not fit on"):
+            exposure_model.predict(predicted_on, sample_weight=predicted_weight)
+
+
+def test_predict_rejects_a_pattern_finer_than_the_triangle(clrd):
+    """github issue #1288
+
+    The reverse direction: a pattern fit per company cannot be applied to a
+    triangle that has aggregated companies away. intersection() leaves the ldf_
+    at the fitted grain, which used to surface as a 775 row ultimate_ from a
+    6 row input.
+    """
+    tri = clrd["CumPaidLoss"]
+    sample_weight = clrd["EarnedPremDIR"].latest_diagonal
+    coarse = tri.groupby("LOB").sum()
+    coarse_weight = sample_weight.groupby("LOB").sum()
+
+    models = [
+        cl.Chainladder().fit(cl.Development().fit_transform(tri)),
+        cl.BornhuetterFerguson(apriori=0.7).fit(tri, sample_weight=sample_weight),
+        cl.CapeCod().fit(tri, sample_weight=sample_weight),
+    ]
+    for model in models:
+        with pytest.raises(ValueError, match="has index levels that X does not"):
+            if isinstance(model, cl.Chainladder):
+                model.predict(coarse)
+            else:
+                model.predict(coarse, sample_weight=coarse_weight)
+
+
+def test_predict_rejects_columns_the_model_was_not_fit_on(clrd):
+    """github issue #1288
+
+    The same mismatch on the columns axis. A paid pattern applied to an incurred
+    triangle used to come back labelled incurred, overstating the ultimate by
+    about half on clrd.
+    """
+    paid = clrd["CumPaidLoss"].groupby("LOB").sum()
+    incurred = clrd["IncurLoss"].groupby("LOB").sum()
+    both = clrd[["CumPaidLoss", "IncurLoss"]].groupby("LOB").sum()
+
+    model = cl.Chainladder().fit(cl.Development().fit_transform(paid))
+    with pytest.raises(ValueError, match="columns the model was not fit on"):
+        model.predict(incurred)
+    with pytest.raises(ValueError, match="columns the model was not fit on"):
+        model.predict(both)
+
+    # the column it was fit on is still fine
+    assert model.predict(paid).ultimate_.shape[0] == paid.shape[0]
+
+
+def test_predict_still_allows_an_aggregate_pattern(clrd):
+    """github issue #1288
+
+    A pattern fit on a fully aggregated triangle carries the "(All)" sentinel
+    rather than any group identity, so it may be broadcast to any grain. Both
+    directions matter: test_different_backends relies on the first.
+    """
+    tri = clrd["CumPaidLoss"]
+    sample_weight = clrd["EarnedPremDIR"].latest_diagonal
+    model = cl.BornhuetterFerguson().fit(tri.sum(), sample_weight=sample_weight.sum())
+
+    per_company = model.predict(tri, sample_weight=sample_weight)
+    assert per_company.ultimate_.shape[0] == tri.shape[0]
+
+    by_lob = model.predict(
+        tri.groupby("LOB").sum(), sample_weight=sample_weight.groupby("LOB").sum()
+    )
+    assert by_lob.ultimate_.shape[0] == tri.groupby("LOB").sum().shape[0]
+
+
+def test_predict_checks_before_the_index_is_borrowed(clrd):
+    """github issue #1288
+
+    MethodBase.predict adds a zeroed slice of self.X_ to X, and for single row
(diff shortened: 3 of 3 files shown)


henrydingliu:
thanks for the great PR! i have 3 very differently scoped comments

- regarding ruff errors, please hold on a couple of days while we land a repo-wide cleanup
- the conditions are pretty comprehensive for index matching. however, there's also matching on column values. i can create a new issue to log a future enhancement for it or we can add it into this PR. let me know
    - this wasn't an issue for sample_weight because sample_weight always has a different column value
- the warning themselves should be a bit more mechanic-agnostic, i.e. just tell the user that key label/value doesn't match; no need to mention intersection.

ppcvote:
All three done or answered. I also refreshed the description above, which still carried the counts from before the rebase and the column check.

**ruff.** Holding until the cleanup lands.

**Messages.** Reworded, no mechanism named:

```
X has index values the model was not fit on: ['wkcomp']
The fitted pattern has index levels that X does not: ['GRNAME']. It cannot be applied to a triangle that does not carry them.
X has columns the model was not fit on: ['IncurLoss']
```

**Columns.** Added here rather than deferred, because the existing suite turned out to be unaffected by it. The case is worse than the index ones. Without this PR's check:

```python
paid = clrd["CumPaidLoss"].groupby("LOB").sum()
incurred = clrd["IncurLoss"].groupby("LOB").sum()
cl.Chainladder().fit(cl.Development().fit_transform(paid)).predict(incurred)
```

returned an ultimate of 228,088,946 labelled `IncurLoss`, where fitting on the incurred triangle gives 150,105,776. A paid pattern on incurred data, overstated by half, with the right column name on it. `paid -> [paid, incurred]` was allowed too, broadcasting the one pattern across both.

What I can say for it is that the 1142 tests on `main` all still pass with the check in. What I cannot say is that no workflow outside this suite relies on the permissive behaviour, so if you would rather it were staged separately I will pull it back out.

The check is one direction, `set(X.columns) - set(ldf.columns)`. The reverse is real and I have left it alone: fitting on `[paid, incurred]` and predicting on `paid` is allowed and returns both columns, which is the column form of the finer-grain case already covered on the index side. I did probe rejecting it, with a positive control confirming the probe fires on `[paid, incurred] -> paid` and passes `paid -> paid`, and the full suite came back clean at 1156. So it looks addable, but it is a second decision rather than part of this one, and I would rather you made it.

Branch as it stands: 1156 passed, 7 skipped. `ruff check` reports the same 15 findings on these three files as `main` does, and `ruff format` the same hunks.

henrydingliu:
clean-up landed!

>returned an ultimate of 228,088,946 labelled IncurLoss, where fitting on the incurred triangle gives 150,105,776. A paid pattern on incurred data, overstated by half, with the right column name on it. paid -> [paid, incurred] was allowed too, broadcasting the one pattern across both.

i wasn't actually thinking about the single column case. if we were to check column value for single column triangle, it would be somewhat contrary to our approach on single index, no? let's lay out the different scenarios and align on the proper treatment

- single column fit, single column predict, column values match
    - all is good. 
- single column fit, single column predict, column values mismatch
    - i'm leaning towards letting this happen
- single column fit, multiple columns predict
    - given how we want the single index parallel to behave, do we also need to just let this happen?
- multiple columns fit, single/multiple columns predict
    - we want this to check for mismatch and fail if any predict columns fall outside of fit columns

ppcvote:
Rebased onto the cleanup. It had landed and touched all three files this PR does, so the branch went into conflict; separately, CI never ran on my previous push, which I cannot explain beyond noting it.

`ruff check` and `ruff format --check` are both clean on these files now, on the branch and on `main` alike, so the red that was sitting on this PR is gone rather than merely explained. CI on this head is 1157 passed, 7 skipped across the matrix, and removing `validate_ldf` from both call sites still fails exactly the five rejection tests and nothing else.

The one conflict was `iloc[0,0]` becoming `iloc[0, 0]` on the line my check sits above. Kept your formatting.

henrydingliu:
@ppcvote reminder we still have some outstanding questions on when to enforce mismatching column labels.

ppcvote:
Matrix below, run on this branch with `validate_ldf` swapped for a no-op to get the unchecked column, so both columns come from one run. `clrd` grouped by LOB throughout.

| scenario | unchecked | this PR |
|---|---|---|
| single fit, single predict, match (`paid -> paid`) | 152,598,500 | same |
| single fit, single predict, mismatch (`paid -> incurred`) | 228,088,946 under the label `IncurLoss`, against 150,105,776 fitting on incurred, +52% | raises |
| single fit, multiple predict (`paid -> [paid, incurred]`) | the paid pattern broadcast across both, `IncurLoss` comes back 228,088,946 | raises |
| multiple fit, predict outside the fit (`[paid, incurred] -> bulk`) | bulk data returned under the labels `CumPaidLoss` and `IncurLoss` | raises |

On whether checking a single column value cuts against the approach on single index: I do not think there is an index precedent to cut against, because before this PR the single row index case had the same defect the column case has. Unchecked, fitting `State Farm Mut Grp` comauto and predicting `State Farm Mut Grp` othliab returns 1,638,150.65 labelled othliab, where othliab's own pattern gives 2,640,829.49, 38% light, no warning. That is the behaviour on `main` today, and it is what `test_predict_rejects_a_different_single_group` describes in the past tense. So the index side is not the strict axis that columns would be departing from; both were permissive, and this PR as written closes both. The question is only whether you want them closed together. If the column mismatch stays open, columns become the looser of the two, which is a choice worth making deliberately rather than by inheritance. Anyone who means to put a paid pattern on incurred data can rename the column and thereby say so.

While rerunning these I found an error in my own PR description, third bullet. It quotes `[['Aegis Grp', 'comauto']]` as the resulting index, which is the fitted group, while the sentence before it says the result is labelled as the group predicted on. `Aegis Grp` comauto also carries a latest diagonal of 21, so it cannot produce the ultimate quoted beside it. I have replaced the example with the State Farm one above.

One precision note on that correction: the result takes the predicted side's label only when the two triangles share at least one key label. Where they share none, the addition that follows borrows the fitted side's labels instead. This PR already raises there, through the `finer` branch, so it needs no decision from you.

The list of scenarios is missing a fifth row, and it is the one I would most like your call on.

**Multiple columns fit, predict on a subset of them.** `[paid, incurred] -> paid` is allowed today. The triangle `predict()` hands back carries one column, `CumPaidLoss`, with a latest diagonal of 127,436,460. The second label appears only once `ultimate_` is computed: that same 127,436,460, the caller's paid data, is developed with incurred's pattern and returned as an `IncurLoss` ultimate of 124,566,047. Incurred's own diagonal is 155,110,733 and its own ultimate 150,105,776. Nothing falls outside the fit, so the rule in your fourth bullet does not reach it.

The index axis does not behave this way. Fit on six lines of business, predict on comauto alone, and you get one row, 8,192,755, equal to fitting comauto alone. `intersection()` narrows. Columns have no counterpart to it, which is why the subset case broadcasts instead of narrowing.

I prototyped the counterpart, six lines after `X_new.ldf_ = self.ldf_`, narrowing the pattern's columns to those `X` carries. `[paid, incurred] -> paid` then gives a single `CumPaidLoss` ultimate of 152,598,500, equal to fitting on paid alone, and `[paid, incurred] -> [paid, incurred]` is unchanged. Suite on the prototype: 1156 passed, 7 skipped, the same as this branch without it on the same machine.

So, your four plus the fifth, as I would set them:

- values match: allow. values mismatch: raise. single fit to multiple predict: raise. anything outside the fit: raise. Those four are what this PR does now.
- subset of the fitted columns: narrow, not raise.

Two calls for you. Whether you want the mismatch case permissive after all, in which case I pull that condition and its tests out. And whether the narrowing belongs here or in its own PR: I lean separate, since it changes the output of a flow that works today and this one is already carrying the index cases.

henrydingliu:
@ppcvote thanks for the detailed investigation! it's taken me a couple of days to digest and distill my thoughts. 

what i meant by single index precedent was the scenario where an index was not specified during Triangle construction, so a default value of "Total" gets assigned. Through this default, an ibnr predictor trained on an unindexed triangle will work on every other unindex triangle, even if we start to enforce strict index matching. 

i'm also trying to think through a more generalized ML lens. if i'm training a model on features A, B, C, and A is always a single value, ML best practice and many pipelines would just drop A out of the model. inference would then only be run on B and C, doesn't matter what other values come through A. 

both these considerations appear to align and so i'm arriving at this revised table. 

|path|train|predict|intended behavior|rationale|
|--|--|--|--|--|
|A|unindexed|unindexed|works|default from triangle construction|
|B|single column|single column|works|precedent from A + ML best practice|
|C|single column|multiple columns|broadcasts|if user can loop through individual columns and not get raised (due to B), we don't need to make their lives harder|
|D|multiple columns|predict⊆train|narrows|standard sklearn behavior|
|E|multiple columns|predict⊄train|raises|standard sklearn behavior (when handle_unknown is not set to "ignore")|

please let me know if my rationale makes sense. and narrows as a new PR sounds good.

ppcvote:
`raa`'s index values are `['Total']` with key labels `['Total']`, and that is a construction rule rather than a property of the sample: `Triangle.__init__` assigns `["Total"]` whenever no index is passed. That is the precedent I had not weighed.

It is worth separating from the one I argued against on the 20th, since both went by the same name. That one was two named groups colliding, fitting comauto and predicting othliab, and this PR still raises there; `test_predict_rejects_a_different_single_group` is that case. Yours is the opposite situation, where nobody named an index at all, so every unindexed triangle carries the same default label and a pattern fit on one already applies to the next.

The one place the analogy is loose is that nobody chooses `Total`, while selecting `CumPaidLoss` out of six columns is deliberate and named. I do not think that difference should decide the API, and the ML framing settles the direction for me: a feature with a single value carries nothing to match on. Once B is granted, C follows, because looping the columns would get there anyway.

Pushed. `validate_ldf` now applies the column check only when the fit has more than one column. Measured against your table, on `clrd` grouped by LOB:

| path | before | now |
|---|---|---|
| A `raa -> raa` | works | unchanged |
| B `paid -> incurred` | raised | works, `IncurLoss` at 228,088,946 |
| C `paid -> [paid, incurred]` | raised | broadcasts, 152,598,500 and 228,088,946 |
| D `[paid, incurred] -> paid` | allowed, two columns | unchanged, narrowing is the separate PR |
| E `[paid, incurred] -> bulk` | raised | unchanged |
| E `[paid, incurred] -> [paid, bulk]` | raised | unchanged |

The test that pinned B and C as rejections is now the one that pins them as permitted, alongside both E cases, and it fails on both parametrised cases with the check removed. Suite: 1156 passed, 7 skipped, the same as this branch before the change. `ruff check` and `ruff format --check` clean on both files.

Two things I did not do, so they are visible rather than assumed. The index axis is untouched: fitting one company and predicting another still raises, and the `"(All)"` exemption still stands. And D still returns the fitted width rather than narrowing, so `[paid, incurred] -> paid` comes back with an `IncurLoss` ultimate of 124,566,047 built from the caller's paid diagonal. I will open that as its own PR once this one lands, since it touches the same lines.

ppcvote:
`ruff check` went red on that push and is green again now. The cause is worth a line, because my local run had said clean.

The branch was still based on the commit it was opened from, so `ruff` locally was reading that `pyproject.toml`, while CI reads the one on `main`, which has since added `D213` and `BLE001`. All seven failures are `D213` on docstrings this PR itself added, so they were mine to catch and the stale ruleset is why I did not. Rebasing onto `main` reproduced all seven locally, `--fix` resolved them, and they sit in their own commit so the behaviour change stays readable on its own. `ruff` is pinned to the same version in both places, so the version was not a factor.

After the rebase, on the configured test paths: 1274 passed, 7 skipped, against 1260 and 7 on `main`. The fourteen are the seven tests across both backend parametrisations, as before. `ruff check` and `ruff format --check` clean on both files, now against the config CI actually uses.

PR #1316, opened 2026-09-09
Rename PTF_formula to ptf_formula with a deprecation alias

Abhayindia:
## Summary of Changes

Renames `PTF_formula` to `ptf_formula`. The old name stays as a thin wrapper that emits a `DeprecationWarning` and forwards to the new one, so nothing breaks for 0.11.0. The internal caller in `barnzehn.py` is switched to the new name.

## Related GitHub Issue(s)

Closes #1291

## Additional Context for Reviewers

Added `test_ptf_formula_deprecated_alias` — checks the alias warns and returns the same formula string. `ruff check --config lint.per-file-ignores={}` and `ruff format --check` pass on the three changed files; `pytest chainladder/development/tests/test_barnzehn.py` passes. Couldn't run the doctest builder locally (no `jb`), but no docs reference `PTF_formula`.

## Checklist
- [x] I passed tests locally for code changes (documentation unaffected)

Issue #1322, opened 2026-09-09
[FEAT] Add decorator to drop an argument


genedan:
When a user uses an argument scheduled to be dropped, they should get a warning:

``python
@deprecated_drop_argument("verbose", version="0.11.0")
def func(x, verbose=False):
    return x + 1

print(func(1, verbose=True))

example.py:8: FutureWarning: 'verbose' is deprecated and will be removed in 0.11.0.
          func(1, verbose=True)
```

Issue #1321, opened 2026-09-09
[FEAT] Add deprecation tool to rename a function argument


genedan:
When a user uses an argument scheduled to be renamed, it should issue a warning:

```python
@deprecated_rename_argument("old_arg", "new_arg", version="0.11.0")
def func(new_arg):
    return new_arg + 1

print(func(old_arg=1))

example.py:8: FutureWarning: 'old_arg' is deprecated and will be renamed to 'new_arg' in 0.11.0. Use 'new_arg' instead.
          func(old_arg=1)
```

Issue #1320, opened 2026-09-09
[FEAT] Add deprecation tool to rename a function


genedan:
When a user uses a function that is scheduled to be renamed, it should print a warning. This can be accomplished with a decorator:

```python
@deprecated_rename("new_func", version="0.11.0")
        def old_func(x):
            return x + 1

old_func(1)

example.py:8: FutureWarning: 'old_func' is deprecated and will be renamed to 'new_func' in 0.11.0. Update your code to use 'new_func' instead.
          old_func(1)
```

henrydingliu:
i think my comments on the PR lack context. 

i'm assuming that we want new_func to be available right away, i.e. user doesn't have to wait till the future release to change the function name. based on my understanding of the implementation in the PR, the decorator doesn't insert new_func into the codebase. if this understanding is off, i will need to withdraw all my comments. 

perhaps the more intuitive pattern to implement is this. calling old_func gives a warning but returns new_func. deprecating old_func down the line is accomplished by simply removing the decorator. 

```python
@deprecate_rename("old_func")
def new_func():
    return "something"
```

genedan:
I figured from your review comments that you envisioned the refactoring steps involved happening in a different order and/or which function serves as the wrapper or gets decorated. The steps in the PR would be:

- new_func gets defined as a wrapper to old_func
- old_func gets decorated with the deprecation warning - it, still works, but warns the user to use new_func instead
- once we're ready to drop old_func, we move old_func's code to new_func and then delete old_func along with its decorator

`old_func` is the one that needs to be decorated, because calling old_func triggers the warning.

henrydingliu:
i think i prefer the work pattern you set for the argument renaming. it front loads the work.

PR #1323, opened 2026-09-09
[FEAT] Add deprecation tools

genedan:
## Summary of Changes  

Needed to close #1316.

Adds three generalized deprecation tools:

- Rename a function
- Rename a function argument
- Drop an argument

This simplifies the deprecation process by allowing contributors to simply mark a feature with a decorator.

This enables a standardized way of marking features for deprecation, and enables us to run checks prior to releasing new versions that we actually deprecated the things we said we would.

## Related GitHub Issue(s)  

#1320 
#1321 
#1322 

## Checklist

---

> [!NOTE]
> **Low Risk**
> Adds internal deprecation helpers and tests only; runtime behavior of the library is unchanged until decorators are applied to specific APIs.
> 
> **Overview**
> Introduces **three decorator factories** in `chainladder/_config/deprecation.py` for a consistent deprecation workflow: `_deprecated_rename` (warn when an old function name is called), `_deprecated_rename_argument` (map old kwargs to new names with a warning, reject both), and `_deprecated_drop_argument` (warn when a kwarg slated for removal is used). Each supports optional target version and warning category, uses `functools.wraps`, and emits `FutureWarning` by default.
> 
> The helpers are **re-exported** from `chainladder._config` and the top-level `chainladder` package alongside existing deprecation utilities. **ARCHITECTURE.md** now documents `chainladder/_config/tests/`. A new **`test_deprecation.py`** suite covers warning text, categories, argument forwarding, metadata preservation, and repeat-call behavior.
> 
> No existing public APIs are wired to these decorators in this PR; this is foundational tooling for upcoming deprecations (e.g. issues #1320–#1322).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a4e1e3235324c825a7b488918f1c8c83cda306e6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/chainladder/_config/deprecation.py
+++ b/chainladder/_config/deprecation.py
@@ -154,7 +154,7 @@ def _resolve_pat(
 _F = TypeVar("_F", bound="Callable[..., object]")
 
 
-def deprecated_rename(
+def _deprecated_rename(
     new_name: str,
     *,
     version: str | None = None,
@@ -187,9 +187,9 @@ def deprecated_rename(
     .. testcode::
         :options: +SKIP
 
-        from chainladder._config.deprecation import deprecated_rename
+        from chainladder._config.deprecation import _deprecated_rename
 
-        @deprecated_rename("new_func", version="0.11.0")
+        @_deprecated_rename("new_func", version="0.11.0")
         def old_func(x):
             return x + 1
 
@@ -219,7 +219,7 @@ def deprecated_rename(
     return decorator
 
 
-def deprecated_rename_argument(
+def _deprecated_rename_argument(
     old_name: str,
     new_name: str,
     *,
@@ -261,9 +261,9 @@ def deprecated_rename_argument(
     .. testcode::
         :options: +SKIP
 
-        from chainladder._config.deprecation import deprecated_rename_argument
+        from chainladder._config.deprecation import _deprecated_rename_argument
 
-        @deprecated_rename_argument("old_arg", "new_arg", version="0.11.0")
+        @_deprecated_rename_argument("old_arg", "new_arg", version="0.11.0")
         def func(new_arg):
             return new_arg + 1
 
@@ -298,7 +298,7 @@ def deprecated_rename_argument(
     return decorator
 
 
-def deprecated_drop_argument(
+def _deprecated_drop_argument(
     name: str,
     *,
     version: str | None = None,
@@ -330,9 +330,9 @@ def deprecated_drop_argument(
     .. testcode::
         :options: +SKIP
 
-        from chainladder._config.deprecation import deprecated_drop_argument
+        from chainladder._config.deprecation import _deprecated_drop_argument
 
-        @deprecated_drop_argument("verbose", version="0.11.0")
+        @_deprecated_drop_argument("verbose", version="0.11.0")
         def func(x, verbose=False):
             return x + 1
 
--- a/chainladder/_config/deprecation.py
+++ b/chainladder/_config/deprecation.py
@@ -7,14 +7,15 @@ Utilities for deprecating chainladder features.
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 from __future__ import annotations
 
+import functools
 import inspect
 import warnings
 
-from typing import overload, TYPE_CHECKING
+from typing import overload, TYPE_CHECKING, TypeVar
 
 if TYPE_CHECKING:
     from types import FrameType
-    from typing import Literal
+    from typing import Callable, Literal
 del TYPE_CHECKING
 del annotations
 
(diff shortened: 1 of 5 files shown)


henrydingliu:
curious why you are skipping all the example testcode blocks?

genedan:
testcode/testoutput doesn't capture warnings - we'd have to test those using pytest

chainladder/_config/deprecation.py
```diff
@@ -147,3 +148,213 @@ def _resolve_pat(
...
+    --------
+
+    .. testcode::
+        :options: +SKIP
+
+        from chainladder._config.deprecation import _deprecated_rename
+
+        @_deprecated_rename("new_func", version="0.11.0")
+        def old_func(x):
+            return x + 1
```
henrydingliu, line 194: should this snippet be

```python 
@_deprecated_rename("new_func", version="0.11.0")
def old_func(x):
    return new_func(x)

print(old_func(1))
```

chainladder/_config/tests/test_deprecation.py
```diff
@@ -0,0 +1,350 @@
...
+    with warnings.catch_warnings(record=True) as caught:
+        warnings.simplefilter("always")
+        result = func(*args, **kwargs)
+    assert caught is not None
+    # Check that the warning was triggered.
+    assert len(caught) == 1
+    return result, caught[0]
+
+
+class TestDeprecatedRename:
```
henrydingliu, line 32: is this a new pattern? grouping similarly purposed tests under a class?

chainladder/_config/tests/test_deprecation.py
```diff
@@ -0,0 +1,350 @@
...
+        def old_func():
+            pass
+
+        with pytest.warns(DeprecationWarning):
+            old_func()
+
+    def test_forwards_args_and_kwargs(self) -> None:
+        """Check that positional and keyword arguments reach the wrapped function unchanged."""
+
+        @_deprecated_rename("new_func")
```
henrydingliu, line 85: similar to my other comment on the docstring, i think the more realistic use pattern is for old_func to simply return new_func. a stronger test here would be to test that positional and keyword arguments reach new_func via old_func.

chainladder/_config/tests/test_deprecation.py
```diff
@@ -0,0 +1,350 @@
...
+            return a, b, c
+
+        with warnings.catch_warnings():  # noqa
+            warnings.simplefilter("ignore")
+            assert old_func(1, 2, c=3) == (1, 2, 3)
+
+    def test_preserves_metadata(self) -> None:
+        """Check that functools.wraps preserves the function's name and docstring."""
+
+        @_deprecated_rename("new_func")
```
henrydingliu, line 96: this test got me thinking. is there a way to automatically populate docstring of old_func with either a copy of new_func docstring or a link to new func (for the API page)? basically, what we would have in the repo would be

```python
def new_func():
    """some docstring"""
    return "something"

@_deprecated_rename("new_func")
def old_func():
    return new_func()
```

but old_func shows up in the api page or at runtime with "some docstring" as the docstring

chainladder/_config/tests/test_deprecation.py
```diff
@@ -0,0 +1,350 @@
...
+            func(old_arg=1)  # pyright: ignore[reportCallIssue]
+
+    def test_positional_args_unaffected(self) -> None:
+        """Check that positional arguments pass through unaffected."""
+
+        @_deprecated_rename_argument("old_arg", "new_arg")
+        def func(a, b, new_arg=None):
+            return a, b, new_arg
+
+        with warnings.catch_warnings(record=True) as caught:
```
henrydingliu, line 217: why not call _warn_once?

chainladder/_config/tests/test_deprecation.py
```diff
@@ -0,0 +1,350 @@
...
+        assert warning.category is FutureWarning
+
+    def test_new_name_no_warning(self) -> None:
+        """Check that calling with the new argument name alone doesn't warn."""
+
+        @_deprecated_rename_argument("old_arg", "new_arg")
+        def func(new_arg):
+            return new_arg
+
+        with warnings.catch_warnings(record=True) as caught:
```
henrydingliu, line 139: i think this happens a couple of times to deserve a _warn_never?

chainladder/_config/tests/test_deprecation.py
genedan: It's definitely a thing, although rarely used in this library. I asked Claude to write me some tests and this is the initial organization. I had no objections, so I kept the structure. I do like how the classes clearly delineate the three deprecation features I added.

chainladder/_config/tests/test_deprecation.py
genedan: Pandas has something similar, they have a custom decorator that lets you share most of a docstring between objects:

https://pandas.pydata.org/docs/development/contributing_docstring.html#sharing-docstrings

Something like this could come in handy in a case like [cl.read_csv](https://github.com/casact/chainladder-python/blob/main/chainladder/utils/utility_functions.py#L250) which copies most of the Triangle docstring.

henrydingliu:
@genedan bump

chainladder/_config/deprecation.py
genedan: You got it. I updated the docstring to change the rest of the steps. It wound up working without having to change the code.

chainladder/_config/tests/test_deprecation.py
genedan: fixed

chainladder/_config/tests/test_deprecation.py
genedan: Yes, added the helper function.

chainladder/_config/tests/test_deprecation.py
genedan: Right, I updated the tests to have old_func be the wrapper. For this one test, I also added an arithmetic example to check the keyword and positional arguments.

genedan:
k, it's ready.

henrydingliu (approved):

henrydingliu:
I started using the argument rename. since the work pattern is to already rename the argument before adding the decorator, the warning message actually should be stronger. will raise a PR shortly

PR #1335, opened 2026-09-11
OLS Bug + misc

### LINKED ISSUE #1326: [BUG] Potential Bugs from #1325

### Describe the bug in words

cursor caught two medium severity bugs in #1325. will address shortly

N/A

N/A

- [x] Yes, absolutely!
- [ ] No.

### LINKED ISSUE #1343: [BUG] Missing sp.repeat

### Describe the bug in words

`xp.repeat` is called in half a dozen modules in the repo, which is a huge joke because so far only the np path is used (mostly thanks to `DevelopmentBase._set_fit_groups`). We know this because repeat was only added to [sparse in 0.18](https://github.com/pydata/sparse/releases#release-0.18.0); however our unittests are [currently running on sparse 0.17](https://github.com/casact/chainladder-python/actions/runs/34560149613/job/103141020602#step:4:124). 

N/A

N/A

- [ ] Yes, absolutely!
- [ ] No.

### LINKED ISSUE #1348: SPEC 0: drop support for python 3.10 by 2026-10-31
Python 3.10 is due to reach end-of-life on **2026-10-31** (within window). Update `pyproject.toml` and related workflow files to drop this version, and consider raising the maximum supported version if a newer Python release is available.

This issue was opened automatically by the nightly SPEC 0 check (`.github/workflows/spec0_check_nightly.yml`). If there's a specific reason to keep supporting this version, close this issue with an explanation. Otherwise, leave it open until the minimum version is bumped.

henrydingliu:
## Summary of Changes  
fixed a bug in OLS path of weighted regression where mismatching NaN in X and y result in incorrect parameters

deprecated python 3.10 in order to raise minimum sparse version to 0.18 for sp.repeat

added a test to show that current implementation of full triangle weight already works on irregular triangles

moved `TriangleWeight` and `WeightedRegressions` tests to their own files

## Related GitHub Issue(s)  
closes #1326 
closes #1343 
closes #1348

we start the simple case
x = [1, 2, 3, 4, 5]
y = [1, 2, 3, 4, 5]
regression will come back with a slope of 1. ezpz

if one of these pairs of observations are incomplete, such as
x = [1, 2, 3, NaN, 5]
y = [1, 2, 3, 4, 5]
convention dictates that the fourth pair of observations is not considered in the regression, i.e. the weight bacomes [1, 1, 1, 0, 1]. 

there is also the extreme case of having incomplete, mismatching observations in both datasets
x = [1, 2, 3, NaN, 5]
y = [1, 2, NaN, 4, 5]
we want weight to become [1, 1, NaN, NaN, 1] to ensure we get the correct regression coefficient. 

## Checklist
- [ ] I passed tests locally for both code (`uv run pytest`) and documentation changes (`uv run --directory docs jb build . --builder=custom --custom-builder=doctest`)

---

> [!NOTE]
> **Medium Risk**
> Changes OLS coefficient logic used by reserving workflows; mitigated by new numpy/sparse tests but still affects numerical results where NaNs were mishandled.
> 
> **Overview**
> Fixes **weighted OLS** in `WeightedRegression._fit_ols` so observations with **NaN in `X` or `y` are excluded** (weights zeroed) instead of producing wrong slopes when missingness does not line up between `X` and `y`. The **sparse** path applies the same rule by masking weights with valid `x`/`y` coordinates.
> 
> Adds **`test_wtd_reg.py`** covering dense NumPy and sparse COO cases, and moves **full-triangle `TriangleWeight`** coverage into **`test_tri_w.py`** (including an irregular-grain case).
> 
> **Support matrix:** minimum Python is now **3.11** (CI matrices and `requires-python`), **`sparse>=0.18`**, and Python 3.10 is dropped from classifiers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ebc24f3e934cd9f2fd870b8c37e24a2cfa5b60e0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


kennethshsu:
@genedan do you want to take this one? Regression is not my strong suit at all.

genedan:
I can try, I think one thing we need to add is review practices on PRs that have a higher level of mathematics than the reviewer does (for me, the bar is not high either).

One day a professor will show up and make a PR with crazy math. If a paper comes from a reputable journal, has a good amount of numerical examples, and the author puts them in the tests, I think we can trust the mathematics.

kennethshsu:
I agree. I think I mentioned something along the lines of us considering the submitter's actuarial credentials, it helps gauge their background so the reviewer can adjust the review depth accordingly. Maybe credentials aren't a good "bar", but considering their background will definitely help. For example, if a very reputable/credible actuary submits a PR because they implemented their own method, I would consider carefully with any pushbacks, unless it's very obvious.

In this case, I know @henrydingliu and his background, and I also know I am not the most knowledgable person especially when it comes to regressions, I feel I can look at the tests and say ya ok good enough lol

henrydingliu:
dudes, there's no math in this PR. the changes are entirely around handling nan. i will be more descriptive in the summary

chainladder/utils/tests/test_wtd_reg.py
```diff
@@ -0,0 +1,43 @@
...
+
+import numpy as np
+from chainladder.utils.sparse import sp
+from chainladder.utils.weighted_regression import WeightedRegression
+
+
+class TestOLS:
+    """Test the OLS calculations"""
+
+    def test_missing_data(self) -> None:
```
genedan, line 11: Noob math questions from me:

- Are these handpicked scenarios where we know the beta will come out to be 1, and we're just making sure it does when there's missing data?
- And, even when we supply those 5 weights of 1 as input internally they get set to zero when there is a missing data point for the pair?

henrydingliu: yes to both. though for second question, you can think of it as weight of 1 being applied to a NaN cell. 

you can try running the test on main. it will fail because the weights get applied incorrectly to one side of the observation pair that has a NaN.

genedan (approved):

Issue #1332, opened 2026-09-11
[ENH] Define cl.Styler class


genedan:
I believe we should start by adapting the existing `pd.io.formats.style.Styler` and then gradually augmenting it with support for actuarial elements (lower triangle, diagonals, etc.).

With #1331 being such an ambitious idea, this may be a case of perfection being the enemy of good. Maybe one day we'll have a 100% custom in-house Styler, but this a very complex class. Since Triangles are printed as `DataFrame`s anyway, I believe starting with the Pandas Styler is appropriate.

Issue #1337, opened 2026-09-11
[ENH] Create Triangle Visualization tutorial


genedan:
The features added in #1331 and its sub-issues should be documented with a Triangle Visualization tutorial, analogous to:

https://pandas.pydata.org/docs/user_guide/style.html

henrydingliu:
@genedan good to close?

Issue #1333, opened 2026-09-11
[ENH] Define Triangle.style() method.


genedan:
`Triangle.style()` should return a `cl.Styler` class, defined in #1332.

Issue #1331, opened 2026-09-10
[ENH] Define Triangle Styler


genedan:
Another multi-part issue...

Referencing #1324, it's time to define a Styler for the Triangle. A Styler is a class that enables aesthetic configuration of the Triangle for output to various displays, in HTML, LaTeX, etc. For the Pandas styler, refer to the following tutorial:

https://pandas.pydata.org/docs/user_guide/style.html

This can be accessed via `Triangle.to_frame().style.format(...)`, but challenges remain:

- The Pandas styler cannot access the actuarial elements of the Triangle, such as latest diagonal, lower triangle (see #1324), which puts a high burden on the user to define complex, custom functions to format these elements.
- No support for origin/development indexes

**Concatenation**

Another cool feature that we can adapt is the [concatenation](https://pandas.pydata.org/docs/user_guide/style.html#Concatenating-DataFrame-Outputs) of stylers. For example, applying this to a Triangle(s) with a compatible table of LDFs should give us the ability to build the exhibits in Friedland, such as:

<img width="1558" height="1258" alt="Image" src="https://github.com/user-attachments/assets/9e286489-3205-4703-ba2e-d10f667fbae9" />

**Builtin styles**

The Pandas styler has [builtin styles](https://pandas.pydata.org/docs/user_guide/style.html#Builtin-Styles). The `Triangle` styler can have its own builtins, where we can format things like latest diagonal, lower triangle, upper triangle, origin, development, etc. This can be done via:

```python
Triangle.style.highlight_lower_triangle(color="blue")
```

Which can be the result of solving #1324.

priyam0k:
Really excited to see this coming together. 
I'm working through the Friedland Chapter 12 recreation right now, and having lower triangle and diagonal styling will make those case outstanding exhibits look just like the text. 
I can jump on #1334 (highlight_diagonal) right after #1342 lands.

PR #1342, opened 2026-09-11
[ENH] Define Triangle Styler

genedan:
## Summary of Changes  

Adds functionality to style Triangles directly.

## Related GitHub Issue(s)  

Closes:

#1332 
#1333 
#1324 

Partially addresses:

#1337

The docs have a stub tutorial in `user_guide/style.ipynb`. This allows the reviewer to visually inspect what the new styling looks like. I'll fill it out with a proper tutorial towards the end of the parent issue #1331.

I recommend the reviewer to first take a look at the notebook to confirm the feature works. `pandas.py` is where `Triangle.style()` returns the styler, which is defined in `style.py` (where the bulk of the formatting logic takes place).

## Checklist

---

> [!NOTE]
> **Low Risk**
> Presentation-layer API addition with validation and broad tests; no changes to reserving math or data mutation paths.
> 
> **Overview**
> Adds a public **`Styler`** (pandas `Styler` subclass) and a **`Triangle.style`** property so single triangles can be formatted and rendered like DataFrames in notebooks and reports. Construction wraps the triangle’s frame via `_repr_format`, applies default numeric formatting, and rejects non-triangles, empty triangles, and multidimensional triangles.
> 
> The actuarial-specific **`highlight_lower_triangle`** highlights cells beyond a valuation cutoff (with optional `valuation_date`, custom CSS, and special handling for fully predicted triangles), and explicitly errors on valuation triangles. **`_copy`** is overridden so chained pandas styler methods keep returning chainladder’s `Styler`.
> 
> **`Styler`** is exported from the top-level API; docs gain a **Triangle Visualization** user guide, API autosummary entries, and architecture notes. Coverage includes parity with pandas HTML output, subclass preservation on chaining, and extensive tests for lower-triangle masking edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3c9f69433bf8ad5c7dfb6dacece26f6d9e25f651. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chainladder/core/typing.py
```diff
@@ -108,24 +115,46 @@ def drop(
...
     def __getitem__(self, key: pd.Series | np.ndarray | list[str]) -> Triangle: ...
     @overload
     def __getitem__(self, key: str | int) -> Triangle | Series: ...
-    def __setitem__(self, key: str | int, value: int | float | TriangleSlicer | Callable[[Triangle], TriangleSlicer]) -> None: ...
\ No newline at end of file
+    def __setitem__(
+        self,
+        key: str | int,
+        value: int | float | TriangleSlicer | Callable[[Triangle], TriangleSlicer],
+    ) -> None: ...
```
henrydingliu, line 160: can you confirm this is ruff formatting changes only?

chainladder/core/style.py
```diff
@@ -0,0 +1,163 @@
...
+        from chainladder.core.style import Styler
+
+        df = pd.DataFrame({"a": [1, 2], "b": [3, 4]})
+        Styler(df).format(precision=1)
+
+    """
+
+    def __init__(
+        self,
+        data: pd.DataFrame | pd.Series,
```
henrydingliu, line 65: is this needed?

chainladder/core/pandas.py
```diff
@@ -236,6 +237,25 @@ def hvplot(self, *args: Any, **kwargs: Any) -> Any:
...
+            A Styler wrapping this Triangle's ``to_frame()`` representation.
+        """
+        data = self.to_frame(origin_as_datetime=False)
+        styler = Styler(data, triangle=self)
+        if self._dimensionality == "single":
+            fmt_str = self._get_format_str(data=data)
+            styler = styler.format(  # pyright: ignore[reportReturnType]
+                fmt_str, na_rep=""
+            )
+        return styler
```
henrydingliu, line 257: is there any value in allowing styler on multi triangle but essentially blocking multi triangle from using any of the custom cl stylers? (i know we only have a single styler right now. but anything we might add, like highlight_diagonal, etc. would all only work on single triangles) like, what if we just let multi triangles error out right here

chainladder/core/style.py
```diff
@@ -0,0 +1,163 @@
...
+    Examples
+    --------
+
+    .. testcode::
+
+        import pandas as pd
+        from chainladder.core.style import Styler
+
+        df = pd.DataFrame({"a": [1, 2], "b": [3, 4]})
+        Styler(df).format(precision=1)
```
henrydingliu, line 59: can this mimic the [pandas doc](https://pandas.pydata.org/docs/reference/api/pandas.io.formats.style.Styler.format.html#pandas.io.formats.style.Styler.format) and actually show an output? something like precision = 1 on a raa.age-to-age

chainladder/core/style.py
```diff
@@ -0,0 +1,163 @@
...
+
+        .. testcode::
+            :options: +SKIP
+
+            raa = cl.load_sample("raa")
+            full = cl.Chainladder().fit(raa).full_triangle_
+            full.style.highlight_lower_triangle(
+                color="lightgray", valuation_date=raa.valuation_date
+            )
+
```
henrydingliu, line 135: can we follow [pandas](https://pandas.pydata.org/docs/reference/api/pandas.io.formats.style.Styler.highlight_max.html#pandas.io.formats.style.Styler.highlight_max) and link to the tutorial notebook?

chainladder/core/style.py
```diff
@@ -0,0 +1,163 @@
...
+            raise ValueError(
+                "highlight_lower_triangle only supports a single (2-D) Triangle."
+            )
+
+        if props is None:
+            props = f"background-color: {color};"
+            if text_color is not None:
+                props += f" color: {text_color};"
+
+        def f(_data: pd.DataFrame) -> np.ndarray:
```
henrydingliu, line 160: i feel like we are gatekeeping a lot of generalizable code inside this single-purpose styler. how about overloading or replacing `cl.Styler.apply` with `apply(indicator_triangle, true_prop, false_prop)`?  then `highlight_lower_triangle` can route through that `apply`.

chainladder/core/style.py
genedan: I think the `triangle` argument is the extraneous one. I can eliminate it, and restrict `data` so that it still matches the Pandas API, but only takes a `Triangle` since a regular DataFrame or Series doesn't have components like a lower diagonal.

chainladder/core/style.py
genedan: Yeah it can, I'll need to get rid of this example though. We really shouldn't be allowing the Styler to take a regular DataFrame or Series since it's designed to work on the Triangle-specific attributes. I'll replace this with a precision example (will add the feature if needed).

chainladder/core/style.py
genedan: Yes

chainladder/core/typing.py
genedan: Yeah, it was ruff.

chainladder/core/pandas.py
genedan: I've updated the code to throw an exception if the triangle is multi. Although I do need your thought on e putting "2-D" in the error message. It's technically 2-D (origin x development), but it's often called 1-D throughout the code. I think "single triangle" is a little less controversial.

henrydingliu: 'single'/'multi' is how we are calling it in the codebase. i'm good with single.

chainladder/core/style.py
genedan: We can get rid of this `f` function and create a generalized method to create an indicator triangle, then pass it to `apply`. Will let you know when it's done.

genedan:
@henrydingliu ready for review.

That last comment about `apply` was a bit tricky. Sticking the indicator triangle function (`_mask_style`) leads to an infinite recursion, so I had to wrap it in `partial`.

henrydingliu:
> @henrydingliu ready for review.
> 
> That last comment about `apply` was a bit tricky. Sticking the indicator triangle function (`_mask_style`) leads to an infinite recursion, so I had to wrap it in `partial`.

will review tonight. but really curious why it would lead to an infinite loop. wasn't the previous implementation already using a triangle-shaped np.array? with `_mask_style` being private, we aren't really opening up the API

chainladder/core/style.py
```diff
@@ -36,39 +40,154 @@ class Styler(_PandasStyler):
...
         self,
-        data: pd.DataFrame | pd.Series,
+        triangle: Triangle,
         *args: Any,
-        triangle: TriangleProtocol | None = None,
         **kwargs: Any,
     ) -> None:
+        # Imported here rather than at module scope: Triangle's own mixins import
+        # this module, so a top-level import would be circular.
+        from chainladder.core.triangle import Triangle
```
henrydingliu: is this the first time we've come across something like this in this package? [pandas implementation](https://github.com/pandas-dev/pandas/blob/2905718b126579d2cf0fe5bae05bd75facb67d77/pandas/io/formats/style.py#L29) doesn't seem to have this problem.

henrydingliu:
apologies for the delay. no other concerns on the code changes. i'm good with getting this in first and revisiting `apply(_mask_style)` later. 

will find time this afternoon to review doc changes.

chainladder/core/style.py
genedan: That's a whoopsie. I thought I'd move this back to the top but I guess it's still there. I'll move it back.

genedan:
Oof, moving that Triangle import caused a bunch of tests to fail. Gonna investigate.

genedan:
Fixed. I don't think you can avoid the mid-module import. Pandas has to do it too, but they import the styler module in the DataFrame class. I copied this pattern for chainladder by moving the styler method to `Triangle` and importing the styler there.

https://github.com/pandas-dev/pandas/blob/2905718b126579d2cf0fe5bae05bd75facb67d77/pandas/core/frame.py#L1432

https://github.com/pandas-dev/pandas/blob/2905718b126579d2cf0fe5bae05bd75facb67d77/pandas/core/frame.py#L12900

chainladder/core/style.py
```diff
@@ -0,0 +1,282 @@
...
+                "index",
+                "columns",
+                "_triangle",
+            ):
+                continue
+            setattr(
+                styler,
+                key,
+                copy_module.deepcopy(value) if deepcopy else value,
+            )
```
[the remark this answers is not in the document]
genedan: Won't change, this behavior is in line with pandas.

genedan:
Trying to set the decimals in the API ref looks a little weird:

image

The styler is meant for html output, so I've moved this example to the tutorial.

henrydingliu:
a few thoughts after reading through doc pages: 

- pandas puts the link to the guide under the class example, rather than a method example. as well as a "See Also" (for `Triangle.style`). I think both would be pretty neat.
image

- `text_color` in `highlight_lower_triangle` should be moved up and follow `color`. 
- Can `None` for valuation_date in `highlight_lower_triangle` have a separate branch for full triangles? i.e. `valuation[0, 0, -1, 0]`

genedan:
We're getting some funky behavior on dev_to_val triangles:

image

What should be highlighted? I was thinking:

- Unpredicted case: Nothing
- full triangle: The rectangle >= 1991

henrydingliu:
i would straight-up not support dev_to_val. if we move onto exam 6 and that year's NAIC returns have some fancy formatting that someone wants to replicate, they can raise an issue.

genedan:
k, made the changes.

There are still some edge cases where the highlighting won't be right (like if the last origin has observed values at 24 instead of 12 and the diagonal starts at 24 months). That's the warning from Bugbot, but rather than try to code up some super complicated inference, I figure the option to manually select the diagonal should cover that case.

A "normal" full triangle can fill without providing the valuation argument.

henrydingliu (approved):

henrydingliu:
@genedan thanks for the extremely meticulous and monumental work!

btw, I really like that our styler throws out errors. Pandas styler is so passive aggressive. when you do something it doesn't like, it just blanks out. this makes debugging more tedious than it has to be.

genedan:
45576d16-dbf8-4b86-a805-4b3a3c1dbd1b_text

PR #1351, opened 2026-09-13
Add an ultimate label alias, global and per-Triangle

### LINKED ISSUE #1345: [ENH]/[BRK] Add ultimate alias
Chainladder sets some far away date/development age to indicate ultimate values:

Image

While this allows the development dimension to contain values of the same data type, I don't think this will fly when it comes to executive presentations, IBNR meetings, etc. This presents a "last mile" type problem where the user is all like, "man, I want to relabel that value, I guess it's time to paste everything in Excel."

Secondly, the placeholder values are unlikely to be indexed directly as the user would have to look up the "2262-01-01" value as it's not easy to remember.

I propose that we allow the user to relabel the value with an alias of their choosing, such as "ULT", "Ultimate", "Ultimate Loss", "Ultimate Cheeseburgers", etc. This would allow them a little bit more flexibility and convenience and more of an incentive to stay within Chainladder. The numeric value (2262...or 9999) can be maintained internally if needed for calculations.

Abhayindia:
Adds `ULT_LABEL` as a global option plus a per-Triangle `ult_label` that overrides it, as discussed on the issue.

The alias is applied at display time only - `_repr_format`, which `__repr__`, `_repr_html_` and `to_frame` all go through - so `ddims` and `.development` keep the `ULT_VAL`/9999 sentinel and nothing downstream changes. That matters for things like `utility_functions.py`'s run-off filter, which string-compares development labels against a valuation date.

```python
>>> ult = cl.Chainladder().fit(cl.load_sample('raa')).ultimate_
>>> list(ult.to_frame(origin_as_datetime=False).columns)
['2261']
>>> cl.options.set_option('ULT_LABEL', 'Ultimate')
>>> list(ult.to_frame(origin_as_datetime=False).columns)
['Ultimate']
>>> override = ult.copy(); override.ult_label = 'ULT'
>>> list(override.to_frame(origin_as_datetime=False).columns)
['ULT']
>>> list(ult.to_frame(origin_as_datetime=False).columns)   # global untouched
['Ultimate']
>>> ult.development.tolist()                               # sentinel intact
['2261']
```

It covers the 9999 development-lag case too, e.g. `full_expectation_` ends `[..., 132, 'Ultimate']`.

I left the default as `None` so nothing changes unless you opt in - happy to default it to `"Ultimate"` if that's the `[BRK]` you had in mind, it's a one-line change.

`pytest chainladder` gives 1142 passed. The 2 remaining failures are `test_heatmap_render`, which fails the same way on a clean checkout here (no IPython in my venv).

Closes #1345.


Abhayindia:
saw @henrydingliu's note on #1345 - i built this to the spec genedan gave there, so no attachment to it either way. happy to close if the team lands on not wanting it.

PR #1356, opened 2026-09-14
Keep latest_diagonal a Triangle when there is one origin period

### LINKED ISSUE #1352: [BUG] latest_diagonal returns a float on a single-origin triangle, breaking Chainladder().fit()

### Describe the bug in words

`Chainladder().fit()` raises an `AttributeError` on a triangle that has a single origin period. The cause is `latest_diagonal`: on a one-origin triangle the diagonal is a single cell, and reducing it with `.sum(axis="development")` gives back a numpy float instead of a Triangle.

It isn't specific to `sum` or to the development axis - any reduction whose result would be shape `(1, 1, 1, 1)` comes back as a scalar. `sum`, `mean`, `min`, `max`, `std`, `median` and `var` all do it, on either axis. On anything larger the same calls keep returning a Triangle, so `latest_diagonal` quietly changes type at the smallest size.

Single-origin triangles come up when you slice one accident year out to look at it on its own, so it's easy to hit.

```python
import chainladder as cl

tri = cl.load_sample('raa')
one = tri[tri.origin == tri.origin[0]]

print(type(one.latest_diagonal))
# <class 'numpy.float64'>

cl.Chainladder().fit(cl.Development().fit_transform(one)).ultimate_
# AttributeError: 'numpy.float64' object has no attribute 'get_array_module'
```

```python
assert isinstance(one.latest_diagonal, cl.Triangle)
assert one.latest_diagonal.shape == (1, 1, 1, 1)
```

Abhayindia:
Closes #1352. `latest_diagonal` now asks for `keepdims=True`, so it stays a Triangle when the diagonal is a single cell.

The scalar return in the aggregation is deliberate - `tri.sum()` giving a number is useful - so this only changes `latest_diagonal`, which documents itself as returning a Triangle. `fit` and `predict` both work on a one-origin Triangle again, and they agree with the full triangle:

```python
>>> raa = cl.load_sample('raa')
>>> one = raa[raa.origin == raa.origin[0]]
>>> type(one.latest_diagonal)
<class 'chainladder.core.triangle.Triangle'>
>>> cl.Chainladder().fit(cl.Development().fit_transform(one)).ultimate_.values[0, 0, 0, 0]
18834.0
>>> cl.Chainladder().fit(raa).predict(one).ultimate_.values[0, 0, 0, 0]
18834.0
```

Two tests added, both fail on main.

One thing I left alone: the development label on the one-origin case comes out as the age (`120`) rather than the valuation date the multi-origin case gets (`1990`), because the relabelling in the aggregation is guarded on `len(obj.ddims) > 1`. Happy to fix that here too if you want it consistent, but it touches every `axis="development"` reduction rather than just this path, so I didn't fold it in.

`pytest chainladder` gives 1143 passed. The 2 failures are `test_heatmap_render`, which fails the same way on a clean checkout here (no IPython in my venv).


henrydingliu:
@Abhayindia thanks for the clean fix and the additional tests

>One thing I left alone: the development label on the one-origin case comes out as the age (120) rather than the valuation date the multi-origin case gets (1990), because the relabelling in the aggregation is guarded on len(obj.ddims) > 1. Happy to fix that here too if you want it consistent, but it touches every axis="development" reduction rather than just this path, so I didn't fold it in.

if it doesn't negatively affect your usage of the package, i'm fine with leaving it alone. would you mind raising a new issue for it?

Abhayindia:
raised it as #1358, thanks.

henrydingliu (approved):

Issue #1355, opened 2026-09-14
[TEST] Convert pytest fixtures to session scope


genedan:
### Description

On a whim, I ran the Pandas test suite on my computer. It has approximately 250,000 unit tests compared to our ~1,200.  Yet the whole thing ran in just a little over 16 minutes, whereas chainladder's runs in about 5 minutes.

Pandas: ~ 250 tests per second
Chainladder: ~ 4 tests per second (!)

This indicates to me that there may be improvements to the way we conduct tests.

### Is your feature request aligned with the scope of the package?

- [x] Yes, absolutely!
- [ ] No, but it's still worth discussing.
- [ ] N/A (this request is not a codebase enhancement).

### Describe the solution you'd like, or your current workaround.

The current scope of the fixtures is not set, meaning triangles like `raa` and `clrd` get reloaded every time a test uses them. I propose the scope to be changed to session-based so that they are not reloaded each time.

### Would you be willing to contribute this ticket?

- [x] Yes, absolutely!
- [ ] Yes, but I would like some help.
- [ ] No.

PR #1357, opened 2026-09-14
[TEST] Cache pytest fixtures across test session

genedan:
## Summary of Changes  

Small change that reduces the runtime of the test suite by ~40%. Runs on my computer in 3 minutes as opposed to 5 minutes previously.

Defines a new function called `_cached_load_sample`. This caches the triangles so that each test loads a cached copy rather than rebuilding the triangle each time.

## Related GitHub Issue(s)  

#1355 

With parallelization, the suite can be executed in less than a minute locally:

image

## Checklist

---

> [!NOTE]
> **Low Risk**
> Test-only fixture plumbing with copy-on-use; no production code or data-handling paths change.
> 
> **Overview**
> **Caches sample triangle loading for the pytest suite** so repeated fixture use reuses in-memory data instead of calling `cl.load_sample` every time, with an reported ~40% runtime reduction.
> 
> Adds `_cached_load_sample` (`functools.lru_cache`) and updates `_sample_fixture` to yield `_cached_load_sample(sample).copy()` before optional transforms and backend selection, so the shared cache stays immutable while each test still gets its own triangle.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b4bdf3b604d7ff3ead8ec686d68838a9e620ad4d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


henrydingliu (approved):

henrydingliu:
this is awesome! is there a way to apply it to doctest also?

genedan:
I think you kind of sort of can but the benefit is really low. `load_sample()` gets constantly called in the unit tests, but not so much in the notebooks.

henrydingliu:
> I think you kind of sort of can but the benefit is really low. `load_sample()` gets constantly called in the unit tests, but not so much in the notebooks.

i don't mean the notebooks. i mean [these](https://chainladder-python.readthedocs.io/stable/library/generated/chainladder.TailConstant.html#chainladder.TailConstant)

genedan:
Oh yeah, I think it's possible to shorten that by limiting the scope to just `library/` so that it doesn't execute all the other docs.

 I think we just need to make sure that the rest of the docs are checked somewhere else, which should be covered by `--nbmake` in the regular pytest run. But some Claude diagnostics are indicating that we might be actually skipping the notebooks, I'll need to check.

henrydingliu:
> Oh yeah, I think it's possible to shorten that by limiting the scope to just `library/` so that it doesn't execute all the other docs.
> 
> I think we just need to make sure that the rest of the docs are checked somewhere else, which should be covered by `--nbmake` in the regular pytest run. But some Claude diagnostics are indicating that we might be actually skipping the notebooks, I'll need to check.

wait what? you want to delete this [line](https://github.com/casact/chainladder-python/blob/8656c03303a9947ae624a4c38071989773659278/.github/workflows/pytest.yml#L83)?

genedan:
No, just change the `jb build .` part from the last line to `jb build docs/library`

henrydingliu:
also claude is hallucinating. see [this](https://github.com/casact/chainladder-python/actions/runs/34667030454/job/103480908280#step:6:259).

henrydingliu:
> No, just change the `jb build .` part from the last line to `jb build docs/library`

so you don't want the examples to be executed?

genedan:
I do, but that's the directory for the API reference examples? What I think can be skipped are things like the gallery or tutorials that don't execute the docstring examples.

henrydingliu:
i was just bringing up the possibility of using cached load_sample in the api reference examples. we can definitely also talk about skipping other portions

PR #1359, opened 2026-09-14
Enable ruff BLE001 and resolve blind-except violations

### LINKED ISSUE #1338: [Ruff] Add rule: bare-except (BLE001)
See: https://docs.astral.sh/ruff/rules/bare-except/

Abhayindia:
## Description

Enables ruff's `BLE001` (`blind-except`) and resolves the five violations it surfaces.

Per the discussion in #1338, the issue originally named `E722` (bare-except), but that rule is already active: `select` includes the `E7` prefix, which covers it, and the codebase has no bare `except:` clauses. `BLE001` is the rule that flags overly broad `except Exception` / `except BaseException` clauses, and `BLE` was not previously selected.

The five violations are four different situations, so they are not handled uniformly.

### `chainladder/core/dunders.py` - narrowed to `KeyError`

This is the only violation in library code and the only behavioural change:

```python
x = x.sort_index()
try:
    y = y.loc[x.index]
except Exception:          # -> except KeyError
    x = x.groupby(list(common))
    y = y.groupby(list(common))
```

The `except` is a fallback for "these two indexes cannot be aligned by label". `KeyError` is what is actually raised there, verified rather than assumed: I temporarily instrumented the clause to record the exception type, then drove the branch through the `+` operator with two triangles that have the same `key_labels` and the same number of `kdims` but different keys. The only exception recorded was `KeyError`, and the groupby fallback produced the expected result.

Catching `Exception` here also swallowed genuine errors from `.loc` and silently substituted a different alignment strategy, which is the failure mode `BLE001` exists to prevent.

### `chainladder/development/tests/test_development.py` - try/except removed

```python
try:
    cl.Development(drop_high=False).fit_transform(tri)
except Exception:
    assert False
```

The intent is "this should not raise", but the `except` converts any specific failure into a bare `assert False` with no traceback. Calling it directly expresses the same intent and lets pytest report the real error, so the try/except is removed rather than narrowed.

### `chainladder/utils/tests/test_utilities.py` - narrowed to `AttributeError`

The existing comment already documents the intent: the conversion can fail when the optional `dask` dependency is absent, and the test only cares that the deprecation warning fired. With `dask` missing, `chainladder/utils/dask.py` falls back to `import numpy as dp`, so `dp.from_array(...)` raises `AttributeError: module 'numpy' has no attribute 'from_array'`. Confirmed by simulating dask's absence; the comment has been updated to say which error and why.

### `docs/getting_started/tutorials/tail-tutorial.ipynb` - grandfathered

Both violations are in the cell that teaches what a `TailCurve` failure looks like:

```python
try:
    cl.TailCurve().fit(cl.Development().fit(quarterly))
    print("This passes.")
except Exception:
    print("This fails because we did not transform the triangle")
```

The broad catch is the point of the lesson, so this is added to `per-file-ignores` alongside the existing `chapter_10.ipynb` and `chapter_7_part_2.ipynb` entries rather than changed.

### Note on the second commit

`chainladder/core/dunders.py` carried pre-existing `ruff format` drift, so touching it made the format job fail on lines I did not write. Following the pattern used elsewhere in this repo, the reformat is isolated in its own commit (`Apply ruff format to chainladder/core/dunders.py`) so the functional change stays reviewable on its own. The first commit is 14 insertions / 10 deletions across four files; the second is formatting only.

## Verification

- `ruff check` with `lint.per-file-ignores={}` passes on every changed file.
- Under the repo's real config, `ruff check .` passes repo-wide with `BLE001` enabled.
- With per-file-ignores wiped repo-wide, `main` reports 5 violations and this branch reports 7: the same 5, plus the 2 `tail-tutorial.ipynb` blind-excepts that this PR grandfathers deliberately.
- `ruff format --check` clean on all changed files.

## Checklist

- [x] Code passes `ruff check` with `lint.per-file-ignores={}` on the changed files
- [x] `ruff format` applied to changed files
- [x] Tests pass locally
- [x] No new dependencies

Closes #1338


pyproject.toml
```diff
@@ -105,7 +105,7 @@ preview = true
 extend-exclude = ["docs/getting_started/online_sandbox/sandbox_workbook_blank.ipynb"]
 
 [tool.ruff.lint]
-select = ["E2", "E4", "E7", "E9", "F", "B018", "UP034", "N802"]
+select = ["E2", "E4", "E7", "E9", "F", "B018", "BLE001", "UP034", "N802"]
```
genedan: Could you please:

- Put these rules in alphabetical order
- Have one rule per line

pyproject.toml
```diff
@@ -115,6 +115,9 @@ select = ["E2", "E4", "E7", "E9", "F", "B018", "UP034", "N802"]
 [tool.ruff.lint.per-file-ignores]
 "docs/friedland/chapter_10.ipynb" = ["E731", "F841"]
 "docs/friedland/chapter_7_part_2.ipynb" = ["N802"]
+# The broad excepts here are the lesson: the cell demonstrates what a
```
genedan: Let's tackle this notebook so that we can remove this entry in `[tool.ruff.lint.per-file-ignores]`, ideally we'd to phase this block out. In the notebook you'll find this block:

```python
print("First attempt:")
try:
    cl.TailCurve().fit(cl.Development().fit(quarterly))
    print("This passes.")
except:
    print("This fails because we did not transform the triangle")

print("\nSecond attempt:")
try:
    cl.TailCurve().fit(cl.Development().fit_transform(quarterly))
    print("This passes because we transformed the triangle")
except:
    print("This fails.")
```

You should change the exception clauses to catch `AttributeError`.

genedan (changes requested):
Thanks for the PR, just some minor change requests from me.

Abhayindia:
both done in 7a69b125 - select list is alphabetical one-per-line, and the tutorial now catches `AttributeError` with the `per-file-ignores` entry removed, so that block is back to just the two friedland notebooks.
checked it rather than assuming: the untransformed call raises `AttributeError: 'Development' object has no attribute 'copy'`, and `pytest --nbmake` on tail-tutorial.ipynb passes, so the cell still prints its message instead of raising. agreed the narrower catch is better for the lesson too.

genedan (approved):

PR #1362, opened 2026-09-15
Label the latest diagonal by valuation date with a single origin period

### LINKED ISSUE #1358: [BUG] latest_diagonal keeps the development age instead of the valuation date on a single-origin triangle

### Describe the bug in words

`latest_diagonal` labels its single development column with the latest valuation date, except when the Triangle has one origin period, where it keeps the development age instead. So the column comes back as `1990` normally and as `120` in the one-origin case - a string vs an int, for the same property.

It comes from the relabelling in the aggregation being guarded on `len(obj.ddims) > 1`: with one origin the diagonal already has a single development column, so the guard skips and the age survives.

Split out of #1356 at @henrydingliu's request - the guard is shared by every `axis="development"` reduction, so changing it is wider than that fix was.

```python
import chainladder as cl

raa = cl.load_sample('raa')
one = raa[raa.origin == raa.origin[0]]

print(list(raa.latest_diagonal.development))
# ['1990']

print(list(one.latest_diagonal.development))
# [120]
```

```python
assert list(one.latest_diagonal.development) == ['1990']
```

Abhayindia:
## Description

Fixes #1358. `latest_diagonal` labels its development column with the latest valuation date, except when the Triangle has one origin period, where the development age survives instead: `['1990']` normally, `[120]` with one origin.

### Cause

`chainladder/core/pandas.py` relabels the development axis after an aggregation only when the pre-aggregation `ddims` had more than one entry:

```python
if axis == 3 and obj.values.shape[axis] == 1 and len(obj.ddims) > 1:
```

With one origin, `self[self.valuation == self.valuation_date]` already comes back with a single development column, so nothing collapses, the guard is false, and the age is left in place. The condition is really a proxy for "did we collapse anything", but the question that matters is whether the surviving label still means the right thing: after a development aggregation on a development-based Triangle, a single column represents a valuation, not an age.

The condition is now widened, never narrowed, so no existing path changes behaviour:

```python
if (
    axis == 3
    and obj.values.shape[axis] == 1
    and (len(obj.ddims) > 1 or not self.is_val_tri)
):
```

### On the `dev_to_val` approach

@henrydingliu suggested reimplementing `latest_diagonal` on `dev_to_val`, which is the more natural operation and does fix the label. I built it first and it is not viable: it fails 4 tests that pass on `main`.

- `test_latest_diagonal_vs_full_tri_raa[sparse_only_run]`
- `test_latest_diagonal_vs_full_tri_clrd[sparse_only_run]`
- `test_misaligned_index[sparse_only_run]`
- `test_single_entry` (`IndexError` on an all-NaN Triangle)

Reduced: with a sparse-backed `raa`, `ultimate_.latest_diagonal` returns values while `full_triangle_.latest_diagonal` returns all NaN. On the numpy backend the two routes are identical, so this is specific to `dev_to_val` over an extended triangle on the sparse backend. Happy to open that as its own issue if it looks like a real bug rather than an unsupported combination.

I also tried keying the guard purely on `is_val_tri`. That passes the whole suite but is wrong: on a valuation-based Triangle summed over development it leaves `len(ddims) == 10` against `shape[3] == 1`. No test covers it, which is why the second test below exists.

## Tests

- `test_latest_diagonal_single_origin_ddim_label` covers the reported bug. Verified to **fail on `main`** in both the `normal_run` and `sparse_only_run` parametrizations.
- `test_sum_development_keeps_val_tri_ddims_consistent` pins the valuation-based case above. It passes on `main` already; it is there to stop a future change to this guard reintroducing the mismatch.

Full suite: **1150 passed, 7 skipped**, same as `main`. `ruff check` with `lint.per-file-ignores={}` and `ruff format --check` both clean.

## Checklist

- [x] Tests added
- [x] Full test suite passes locally
- [x] `ruff check` and `ruff format` clean
- [x] No new dependencies

Closes #1358


henrydingliu:
my main gripe with this implementation is that the condition for relabeling is only applicable when called by latest_diagonal and not actually unconditionally correct. meaning, i would classify the current implementation as a bug. 

ordinarily, because we mainly work with cumulative triangles, `sum(axis=development)` is never called. the one place where we sum incremental triangle into cumulative uses `cumsum`. try `raa.cum_to_incr().iloc[..., 0:7, 2:4].sum(axis='development')`, which is a realistic operation. the resulting label makes no sense.

i can look into the sparse failed tests this evening.

Abhayindia:
you're right, and i've reworked it - `core/pandas.py` is now byte-identical to main and the relabel happens inside `latest_diagonal` instead, where the column genuinely is the latest valuation. your example confirms the wider problem is real: `raa.cum_to_incr().iloc[..., 0:7, 2:4].sum(axis='development')` sums ages 36 and 48 and labels them '1990' - but it does that on main too, so it's pre-existing rather than something this PR introduces, and it's untouched now.
full suite 1154 passed, and the regression test is verified to fail on main in both parametrizations. happy to file the shared-guard mislabelling separately if you want it tracked - it needs the relabel removed from the aggregation path entirely, which affects every axis='development' caller, so that felt like your call rather than mine to bundle in here.

henrydingliu (approved):

PR #1365, opened 2026-09-15
Render reStructuredText roles in the docs as MyST roles

### LINKED ISSUE #1312: [DOC] Dead in-doc links

### Describe the bug in words

dead links

Image

https://chainladder-python.readthedocs.io/main/user_guide/adjustments.html#deterministic-methods

https://chainladder-python.readthedocs.io/main/user_guide/methods.html#bornhuetterferguson

valid link to the intended section in the doc

- [ ] Yes, absolutely!
- [ ] No.

Abhayindia:
## Description

Fixes #1312.

The dead links are reStructuredText roles written inside MyST markdown cells. MyST does not recognise `:class:` or `:ref:`, so they are emitted as literal text rather than links. The screenshot on the issue shows one of them:

> ...with a suite of "dropping" parameters. See `:ref:` `Omitting Link Ratios<dropping>`.

Nothing warns about this, because MyST treats an unknown role as plain text, so they accumulated quietly. The two pages named in the issue are the worst affected, but the same fault is on seven notebooks:

| notebook | roles |
| --- | --- |
| `user_guide/methods.ipynb` | 17 |
| `user_guide/workflow.ipynb` | 6 |
| `user_guide/adjustments.ipynb` | 5 + the `:ref:` |
| `user_guide/triangle.ipynb` | 5 + 1 `:func:` |
| `gallery/plot_clarkldf.ipynb` | 1 |
| `gallery/plot_stochastic_bornferg.ipynb` | 1 |
| `gallery/plot_triangle_slicing.ipynb` | 1 |

All 39 are converted to MyST role syntax, matching the `{cite}` roles already used in `development.ipynb`:

```
:class:`BornhuetterFerguson`   ->   {class}`~chainladder.BornhuetterFerguson`
```

The `~` keeps the rendered text as the bare class name while the full dotted path resolves from any page. Every referenced class has a generated API page, so these resolve to the existing reference docs.

### The `:ref:` needed a second fix

`:ref:`Omitting Link Ratios<dropping>`` also pointed at a label `dropping` that is not defined anywhere in the repo, so correcting only the syntax would have produced a link that renders properly and still goes nowhere. The section it means is `### Omitting link ratios` in `development.ipynb`, so that heading gets an explicit target following the convention already used there (`(development:clarkldf)=` and friends), and the reference now points at it.

## Verification

Built the HTML before and after with `jb build docs --builder html`:

| | `main` | this branch |
| --- | --- | --- |
| literal `:class:`/`:ref:`/`:func:` in built HTML | 36 | 0 |
| build warnings | 4 | 4 |

The warning count is the important half: it confirms none of the replacements introduced an unresolved reference, which would have swapped a visibly broken role for an invisibly broken one. The 4 warnings are identical on both sides and come from `library/governance.md` and `user_guide/index.md`, neither of which this PR touches.

`ruff format --check` is clean on the changed notebooks. `ruff check` with `lint.per-file-ignores={}` reports only the 5 pre-existing `docs/friedland/` violations, which are unrelated and already grandfathered.

## Checklist

- [x] Verified the fix by building the docs
- [x] No new dependencies
- [x] `ruff check` / `ruff format` clean on changed files

Closes #1312


henrydingliu (approved):
inspected class, func, and ref links in per-PR RTD version. All correct

PR #1383, opened 2026-09-17
[ENH] warn on drop exclusions only when the triangle has holes

### LINKED ISSUE #1307: [ENH] Revisiting drop warnings
revisiting #765 

inspired by some recent comments by @ppcvote

we've come a long way in the past few months, particularly around documentation. i think we can safely reduce the amount of warnings fired by the various drop functions in `DevelopmentBase`. Specifically, ignoring drops around the later development periods that should be expected by the user. 

for example, if we have a 10x10 triangle, that means we have a 9x9 age-to-age triangle. putting drop_high=True and a drop_low=True on this triangle should only kick out a warning for the first 7 development periods, i.e. when there's more than 2 potentially available origins. 

we can also route the warning through a dedicated function, in order to avoid duplicate code.

ppcvote:
Closes #1307.

**The gate is the shape of the triangle, not how many factors are populated.** A development period that is short because the triangle ends there cannot satisfy `preserve` whatever the caller does, so its exclusion being ignored is expected. A period that is short because values are missing is the case the warning exists for. The count a period could hold now comes from `nan_triangle`, narrowed by `n_periods`, and the warning fires only when that count could have satisfied `preserve`.

On `raa`, `drop_high=True, drop_low=True` is silent at `preserve` 1, 2 and 3. Blank the first development period for seven origins, so the 12-24 period holds two factors where the shape allows nine, and it warns at `preserve=1`, which is the umbrella case you described.

**The same gate is applied on both paths.** `TriangleWeight` is constructed directly by `Disposal`, `IncrementalAdditive` and `Learning`, none of which go through `Development`, so gating only `DevelopmentBase` would have left `TriangleWeight(n_periods=2, drop_high=1, drop_low=1, preserve=1).fit(raa.age_to_age)` warning on a triangle with no holes. There is a test on that path now; nothing in the suite covered it before.

**`drop_above`/`drop_below` warn only when a threshold actually removed a factor.** A period that already held fewer than `preserve` is not having an exclusion ignored.

**One `fit()` now warns once.** `Development.fit` computes the same drops twice, for `w_` and for `w_v2_`, and both warned. The `w_v2_` call is wrapped in a filter for this message only, as you suggested. Worth knowing: that wrapper also hides any future disagreement between the two paths. The `TriangleWeight` gap above was invisible through `Development` for exactly that reason, so once `w_v2_` lands it may be worth asserting the two agree rather than silencing one.

The four copies of the message are now one helper in `utils/utility_functions.py`, with the text unchanged.

**Tests.** Eight existing `pytest.warns` assertions covered the tail-period warning this issue removes, in `test_drophighlow`, `test_new_drop_5` and `test_new_drop_5a`; those wrappers are gone and the assertions they guarded are kept. The three in `test_new_drop_7` and `test_pipeline` still fire and are untouched. Five tests are added, ten cases across both backends. `chainladder/development`, `utils`, `methods` and `adjustments` are 307 passed.

Reverting all four source files to `main` fails 8 of the 10: 4 come from `base.py`, 2 from the `w_v2_` wrapper, 2 from the `TriangleWeight` gate. The other 2 are `test_drop_warning_fires_when_the_triangle_has_holes`, which passes either way, since the old rule warned there too; it documents the intent rather than guarding it.

`ruff check` runs on the changed files, so the 13 `multi-line-summary-second-line` findings that already sat in `triangle_weight.py` and `utility_functions.py` failed the job as soon as this PR touched them. They are the autofix and docstring text only, in a separate commit so the logic diff stays readable. Happy to drop that commit if you would rather clear them repo-wide on their own.


henrydingliu (approved):
unittest warnings went down by about 40

the much more impactful benefit was on the docsite itself. 
Image

henrydingliu:
@ppcvote much thanks for this!

henrydingliu:
@ppcvote did we cook too hard? the following is emitting a warning, and i don't think we meant it to. 

```python
cl.Development(drop_above=1.2).fit(raa)
```

ppcvote:
Two of raa's nine link ratio columns are entirely above 1.2, and I read that warning as doing its job rather than misfiring: 9 of 9 at age 12, 8 of 8 at 24. The threshold empties both, `preserve` holds them back, and the fitted pattern at those ages comes back 2.99936 and 1.62352, identical to `cl.Development().fit(raa)` with nothing dropped at all. Ages 36 and 48 do move, 1.27089 to 1.12159 and 1.17167 to 1.14392. So two of the nine columns ignored the threshold completely, and that warning is the only thing saying so.

Worth being clear about how much of that was actually agreed, though. #1307 defines one rule, about shape: a 10x10 triangle gives a 9x9 age to age, so the later periods cannot hold enough origins and should not warn. That rule is about what the triangle can supply, and it is the one `drop_high` and `drop_low` are measured against. Thresholds needed a different rule, because how many ratios `drop_above` removes is a property of the data rather than a count you asked for. The hole rule only ever reads which cells the triangle's shape provides, never what is inside them, so `drop_high=2` either has two ranks to take or the shape says it does not, and that is settled before a single ratio is read. `drop_above` cannot be settled that way: whether it removes anything is knowable only after reading the values. The rule I wrote for it is "warn only when the threshold actually removed something". That was my choice in the PR, not something #1307 settled, and `test_drop_x_warning_needs_something_dropped` only pins the silent direction. The case you just hit, where a threshold empties a whole column on a triangle with no holes at all, has never been pinned either way.

On provenance, so it is not a variable here: on 98f822a, the commit before #1383, that same sentence fires twice for this call, once from `triangle_weight.py:510` and once from `development/base.py:358`. On `main` it fires once. Both paths still call `warn_exclusions_ignored`; the `w_v2_` one sits inside the `catch_warnings` block in `Development.fit`, so only the `base.py` one surfaces.

What I think is genuinely wrong with the warning is that it names no age, so on a triangle with no holes it reads as noise, which is how it read to you. I would rather it said the exclusion was ignored at 12 and 24. That is contained but not a one liner: all four call sites currently collapse the flag with `np.any` over everything, so they would need to keep the development axis and reduce only over index and column, and the wording has to survive different index and column combinations hitting different ages, which raa does not exercise.

My own read is to keep it and name the ages, since raa shows it catches something real. If you would rather a threshold that empties a whole column went quiet, the way the hole rule lets shape off, I will scope it that way instead. Either way I will add the test that pins this case so it stops being undefined.

henrydingliu:
you are right. i got spooked by the warning message and forgot that it's actually the drop parameter itself being set way too low. i can find my way out of this hole. would really appreciate if you could help me take a look at #1411

Issue #1193, opened 2026-08-04
[REFACTOR] Deprecate `development` in Triangle constructor


henrydingliu:
### Description

The `development` argument in the `Triangle` constructor is a misnomer. We are currently feeding in valuation

```python
raa_df = pd.read_csv(
    "https://raw.githubusercontent.com/casact/chainladder-python/master/chainladder/utils/data/raa.csv"
)
raa = cl.Triangle(
    raa_df,
    origin="origin",
    development="development",
    columns="values",
    cumulative=True,
)
print(raa_df.head(10))
```

### Is your feature request aligned with the scope of the package?

- [x] Yes, absolutely!
- [ ] No, but it's still worth discussing.
- [ ] N/A (this request is not a codebase enhancement).

### Describe the solution you'd like, or your current workaround.

with #1229, `development` can already take age. the updated proposal is as follows:

1. at some point around 1.0
    a. introduce `valuation` as an alias of `development`, but only takes timestamp
    b. introduce `age` as another alias of `development`, but only takes age in months
    c. add deprecation notice for `development`
2. by 2.0
    a. drop `development`

### Would you be willing to contribute this ticket?

- [ ] Yes, absolutely!
- [ ] Yes, but I would like some help.
- [ ] No.

Issue #1400, opened 2026-09-18
[DOC] Example for using list of columns for valuation when declaring triangle.


henrydingliu:
### Description

The Triangle Constructor can take a list of columns, such as year, month. This is used [here](https://github.com/casact/chainladder-python/blob/05a9825fe9f69623914b9e764b2d84ad31bfdc7f/chainladder/methods/tests/test_predict.py#L216). but no public facing example, either in the docstring or in the user guide exists

### Is your feature request aligned with the scope of the package?

- [x] Yes, absolutely!
- [ ] No, but it's still worth discussing.
- [ ] N/A (this request is not a codebase enhancement).

### Would you be willing to contribute this ticket?

- [ ] Yes, absolutely!
- [ ] Yes, but I would like some help.
- [ ] No.

henrydingliu:
assigning to @kennethshsu since he was also surprised by this

henrydingliu:
Here is a screenshot of the current Triangle user guide, which very briefly describes how to do this, without a concrete example. 

<img width="774" height="420" alt="Image" src="https://github.com/user-attachments/assets/d54cddf1-3468-4fae-8bc1-fa2f7dc6d712" />

PR #1401, opened 2026-09-18
[FEAT] Deprecate `development` in `Triangle` constructor

henrydingliu:
## Summary of Changes
Uses the newly minted deprecation decorator to deprecate the `development` argument to `valuation` and `age`. 

wording changes to the decorator warning

replace all instances of development in triangle declaration in the codebase. will replace all instances in docs in a later PR

changed the stacklevel for dask deprecation warning in order to pass test

## Related GitHub Issue(s)

Refers #1193

## AI/LLM Usage

A lot of small changes scatters over the repo. Only used AI to triage the previously failing dask test

this PR exclusively changes a triangle declaration like this

```python
    df = pd.DataFrame(
        data={
            "origin": [2022, 2022, 2023],
            "development": [2022, 2023, 2023],
            "reported": [78000, 222000, 78000],
        }
    )
    tri_from_df = cl.Triangle(
        data=df,
        origin="origin",
        development="development",
        columns=["reported"],
        cumulative=True,
    )
```

to this

```python
    df = pd.DataFrame(
        data={
            "origin": [2022, 2022, 2023],
            "development": [2022, 2023, 2023],
            "reported": [78000, 222000, 78000],
        }
    )
    tri_from_df = cl.Triangle(
        data=df,
        origin="origin",
        valuation="development",
        columns=["reported"],
        cumulative=True,
    )
```

the `Triangle.development` and `Triangle.valuation` properties are unaffected. 

bifurcating valuation and age is technically an API change. however, age support for `Development` was only added very recently in #1229.

## Submitter's Checklist
- [X] I have reviewed and am adhering to the standards outlined in the project [Governing Doc](https://github.com/casact/chainladder-python/blob/main/docs/library/governance.md).
- [X] The PR subject title summarizes the changes, with one proper prefix (`[FIX]`, `[FEAT]`, `[DOCS]`, `[TST]`, `[CHORE]`, or `[BRK]`).
- [X] I am a human (not a bot), and this PR form is written by a human.

## Reviewer's Checklist
- [ ] The implementation addresses the associated issue(s).
- [ ] The implementation is appropriate, maintainable, and follows ARCHITECTURE.md.
- [ ] PR subject title has the proper prefix and the subject is appropriate.
- [ ] Relevant issue(s) are linked.
- [ ] AI/LLM usage is disclosed and appropriate.
- [ ] Documentation and tests are appropriate.
- [ ] CI tests passed, or any failures are acceptable.
- [ ] Leave a comment with the final recommendation (e.g. approve as is, request a secondary review, or flag an area for more review).

---

> [!NOTE]
> **Medium Risk**
> Changes the public `Triangle`/`read_csv` keyword API and splits age vs valuation date handling, though old kwargs still work with warnings until v1.5.
> 
> **Overview**
> Renames the primary **`Triangle`** and **`read_csv`** constructor API from **`development`** / **`development_format`** to **`valuation`** / **`valuation_format`**, with **`_deprecated_rename_argument`** emitting **FutureWarning**s (removal in **v1.5**) when callers still pass the old names. **`development_format`** is deprecated the same way.
> 
> **Development age** is no longer inferred from numeric **`valuation`** columns via **`allow_age`** in **`_to_datetime`**. Callers must pass a dedicated **`age`** column (mutually exclusive with **`valuation`**); **`_set_development`** builds valuation dates from age when **`age`** is set. Numeric columns passed as **`valuation`** now raise a clearer error instead of silently treating them as age.
> 
> Internal tests, docs examples, **`load_sample`** / **`SAMPLES`**, and other in-repo **`Triangle(...)`** calls are updated to **`valuation=`** (often mapping existing **`development`** column names). **`_deprecated_rename_argument`** warning copy now says the old name **will be removed** (including **"soon"** when no version). Dask backend deprecation **`stacklevel`** is raised so warnings still point at user code through the new decorator wrappers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d6a15aa2107b4440b073aea5921f170e7669df6c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


henrydingliu:
Major API update. Will need review from two codeowners.

genedan:
I have some clarifying questions (since actuarial lingo varies a lot):

- Why is the `development` parameter a misnomer?
- Will the development axis still refer to the 12, 24, 36, etc. that we see at the top of the triangle?
- Will the valuation axis (used in trend) still refer to the diagonals?

henrydingliu:
> Why is the development parameter a misnomer?

after constructing the triangle, the triangle.development property refers to the development age. the triangle.valuation virtual column is the "as of" date. this is inconsistent with the `development` argument, which could only take "as of" dates till very recently. so it's less of actuarial dialect between different shops, more of an internal consistency within the package

will add some notes to the PR summary

> Will the development axis still refer to the 12, 24, 36, etc. that we see at the top of the triangle?

this will not change. only the declaration is change. 

> Will the valuation axis (used in trend) still refer to the diagonals?

yes

chainladder/core/base.py
```diff
@@ -140,8 +140,9 @@ def _input_validation(
         index: str | list | None,
         columns: str | list,
         origin: str | list,
-        development: str | list,
-    ) -> tuple[None | list, None | list, None | list, None | list]:
+        valuation: str | list,
+        age: str,
```
kennethshsu, line 144: Is there a way to say only one of the two are required? If fact, what if the user supplies both (and they are not consistent), maybe tomorrow's problem.

chainladder/utils/data/_manifest.py
```diff
@@ -19,50 +20,50 @@
 
 ``origin``
     Column name(s) for the origin period.
-``development``
+``valuation``
     Column name(s) for the development period.
```
kennethshsu, line 24: We should update this. Column name (can it be plural?) for the valuation period.

chainladder/utils/data/_manifest.py
```diff
@@ -19,50 +20,50 @@
 
 ``origin``
     Column name(s) for the origin period.
```
kennethshsu, line 22: Same here? Plural on column?

chainladder/utils/utility_functions.py
```diff
@@ -282,8 +287,11 @@ def read_csv(
...
          A representation of the accident, reporting or more generally the
          origin period of the triangle that will map to the Origin dimension
-    development: str or list
-        A representation of the development/valuation periods of the triangle
+    valuation: str or list
+        A representation of the valuation dates of the triangle
+        that will map to the Development dimension
+    age: str
+        A representation of the development ages of the triangle
         that will map to the Development dimension
```
kennethshsu, line 295: Can we add "in month" somewhere?

chainladder/_config/deprecation.py
```diff
@@ -256,7 +256,7 @@ def _deprecated_rename_argument(
     new_name: str
         The keyword argument name it will be renamed to.
     version: str | None
-        The release the rename is expected to land in, e.g. "0.11.0".
+        The release the deprecation is expected to land in, e.g. "0.11.0".
```
kennethshsu: The release "and" the deprecation "that" is expected to land in (or on?).

chainladder/_config/deprecation.py
```diff
@@ -283,15 +283,17 @@ def func(new_arg):
...
 
     """
 
     def decorator(func: _F) -> _F:
-        message = f"'{old_name}' is deprecated and will be renamed to '{new_name}'"
+        message = f"'{old_name}' has been deprecated in favor of '{new_name}' and will be removed"
         if version:
             message += f" in {version}"
+        else:
+            message += " soon"
```
kennethshsu, line 296: I think the message can be improved, maybe:

"The function has been deprecated since 0.11.0, and will be removed in a future version" or something like that. Do we want to commit to something will be deprecated ON a specific version? The timing could be hard to manage.

kennethshsu: And I think we said we want the deprecation cycle to be at least 3 months long

chainladder/core/triangle.py
```diff
@@ -426,19 +432,24 @@ class Triangle(TriangleBase):
...
         columns: Optional[str | list] = None,
         index: Optional[str | list] = None,
         origin_format: Optional[str] = None,
-        development_format: Optional[str] = None,
+        valuation_format: Optional[str] = None,
         cumulative: Optional[bool] = None,
         array_backend: str = None,
         pattern=False,
         trailing: bool = True,
+        age: Optional[str | list] = None,
```
kennethshsu, line 452: I don't like the order of these parameters, but not sure what is best, my proposal:

self
data
origin
origin_format
valuation
valuation_format
age
index
columns
cumulative
trailing
pattern
array_backend

Hopefully I didn't miss any

kennethshsu:
I think this is going to be my favorite PR in a long time.

@genedan  does what @henrydingliu make sense? This is one of those things that had bothered me a lot. `development` is to describe how old, not when, and `valuation` is a better way to describe "when", and `age` is the better way to describe how old.

genedan:
I'm in agreement, I'll put in the review soon.

chainladder/utils/data/_manifest.py
henrydingliu: i was surprised too. raised a new issue #1400 to bring awareness

chainladder/utils/data/_manifest.py
henrydingliu: yeah, basically, they can do ["year_col","month_col"]. gonna assign you #1400 for noticing!

kennethshsu: Ahh ok, maybe just need to improve the docstring a bit better.

chainladder/core/triangle.py
henrydingliu: yeah cursor screams at you if you don't put age at the end. basically, someone who passes parameters by position would get messed up if we put age earlier up the argument list.

kennethshsu: Ahhh... I see, but that's only if you use the implied order right? Ok I hate it but I guess we can't do much about it.

Do you think people often use the implied parameters ordering without explicitly saying what is what?

chainladder/_config/deprecation.py
henrydingliu: yeah completely open to wordsmithing here. will let gene propose something since i already butchered his original message. 

the overall intent is, assuming we are aligned. we are gonna deprecate this name (either function or arg), you can still use it for now, till this new version. 

i have been tying deprecation to versions. but also open to tying to date.

chainladder/core/triangle.py
henrydingliu: i mean, in general the best practice guides tell you to not use positional arguments for anything with more than a couple of arguments. i actually wouldn't mind at all if we did reorder. 

this is why we need multiple codeowners to review a major API change :)

genedan:
What do you think of having the deprecation version being optional? I'm leaning towards making it a mandatory parameter. If we don't assign a version number, it's going to be harder to track down and schedule features for removal.

henrydingliu:
> What do you think of having the deprecation version being optional? I'm leaning towards making it a mandatory parameter. If we don't assign a version number, it's going to be harder to track down and schedule features for removal.

do you think other contributors will be comfortable naming versions? we can always recommend a version during PR review

kennethshsu:
> What do you think of having the deprecation version being optional? I'm leaning towards making it a mandatory parameter. If we don't assign a version number, it's going to be harder to track down and schedule features for removal.

I think we should announce when it began being deprecated, and just say we will remove it in a future version, then just follow the governing doc and leave it for 3 months min after it's been released.

So basically, "this function is deprecated on v0.x.0, and will be removed in an upcoming release, use ____ instead". That way we have reasonable documentation on when deprecation began, and it is also easy to just check the time requirement. Thoughts?

henrydingliu:
> > What do you think of having the deprecation version being optional? I'm leaning towards making it a mandatory parameter. If we don't assign a version number, it's going to be harder to track down and schedule features for removal.
> 
> I think we should announce when it began being deprecated, and just say we will remove it in a future version, then just follow the governing doc and leave it for 3 months min after it's been released.
> 
> So basically, "this function is deprecated on v0.x.0, and will be removed in an upcoming release, use ____ instead". That way we have reasonable documentation on when deprecation began, and it is also easy to just check the time requirement. Thoughts?

Did you comment in the right PR? we are not deprecating a function here.

kennethshsu:
Bad wording on my part, but my point is the same, it should say something like 

"`development` is deprecated on v0.x.0, and will be removed in an upcoming release, use `valuation` instead."

My point is that we just announce when something will begin being deprecated, without saying when the end of life is.

kennethshsu:
But it should be mandatory, basically I agree with what @genedan proposed. 

> If we don't assign a version number, it's going to be harder to track down and schedule features for removal.

henrydingliu:
> Bad wording on my part, but my point is the same, it should say something like
> 
> "`development` is deprecated on v0.x.0, and will be removed in an upcoming release, use `valuation` instead."
> 
> My point is that we just announce when something will begin being deprecated, without saying when the end of life is.



> But it should be mandatory, basically I agree with what @genedan proposed.
> 
> > If we don't assign a version number, it's going to be harder to track down and schedule features for removal.

to be clear, you just said two conflicting things. 

earlier you said, do not promise a version where it's removed

  here you just said, you agree with gene, whose question was whether we need to make the removal version mandatory

kennethshsu:
> > Bad wording on my part, but my point is the same, it should say something like
> > "`development` is deprecated on v0.x.0, and will be removed in an upcoming release, use `valuation` instead."
> > My point is that we just announce when something will begin being deprecated, without saying when the end of life is.
> 
> > But it should be mandatory, basically I agree with what @genedan proposed.
> > > If we don't assign a version number, it's going to be harder to track down and schedule features for removal.
> 
> to be clear, you just said two conflicting things.
> 
> earlier you said, do not promise a version where it's removed
> 
> here you just said, you agree with gene, whose question was whether we need to make the removal version mandatory

I meant we document when something is beginning to be deprecated, and no need to document when it will be deprecated, so use

"`development` is deprecated on vX, and will be removed in an upcoming release, use `valuation` instead."

instead of 

"`development` will be deprecated on vY,, use `valuation` instead."

See [here](https://github.com/casact/chainladder-python/pull/1401/changes#diff-b02b84fb388b12d2833f020a0b678b6410b67efe3ea524b8067707ff6568400aR292).

henrydingliu:
okay. say you are a user of this package. and you get a deprecation warning like this. what exactly is the value of knowing when a deprecation was first introduced?

kennethshsu:
> okay. say you are a user of this package. and you get a deprecation warning like this. what exactly is the value of knowing when a deprecation was first introduced?

To know when it began? I.e. you've been warned since ____. I'm not saying it's more valuable than knowing "this function will no longer work by this date" but it is easier for us maintainers to keep track. It's a good balance between keeping the users informed but without overburdening ourselves with promising a date that we must keep to.

henrydingliu:
>To know when it began? I.e. you've been warned since ____. 

dude, that's so passive aggressive. do you not hear it?

>but it is easier for us maintainers to keep track

we shouldn't be relying on warning messages to keep track. a centralized list somewhere on the repo is much more effective.

kennethshsu:
> > To know when it began? I.e. you've been warned since ____.
> 
> dude, that's so passive aggressive. do you not hear it?
> 

Well you asked and I answered lol. I also think a bit of passive aggressiveness here isn't a bad thing? We want to warn the user to basically migrate before they complain? All jokes aside, what do you not like about this?

> > but it is easier for us maintainers to keep track
> 
> we shouldn't be relying on warning messages to keep track. a centralized list somewhere on the repo is much more effective.

That's a fair point, but it does mean we will need to maintain another list, which isn't bad I think.

Let me reply and add more context in the other [thread](https://github.com/casact/chainladder-python/discussions/1027#discussioncomment-18560975).

chainladder/_config/deprecation.py
```diff
@@ -239,11 +239,11 @@ def _deprecated_rename_argument(
     category: type[Warning] = FutureWarning,
 ) -> Callable[[_F], _F]:
     """
-    Decorator factory that marks a keyword argument as scheduled to be
-    renamed.
+    Decorator factory that marks a keyword argument that has been rename
```
genedan: renamed*

chainladder/_config/deprecation.py
genedan: I had commented earlier on #1027, but I think the 3 month thing needs to be revised in favor of deprecating by version. Historically we haven't been so good at committing to a specific release date, but we have been pretty good at releasing once we've closed out enough tickets. @henrydingliu's comment about copying pandas works for me. They do at least 2 minor releases which would work for me.

chainladder/core/triangle.py
genedan: What if we reach 2.0.0 before we ever reach 1.5?

chainladder/core/triangle.py
```diff
@@ -50,11 +51,16 @@ class Triangle(TriangleBase):
...
-    development: str
-        Name of the column in ``data`` representing the development or
-        valuation period. Maps to the Development dimension. If omitted, the
+    valuation: str
+        Name of the column in ``data`` representing the
+        valuation date. Maps to the Development dimension. If omitted, the
         Triangle is treated as having a single development period (e.g. a
-        latest-diagonal-only view).
+        latest-diagonal-only view). Cannot be used in conjunction with ``age``
+    age: str
```
genedan: Could we add an example where the new age parameter is used?

genedan:
I'm good with the syntax, I only just need an example of supplying ages because I've never actually seen that used before and it would be good to have an example of that in the Triangle docstring.

For the deprecation though we do need to iron out some details in the policy on #1027 .

PR #1409, opened 2026-09-20
Deprecate the inplace parameter of Triangle.astype

### LINKED ISSUE #1063: Deprecate inplace parameter in TrianglePandas.astpye().
### Description

The Pandas analogue `DataFrame.astype()` does not have an `inplace` parameter. Since inplace replacement is a discouraged practice, I think we should drop this parameter. A search of the codebase and docs yielded no current use cases of this parameter.

- [x] Yes, absolutely!

_No response_

_No response_

- [ ] Yes, absolutely!
- [ ] No.

Abhayindia:
Closes #1063.

Applies the new `_deprecated_drop_argument` helper from #1323 to the `inplace` parameter of `TrianglePandas.astype()`, and fixes the docstring, which documented a `copy` parameter that does not exist and never documented `inplace`.

Behaviour is unchanged: the parameter still works exactly as before and only emits a `FutureWarning` when it is supplied.

Two things worth a maintainer decision rather than my guess:

**1. No `version` is passed.** The warning reads "will be removed in a future release." #1063 has no milestone, and removing a parameter is breaking, so I did not want to assert a release that has not been agreed. Happy to add `version=` once it is milestoned.

**2. Positional callers are not warned.** `_deprecated_drop_argument` warns only when the name is in `kwargs`, and `astype(self, dtype, inplace=True)` accepts `inplace` positionally. Measured:

| call | warns |
|---|---|
| `astype("float32", inplace=False)` | yes |
| `astype("float32", False)` | **no** |
| `astype("float32")` | no (intended) |

Closing that would mean either making `inplace` keyword-only, which is itself a breaking change, or hand-rolling a check instead of using the shared helper. I did neither, since both go beyond this ticket.

Also worth flagging for whoever writes the removal PR: `inplace` defaults to **True**, so `astype(dtype)` currently mutates in place. If removal makes `astype` match `DataFrame.astype` and always return a new object, that silently flips the default for every existing caller. This is not theoretical, it caught me while writing the test for it.

Tests: `test_astype_inplace_deprecated` covers the copy path, the in-place path, and asserts the parameter-free spelling stays silent. Full suite locally: 1212 passed, 7 skipped. The one failure, `test_wtd_reg.py::TestOLS::test_missing_data`, reproduces unchanged on clean `main` (`module 'sparse' has no attribute 'repeat'`) and is unrelated.


henrydingliu:
a few thoughts:

- are tests necessary for adding a deprecation decorator? is it reasonable to say that as long as the decorator itself is tested (which they are), there's no need to add tests for individual uses of the decorators?
- please use 2.0 as the version
- i'm starting to think we underbaked the appropriate deprecation path for dropping an argument. before we can deprecate an argument, we have to first make sure the default value is consistent with the behavior of the post-drop method. basically, when the user gets the drop deprecation warning, they can just get rid of the argument and forget about it. that's not where we are. so what we actually need is
    1. modify the drop decorator with a mandatory message argument
    2. for this deprecation, add the message that the user should do this for now. (please double check that this actually works)

```python
yeet_tri = tri.copy()
yeet_tri = yeet_tri.astype(new_dtype)
can_be_tri_or_newtri = yeet_tri.copy()
```

- please use the deprecated sphinx directive in the docstring and add the future behavior of the method, as well as the above example for how to use astype in a futureproof manner

Abhayindia:
done: `version="2.0"`, `_deprecated_drop_argument` now takes a mandatory `guidance` string appended to the warning (this PR is its only production caller, so nothing else breaks), and the docstring carries a `.. deprecated:: 2.0` directive giving the post-drop behaviour plus your copy-then-astype example.
double-checked your snippet: copying first gives `('float64','float32','float32')` both with the current default and with `inplace=False` standing in for post-drop semantics, so it is identical across the removal - drop the leading copy and the original is mutated today but would not be afterwards, which is exactly why the copy is needed. on tests: agreed the decorator does not need re-testing per use, so that one is gone and the remaining test only pins that the documented migration is drop-safe; happy to drop that too if you would rather.

chainladder/core/pandas.py
```diff
@@ -1058,16 +1059,36 @@ def rename(
...
+            The ``inplace`` parameter is deprecated and will be removed in 2.0.
+            ``astype`` will then always return a new Triangle and never modify
+            the original, matching ``DataFrame.astype``, which has no
+            ``inplace`` parameter. Because ``inplace`` currently defaults to
+            ``True``, that is a change of behaviour for callers who rely on the
+            default. Copy first and use the return value, which behaves the
+            same before and after the removal::
+
+                tri = tri.copy()
+                tri = tri.astype(new_dtype)
```
henrydingliu, line 1085: this displays weirdly. could be my browser

Image

Abhayindia: not your browser, it's real - built it locally and the `::` shorthand isn't honoured inside a `.. deprecated::` directive, so the block degraded to `<blockquote><div><p>` and the browser collapsed the newline, which is the joined line you saw; the trailing `::` then stayed as visible text for the same reason.
switched it to an explicit `.. code-block:: python`, and re-rendered the actual docstring through sphinx to confirm: no stray `::`, one `<div class="highlight">` block, and `'tri = tri.copy()\ntri = tri.astype(new_dtype)'` on two lines with no blockquote fallback.

henrydingliu (approved):

henrydingliu:
@Abhayindia thanks so much!
@genedan i approved but would greatly appreciate a second look!

Issue #1314, opened 2026-09-08
[DOC] Move code blocks in Jupyter Markdown cells to their own Code cells.


genedan:
The Jupyter docs have two different types of cells: "Markdown", and "Code".

Ruff is unable to check Python syntax within Markdown cells. For example, in the following Markdown cell, the Python code wrapped within the triple backticks (```) was not caught by Ruff:

<img width="1381" height="789" alt="Image" src="https://github.com/user-attachments/assets/4bda38c9-2e4f-44b2-9c21-b4124bae2818" />

What should have happened was that single quotes should have been flagged and replaced by double quotes. When possible, we should move these blocks into separate cells.

Abhayindia:
Happy to take this. Do you want every ```python block in a markdown cell moved to a code cell across all the docs, or just the ones ruff would actually flag (e.g. quote style)? A repo-wide sweep touches a lot of notebooks, so I'd rather match the scope you have in mind.

genedan:
You could do a stacked pull request to make thing easier for the reviewers. I think one folder's worth of notebooks per stack would do.

We should move the blocks into their own cells (where possible) across the docs, because in the event that they might be edited in the future, we would like for Ruff to examine them. I think there may be certain cases where this might not be possible (such as for inline code).

Abhayindia:
took a look - all 19 python blocks in markdown cells are non-executable (undefined names like `Estimator`/`data`, or deliberate error demos like the non-sliceable model in methods.ipynb), so moving them as-is fails the build since execute_notebooks is force.
could put them in code cells tagged skip-execution / raises-exception instead - ruff lints them, build stays green. want me to go that way?

henrydingliu:
@genedan bumping

genedan:
Some of these may involve a bit of judgment. In the notebook you mentioned, I saw this cell:

<img width="1361" height="231" alt="Image" src="https://github.com/user-attachments/assets/337f4f26-cfb1-4a6e-94b1-6fa3870c9fc8" />

We should move that code into its own cell, and deliberately print the exception (this would be consistent with other online package tutorials) so that the reader knows what an exception looks like (we might have to explain what an exception is in one of the beginner guides if we haven't already). 

The cell can then be marked with metadata so it doesn't crash our workflows:
````
  ```{code-cell} ipython3
  :tags: [raises-exception]
 ```
````

[...]

PR #1410, opened 2026-09-20
[DOC] Move user_guide code blocks into their own code cells

Abhayindia:
Part of #1314. First stack: `docs/user_guide`.

Moves the four executable python blocks out of markdown cells and into real code cells tagged `raises-exception`, as suggested in [this comment](https://github.com/casact/chainladder-python/issues/1314#issuecomment-5744708973). Ruff now lints them, and the traceback renders in the docs so readers see what the exception actually looks like.

| notebook | block | raises |
|---|---|---|
| `methods.ipynb` | `model['paid'].ultimate_` | `TypeError: 'Chainladder' object is not subscriptable` |
| `triangle.ipynb` | numpy broadcasting equivalent | `ValueError: operands could not be broadcast together` |
| `triangle.ipynb` | virtual column missing `EarnedPremDIR` | `ValueError: list.index(x): x not in list` |
| `triangle.ipynb` | boolean filter across two axes | `ValueError: Slice returns empty Triangle.` |

The last one is split into two cells, since a `raises-exception` cell stops at the exception and the corrected chained form needs to run.

Verified: `pytest --nbmake` passes on both notebooks, and ruff check/format are clean. I also confirmed the tag is load-bearing rather than decorative: removing `raises-exception` from the `methods.ipynb` cell makes nbmake fail with that exact `TypeError`, and restoring it passes.

**Seven blocks are deliberately left alone, because they cannot execute.** Flagging rather than guessing:

- `development.ipynb` (6 blocks) illustrate API shape with a placeholder `Estimator` and `data`, for example `estimator = Estimator(param1=1, param2=2)`.
- `triangle.ipynb` cells 4 and 7 call `cl.Triangle(data, ...)`, and `data` is never defined anywhere in that notebook.
- `triangle.ipynb` has `cl.Triangle(..., array_backend='cupy')`, a literal ellipsis that also needs a CUDA card.

`F` is in the ruff select list, so moving any of these into a code cell is an F821 undefined-name failure, and `nb_execution_mode = "force"` means they would also break the docs build. Options as I see them: leave them as markdown, add stub definitions so they genuinely run, or move them with a `skip-execution` tag plus a per-file F821 ignore. Happy to take whichever you prefer in the next stack.

Also worth noting for later: `triangle.ipynb` cell 4's fence was written as ```` ```python ```` with a trailing space, which renders identically and is easy to miss when grepping for these blocks.

Next stacks: `docs/user_guide/development.ipynb` (pending the question above) and `docs/getting_started/tutorials` (2 blocks).


henrydingliu:
looks good. mind picking up this nearby dead link?

image

Abhayindia:
done in 297111b7. it wasn't http-dead (the 1.18 page still 200s) - it was written as rST ```text <url>`_`` inside a MyST cell, so it never linkified and rendered raw; repointed at `stable/user/basics.broadcasting.html` in markdown syntax so it can't go stale.
swept the rest of docs for the same pattern and that was the only one, and no other version-pinned doc URLs. nbmake still green.

henrydingliu (approved):

henrydingliu:
@Abhayindia thanks!

Issue #1396, opened 2026-09-18
[BUG] Dunders bugs


genedan:
There were several dunders bugs discovered when working on #1342. Highlighting cells depends heavily on boolean logic so we should make sure we have that locked down. This issue serves as a parent issue for each of the dunder bugs.

I believe most of these are quite small so they are good for newer contributors, with the exception of #1395 which is a little more involved.

Issue #1395, opened 2026-09-18
[BUG] Use __eq__ and .equals() to distinguish between elementwise and whole-object comparison


genedan:
### Are you on the latest chainladder version?

- [x] Yes, this bug occurs on the latest version.

### Describe the bug in words

Consider the following example:

```python
import pandas as pd

df1 = pd.DataFrame({'a': [1, 2], 'b': [4, 5]})
df2 = pd.DataFrame({'a': [1, 3], 'b': [7, 5]})

df1 == df2

       a      b
0   True  False
1  False   True

df1.equals(df2)

True
```

The `==` operator does elementwise comparison, returning an array of bools. The `.equals` method compares the whole object, returning a boolean scalar.

### How can the bug be reproduced?

Chainladder does not have elementwise comparison, and the `==` acts as whole-object comparison

```python
import chainladder as cl

raa = cl.load_sample('raa')

raa == raa

True

raa.equals(raa)

Traceback (most recent call last):
    ~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<ipython-input-36-a11e75dbc397>", line 1, in <module>
    raa.equals(raa)
    ^^^^^^^^^^
AttributeError: 'Triangle' object has no attribute 'equals'
```

### What is the expected behavior?

`==` should do elementwise comparison. The `.equals()` method should be defined to do whole-object comparison, replacing the current functionality of `==`.

### Would you be willing to contribute this ticket?

- [ ] Yes, absolutely!
- [ ] Yes, but I would like some help.
- [ ] No.

henrydingliu:
Is this {BUG] or [BRK]? can we do anything more than a deprecation warning?

genedan:
I'd classify it as a bug since `==` shouldn't have been returning a scalar in the first place. According to PDEP-17, breaking changes caused by bugs don't need a warning.

PR #1412, opened 2026-09-20
Fix the ordering comparison dunders and support Triangle operands

### LINKED ISSUE #1391: [BUG] __gt__ and __ge__ dunders are undefined

### Describe the bug in words

Discovered when working on #1342. Trying to do something like `raa > 10000` fails.

This comparison works:

```python
import chainladder as cl
raa = cl.load_sample('raa')

raa < 10000

       12     24     36     48     60     72     84     96     108    120
1981  True   True  False  False  False  False  False  False  False  False
1982  True   True   True  False  False  False  False  False  False   True
1983  True   True  False  False  False  False  False  False   True   True
1984  True  False  False  False  False  False  False   True   True   True
1985  True   True  False  False  False  False   True   True   True   True
1986  True   True  False  False  False   True   True   True   True   True
1987  True   True  False  False   True   True   True   True   True   True
1988  True   True  False   True   True   True   True   True   True   True
1989  True   True   True   True   True   True   True   True   True   True
1990  True   True   True   True   True   True   True   True   True   True
```

But this one doesn't:

```python
raa > 10000

Traceback (most recent call last):
...
  File "<ipython-input-16-bbd772ba6406>", line 1, in <module>
    raa > 10000
TypeError: '>' not supported between instances of 'Triangle' and 'int'
```

We should get a triangle back filled with the correct bools. The `__gt__` and `__ge__` dunders need to be defined.

- [ ] Yes, absolutely!
- [ ] No.

### LINKED ISSUE #1392: [BUG] Bug in __le__ dunder

### Describe the bug in words

The `__le__` (less than or equal to) dunder is a copy of `__lt__` (less than)

Consider the following example:

```python
import chainladder as cl
raa = cl.load_sample('raa')

raa

         12       24       36       48       60       72       84       96       108      120
1981  5012.0   8269.0  10907.0  11805.0  13539.0  16181.0  18009.0  18608.0  18662.0  18834.0
1982   106.0   4285.0   5396.0  10666.0  13782.0  15599.0  15496.0  16169.0  16704.0      NaN
1983  3410.0   8992.0  13873.0  16141.0  18735.0  22214.0  22863.0  23466.0      NaN      NaN
1984  5655.0  11555.0  15766.0  21266.0  23425.0  26083.0  27067.0      NaN      NaN      NaN
1985  1092.0   9565.0  15836.0  22169.0  25955.0  26180.0      NaN      NaN      NaN      NaN
1986  1513.0   6445.0  11702.0  12935.0  15852.0      NaN      NaN      NaN      NaN      NaN
1987   557.0   4020.0  10946.0  12314.0      NaN      NaN      NaN      NaN      NaN      NaN
1988  1351.0   6947.0  13112.0      NaN      NaN      NaN      NaN      NaN      NaN      NaN
1989  3133.0   5395.0      NaN      NaN      NaN      NaN      NaN      NaN      NaN      NaN
1990  2063.0      NaN      NaN      NaN      NaN      NaN      NaN      NaN      NaN      NaN
```

Pay attention to what happens to the `5012.0` value in the top left cell:

```python
raa <= 5012.0

        12     24     36     48     60     72     84     96     108    120
1981  False  False  False  False  False  False  False  False  False  False
1982   True   True  False  False  False  False  False  False  False   True
1983   True  False  False  False  False  False  False  False   True   True
1984  False  False  False  False  False  False  False   True   True   True
1985   True  False  False  False  False  False   True   True   True   True
1986   True  False  False  False  False   True   True   True   True   True
1987   True   True  False  False   True   True   True   True   True   True
1988   True  False  False   True   True   True   True   True   True   True
1989   True  False   True   True   True   True   True   True   True   True
1990   True   True   True   True   True   True   True   True   True   True
```

The upper-left cell should have been `True` instead of `False`.

- [ ] Yes, absolutely!
- [ ] No.

### LINKED ISSUE #1394: [BUG] Triangle-to-Triangle comparison fails

### Describe the bug in words

A comparison such as `raa < raa` fails.

```python
import chainladder as cl
raa = cl.load_sample('raa')

raa < raa
Traceback (most recent call last):
 ... 
line 524, in __lt__
    obj.values = xp.nan_to_num(obj.values) < xp.nan_to_num(value)
                                             ~~~~~~~~~~~~~^^^^^^^
TypeError: no implementation found for 'numpy.nan_to_num' on types that implement __array_function__: [<class 'chainladder.core.triangle.Triangle'>]
```

This should behave like the numpy and Pandas analogues:

```python
import numpy as np

a = np.array([[1.0, 2.0], [3.0, np.nan]])
b = np.array([[1.0, 1.0], [9.0, 9.0]])

print(a)
array([[ 1.,  2.],
       [ 3., nan]])

print(b)

array([[1., 1.],
       [9., 9.]])

a < b
# array([[False, False],
#        [ True, False]])      NaN compares False

a <= b
# array([[ True, False],
#        [ True, False]])      differs from < at the equal cell

a < np.array([[1.0, 2.0, 3.0]])
# ValueError: operands could not be broadcast together with shapes (2,2) (1,3)


import numpy as np
import pandas as pd

d1 = pd.DataFrame({"x": [1.0, 3.0], "y": [2.0, np.nan]}, index=["a", "b"])
d2 = pd.DataFrame({"x": [1.0, 9.0], "y": [1.0, 9.0]}, index=["a", "b"])

print(d1)
     x    y
a  1.0  2.0
b  3.0  NaN

print(d2)
     x    y
a  1.0  1.0
b  9.0  9.0

d1 < d2
#        x      y
# a  False  False
# b   True  False       NaN compares False

d1 <= d2
#       x      y
# a  True  False        differs from < at the equal cell
# b  True  False

d3 = pd.DataFrame({"x": [1.0, 9.0]}, index=["a", "b"])

d1 < d3
#       x   y
# a   2.0 NaN
# b  12.0 NaN
```

- [ ] Yes, absolutely!
- [ ] No.

Abhayindia:
Closes #1391. Closes #1392. Closes #1394. Part of #1396.

Three of the four dunder bugs, fixed together because they are all the same handful of lines.

| Issue | Before |
|---|---|
| #1391 | `raa > 10000` and `raa >= 10000` raise `TypeError`; `__gt__`/`__ge__` were never defined, so Python fell back to the reflected `int` operators |
| #1392 | `__le__` was a verbatim copy of `__lt__`, so it used `<` and excluded the boundary |
| #1394 | `raa < raa` raised `TypeError`, because the right-hand Triangle was passed straight to `nan_to_num` |

Rather than patch the one wrong character in `__le__` and paste two more near-identical methods for `__gt__`/`__ge__`, all four now route through a shared `_compare(value, op)`. That is what makes #1392 structurally unable to recur: there is one implementation, and the operator is a parameter.

`_compare` also unwraps a Triangle operand to its values, moving both sides onto a common backend first using the same `ARRAY_PRIORITY` logic `__eq__` already uses. NaN-as-zero handling is unchanged from the existing `__lt__`.

Verified all four operators agree elementwise with the corresponding numpy operation, and the boundary case is now correct at a value equal to the operand:

```
raa[1981, 12] = 5012.0
 <  : False      <= : True      >  : False      >= : True
```

Backends exercised: sparse vs scalar, sparse vs sparse, numpy vs sparse, and sparse vs numpy all work, the mixed cases resolving to the sparse backend.

Tests: 3 added, covering scalar comparison against numpy, the inclusive boundary, and Triangle-to-Triangle. **I confirmed they are load-bearing** by reverting `dunders.py` to `main` and re-running: 6 of 6 parametrised cases fail without the fix, all pass with it.

Full suite locally: 1216 passed, 7 skipped. The one failure, `test_wtd_reg.py::TestOLS::test_missing_data`, reproduces identically on clean `main` (`module 'sparse' has no attribute 'repeat'`) and is unrelated to this change.

I have deliberately left **#1395** out. Making `==` elementwise and moving whole-object comparison to `.equals()` changes the meaning of every existing `==` in the codebase and in user code, so it wants its own PR and a decision from you rather than being folded in here.


henrydingliu:
@Abhayindia please take a look at the updated requirements in #1394

Abhayindia:
@genedan @henrydingliu done - dropped the `nan_to_num` so NaN propagates and all four operators return `False` at a missing cell, per IEEE 754. On `raa` the result is now identical to pandas for `<`, `<=`, `>` and `>=`, and `5012.0 <= 5012.0` is `True`.
Added a test pinning the missing-value case and updated the two that asserted the old NaN-as-zero behaviour. Full suite green (1218 passed).

henrydingliu (approved):

Issue #1411, opened 2026-09-20
[TEST] Can't replicate a Pandas4 warning


henrydingliu:
Please help!

I'm trying to track down this warning from the [Pandas3 unittest](https://github.com/casact/chainladder-python/actions/runs/35489299340/job/106021367968#step:5:120)

<img width="680" height="253" alt="Image" src="https://github.com/user-attachments/assets/35e0bfc5-29e0-4f2f-9fa8-659754ec9966" />

I'm currently unable to get this warning to emit in my environment. I double checked that I'm on Pandas 3.

ppcvote:
Reproduced. The reason it stays invisible locally is the warning category, not your pandas version.

`Pandas4Warning` subclasses `DeprecationWarning`: `Pandas4Warning -> PandasDeprecationWarning -> PandasChangeWarning -> DeprecationWarning`. Python's default filters ignore `DeprecationWarning` unless it is raised from `__main__`, and this one is raised inside `chainladder/core/dunders.py`, so a plain script or a REPL session stays quiet. pytest switches `always::DeprecationWarning` on, which is why only CI shows it.

On pandas 3.0.6, from the repo root:

```
python -W always::DeprecationWarning -c "
import chainladder as cl
tri = cl.load_sample('usauto')
m = cl.CaseOutstanding(paid_to_incurred=('paid','incurred')).fit(tri)
cl.Chainladder().fit(m.fit_transform(tri))"
```

Three of them, all at `dunders.py:187`. The same script with no `-W` prints nothing at all. `-W error::DeprecationWarning` turns it into a traceback that lands in `pandas/core/reshape/concat.py:498`.

Under pytest you do not need the flag: `pytest chainladder/development/tests/test_outstanding.py` on its own reports all four, because pytest already sets `always::DeprecationWarning`. So it is not test ordering either, that one file reaches the line by itself.

The site is the `odims` concat:

```python
odims = pd.concat(
    (
        pd.Series(obj.odims, index=obj.odims),
        pd.Series(other.odims, index=other.odims),
    ),
    axis=1,
).sort_index()
```

Both indexes are origin dates, so this is the all-DatetimeIndex case pandas is deprecating. The `ddims` concat seven lines above never warns, because development lags are integers.

[...]

PR #1415, opened 2026-09-21
[CHORE] Clear the Pandas4Warning from the odims concat

ppcvote:
## Summary of Changes

The `odims` concat passes `sort=False`. Both sides are origin dates, so it is the all-DatetimeIndex case pandas 3 deprecates, and it accounts for 104 of the pandas3 job's warnings: 4 in `test_triangle.py`, 4 in `test_outstanding.py`, 96 in `test_workflow.py`.

The result does not change, because `.sort_index()` runs on the concatenated frame in the same statement, so the concat's own ordering never survives. I compared `sort=True` against `sort=False` followed by `.sort_index()` on overlapping, disjoint, unsorted and NaT carrying DatetimeIndex pairs, on pandas 2.3.3 and 3.0.6, and the frames compare equal in each, column order included, which is what the `odims[0]` and `odims[1]` lookups below depend on. Duplicate labels that actually differ raise the same `ValueError` either way, so `sort` never gets to matter there.

The `ddims` concat seven lines above is left alone: development lags are integers, and it does not warn.

## Related GitHub Issue(s)

Addresses the `dunders.py:187` half of #1411, which @henrydingliu asked me to open.

The job's other `Pandas4Warning`, `io.py:101` at 17 occurrences in `test_utilities.py`, is a `to_json` `date_format` compatibility call rather than a no-op, so it is out of scope here. Happy to open a follow up issue for it.

## AI/LLM Usage

I used Claude Code. It built the pandas 3 environment, reproduced the warning and ran the before and after comparisons and the `sort` equivalence checks above. The one line change and the decision to leave `io.py` out are mine, and I ran both suites myself.

- pandas 3.0.6, the three files named in the report: 775 warnings before, 671 after, `dunders.py:187` gone, 560 passed and 1 skipped either way. Two `test_repr` failures there are `lxml` missing from the throwaway venv I built for pandas 3, present with and without this change.
- pandas 2.3.3, full suite: 1212 passed, 7 skipped.
- `ruff check` and `ruff format --check` are clean on `dunders.py`, which is not in `per-file-ignores`.

## Submitter's Checklist
- [x] I have reviewed and am adhering to the standards outlined in the project [Governing Doc](https://github.com/casact/chainladder-python/blob/main/docs/library/governance.md).
- [x] The PR subject title summarizes the changes, with one proper prefix (`[FIX]`, `[FEAT]`, `[DOCS]`, `[TST]`, `[CHORE]`, or `[BRK]`).
- [x] I am a human (not a bot), and this PR form is written by a human.

## Reviewer's Checklist
- [ ] The implementation addresses the associated issue(s).
- [ ] The implementation is appropriate, maintainable, and follows ARCHITECTURE.md.
- [ ] PR subject title has the proper prefix and the subject is appropriate.
- [ ] Relevant issue(s) are linked.
- [ ] AI/LLM usage is disclosed and appropriate.
- [ ] Documentation and tests are appropriate.
- [ ] CI tests passed, or any failures are acceptable.
- [ ] Leave a comment with the final recommendation (e.g. approve as is, request a secondary review, or flag an area for more review).


henrydingliu:
ah, i was working on the to_json warning independently (still in draft). thanks for putting this in!

henrydingliu (approved):

PR #1418, opened 2026-09-21
[DOC] Make the triangle creation examples executable

Abhayindia:
Part of #1314, second stack. Follows [henrydingliu's direction](https://github.com/casact/chainladder-python/issues/1314#issuecomment-5755227860) to hardcode a dataset and let the existing code execute against it.

Adds a 2 line-of-business, 2 state, 3x3 annual DataFrame and turns the three python blocks in the Triangle-creation section into real code cells against it.

- The first block drops `"Cal Month"`, as you suggested, so one dataset serves the whole section. #1400 covers a fuller multi-column origin/valuation example.
- The multidimensional block now builds the real 4D object: shape `(4, 2, 3, 3)`, index `Auto/CA, Auto/NY, Home/CA, Home/NY`.
- The third block becomes the comparison you asked for. The sample data is cumulative, so `incr_to_cum` on a Triangle that has not been told this cumulates a second time:

| | 12 | 24 | 36 |
|---|---|---|---|
| `cumulative=True` | 570 | 970 | **1170** |
| unset | 570 | 1540 | **2710** |

`pandas` had to be imported in the first cell rather than beside the data, because `E402` is inside the repo's selected `E4` rules.

**One judgement call worth your view.** The first two blocks do not pass `cumulative`, so they each emit the "cumulative property of your triangle is not set" warning in the rendered docs. I left them that way deliberately: it keeps the basic example minimal, and the warning lands immediately before the "Other Parameters" section that explains it. If you would rather the docs be warning-free I will set `cumulative=True` on both, though that does make the later `cumulative=True` example somewhat redundant.

**Still not movable, and not touched here:** `cl.Triangle(..., array_backend='cupy')` in this notebook is a literal ellipsis and needs a CUDA card, and the six blocks in `development.ipynb` use a placeholder `Estimator` and `data`. Both are F821 under the selected rules. Happy to take a direction on `development.ipynb` in a third stack.

Verified: `pytest --nbmake docs/user_guide/triangle.ipynb` passes, ruff check and format are clean, and I executed the notebook and read back the rendered outputs rather than assuming.


henrydingliu:
>One judgement call worth your view. The first two blocks do not pass cumulative, so they each emit the "cumulative property of your triangle is not set" warning in the rendered docs. I left them that way deliberately: it keeps the basic example minimal, and the warning lands immediately before the "Other Parameters" section that explains it. If you would rather the docs be warning-free I will set cumulative=True on both, though that does make the later cumulative=True example somewhat redundant.

can we set in the first two to get rid of the warning; then leave it unset for the 3rd, which will emit the warning as well as resulting in different values? we should also get rid of the "note" block. 

one other visual thing, can you please hardcode the triangle data in tuples? i.e. how it would typically get stored in a csv. it'll make a new user feel more at home.

Abhayindia:
done: `cumulative=True` on the first two blocks so they no longer warn, left unset on the third so it still warns and still gives 2710 against 1170, and the `{note}` block is gone. the standalone `cumulative=True` example went with it, since the first block now carries the parameter anyway.
the data is row tuples with the column names underneath, closer to how it would come out of a csv. re-executed the notebook and confirmed the warning now appears only in that one deliberate cell; nbmake green and ruff clean.

henrydingliu (approved):

PR #1315, opened 2026-09-08
[CHORE] spec0 dependency workflow

genedan:
## Summary of Changes  

Adds a nightly workflow to check if dependencies are out of date. 

- For the minimum Python version, checks against EOL.
- For core dependencies, checks against Spec-0.

If a dependency is within 3 months of getting dropped, the workflow will create an issue.

## Related GitHub Issue(s)  

#1130

## Additional Context for Reviewers  

If this works correctly, the next day after approval we should see 3 issues created for:

- Python 3.10
- Numpy
- Scikit-Learn

since these are due to be dropped according to the above criteria.

## Checklist
- [ ] I passed tests locally for both code (`uv run pytest`) and documentation changes (`uv run --directory docs jb build . --builder=custom --custom-builder=doctest`)

---

> [!NOTE]
> **Low Risk**
> CI-only automation with no runtime or library code changes; issue creation is scoped to the repo token and deduplicated by issue title.
> 
> **Overview**
> Adds **automated SPEC 0 / Python EOL monitoring** for minimum supported versions declared in `pyproject.toml`.
> 
> A new script (`check_spec0.py`) reads `requires-python` and core floors (numpy, pandas, scikit-learn, matplotlib), compares Python against **endoflife.date** EOL dates and packages against **SPEC 0’s 2-year rule** (using PyPI release history), and flags items **within 90 days** or **past** the recommended drop date. It can emit a Markdown table, JSON violations, and **open GitHub issues** via `gh` when no matching title already exists.
> 
> A **daily GitHub Actions workflow** (`spec0_check_nightly.yml`) runs the check with `uv`, appends the summary to the job summary, and enables `--create-issues` using `GITHUB_TOKEN` (`issues: write`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3c56ace4b866f48478abb53ca9ec8471cd810694. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #1420, opened 2026-09-21
Stop the codecov patch status failing on non-library diffs and merge commits

### LINKED ISSUE #1353: [Bug] Codecov patch bugginess
From our observations, the Codecov patch workflow has been failing in some weird ways:

- Failing when a PR, even if it does not modify library code
- Failing when merging, even after passing during a PR

Abhayindia:
Closes #1353.

The issue lists two symptoms. They have different causes, so this fixes them separately.

**"Failing when a PR, even if it does not modify library code"**

`codecov.yml` configured only `project`, so `patch` fell back to `target: auto` and there was no `ignore`. Every file in a diff was scored, including files the test suite never imports. 6b51f649 is the clearest case: it changed only `.github/scripts/check_spec0.py` and `.github/workflows/spec0_check_nightly.yml`, no library code at all, and `codecov/patch` failed at 85.71% against a 91.70% target while `codecov/project` passed.

The `ignore` list covers `.github/**`, `docs/**`, `conftest.py` and `**/tests/**`. Checking the compiled patterns Codecov returns against the current file list: 58 library `.py` files remain scored, all under `chainladder/`, and no non-test file under `chainladder/` is ignored.

**"Failing when merging, even after passing during a PR"**

Codecov compares a merge commit against its **second** parent, the PR branch tip, rather than the first. So after the merge it re-scores a diff that is not the change that was reviewed. Current examples:

| merge commit | first parent | second parent | codecov says |
|---|---|---|---|
| 2f302f91d | a95b86a56 | 75e01efa3 | "compared to 75e01ef" |
| 8a42e835 | 7f8f616f5 | 445a2eff2 | "compared to 445a2ef" |

No `codecov.yml` key controls parent selection, so this cannot be fixed from the repo. `only_pulls: true` avoids it instead: the patch status is only posted on pull requests, which is the only place its result is meaningful. The `project` status is unaffected and still runs everywhere.

The file validates against `https://codecov.io/validate`.


henrydingliu:
@Abhayindia thanks for looking into this!

on the first piece, it looks like the net new thing we are excluding is just the rtests, am i reading that correctly?

on the second piece, i'm not sure it's a best practice. basically, if main get updated between PR raising and PR merging, there could be uncatched indirect coverage change?

Abhayindia:
@henrydingliu good catch on both - narrowed `ignore` to just the six `rtest_*.py` files, which pytest never collects so they report no coverage at all. Everything else was a no-op anyway since CI runs `--cov=chainladder`.
On the second: `project` still runs everywhere including merge commits, so indirect changes from main are still caught - `patch` only ever scores lines inside the diff, so it could not catch those either way. Also found 6b51f649 is itself a merge commit compared against #1315's head, so both symptoms in the issue are the same bug; PR body updated.

henrydingliu (approved):

